PR #4, opened 2019-04-03
added delete script

DanSanche:
As discussed, I added a script to delete resources provisioned by Stackdriver Sandbox

I decided to just find and delete the GCP project (after a prompt to the user). Deleting the project will also delete all resources, so it seems like the cleanest way to go.


terraform/delete.sh
```diff
@@ -0,0 +1,55 @@
...
+# This script deletes the GCP project provisioned by Stackdriver Sandbox
+
+#set -euo pipefail
+set -o errexit  # Exit on error
+
+log() { echo "$1" >&2;  }
+
+get_project(){
+    gcloud projects list --filter="name:stackdriver-sandbox-*" \
+                         --format="value(projectId)"
```
bleything: what happens if they somehow have more than one stackdriver-sandbox project? like if several people sharing an org ran the tool?

terraform/delete.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# ensure the working dir is the script's folder
+SCRIPT_DIR=$(realpath $(dirname "$0"))
+cd $SCRIPT_DIR
+
+PROJECT_ID=$(get_project)
+
+# find and delete the Stackdriver Sandbox GCP project
+if [[ -z $PROJECT_ID  ]]; then
+    log "error: Stackdriver Sandbox project not found"
```
bleything: I like to do sanity checking before anything else; consider moving this above the `cd`

terraform/delete.sh
```diff
@@ -0,0 +1,55 @@
...
+    exit 1
+fi
+log "attempting to delete $PROJECT_ID"
+gcloud projects delete $PROJECT_ID
+
+
+# if user backed out of project deletion, stop script
+PROJECT_ID=$(get_project)
```
bleything: this'll bite you in the case where there are multiple sandbox projects. Maybe use `gcloud projects describe` instead, since you already have the project ID.

this comment is irrelevant if you switch to `terraform destroy`

simonz130:
Could you also update the readme.md with project deletion instruction?

terraform/delete.sh
```diff
@@ -0,0 +1,55 @@
...
+
+PROJECT_ID=$(get_project)
+
+# find and delete the Stackdriver Sandbox GCP project
+if [[ -z $PROJECT_ID  ]]; then
+    log "error: Stackdriver Sandbox project not found"
+    exit 1
+fi
+log "attempting to delete $PROJECT_ID"
+gcloud projects delete $PROJECT_ID
```
bleything: consider using `terraform destroy` instead

terraform/delete.sh
```diff
@@ -0,0 +1,55 @@
...
+
+
+# if user backed out of project deletion, stop script
+PROJECT_ID=$(get_project)
+if [[ ! -z $PROJECT_ID  ]]; then
+    log "project $PROJECT_ID not deleted"
+    exit 1
+fi
+
+# remove tfstate file so a new project id will be generated next time
```
bleything: This shouldn't be necessary if you let terraform delete the resources

terraform/delete.sh
```diff
@@ -0,0 +1,55 @@
...
+log "removing tfstate file"
+if [[ -f "./terraform.tfstate"  ]]; then
+    rm terraform.tfstate
```
bleything: you could use `rm -f` instead of checking the file exists. If you want to stick with the conditional mode, I'd suggest moving the log line inside the conditional.

terraform/delete.sh
```diff
@@ -0,0 +1,55 @@
...
+    log "error: Stackdriver Sandbox project not found"
+    exit 1
+fi
+log "attempting to delete $PROJECT_ID"
+gcloud projects delete $PROJECT_ID
+
+
+# if user backed out of project deletion, stop script
+PROJECT_ID=$(get_project)
+if [[ ! -z $PROJECT_ID  ]]; then
```
bleything: `-n $PROJECT_ID` is a little cleaner than `! -z $PROJECT_ID`, I think? I'm not 100% sure but I think you can drop the test entirely and just do `if [[ $PROJECT_ID ]]; then`, too

terraform/delete.sh
DanSanche: What benefits would terraform destroy give us over just deleting the whole project? I can think of a couple downsides:
- What if someone accidentally deletes the tfstate files? 
- using terraform requires checking for/installing terraform if it's not installed
- in general, I prefer to avoid relying external dependencies when they're not needed. gcloud is more fool-proof if someone just wants to tear everything down. Less potential erros/bugs

terraform/delete.sh
DanSanche: I'd argue that ensuring we're in the appropriate directory is a form of sanity checking. For example, what if we enter another sanity check that references a local file? I think it makes sense to ensure the whole script is running in the expected context right at the top

terraform/delete.sh
DanSanche: good point. For now it should prompt to delete them all, but it's probably more useful to let the user select out of a list. I'll add that

terraform/delete.sh
bleything: Totes, I get that. This was a minor stylistic suggestion, I don't feel strongly about it :)

terraform/delete.sh
bleything: The project and most of its resources were created by terraform, so I think we've already burned the external dependency bridge, at least as far as terraform is concerned. We know it's installed because if it wasn't they wouldn't have been able to set up the sandbox in the first place.

I'm not sure I agree that gcloud is less error-prone. I got an error just now trying to delete a project that I definitely own... and that project was still deleted, despite gcloud telling me I didn't have permission to delete it.

One possible advantage to using terraform is that it will delete all of the underlying resources before deleting the actual project. This means that if someone restores the project during the grace period, it'll be empty. Deleting the project with gcloud means that if you restore the project all of your resources will come back with it.

Ultimately it probably doesn't matter which tool we use. Either way we'll end up with a deleted project. But, again, one of the reasons we chose terraform was its ability to clean up after itself. Maybe that's no longer important, I've been out of the loop for a while.

terraform/delete.sh
bleything: you could also have it throw an error if there's more than one and give the user a way to specify which they mean at invocation time? I dunno if that's any better though

terraform/delete.sh
DanSanche: - My main concern is still if the tfstate files go missing (again, cloud shell isn't always a fully stable environment). Wouldn't that impact terraform's ability to delete what it set up?
- The dependency is still there since cloudshell installations don't persist, so we'd need to treat the environment like a blank slate
- is the advantage you mentioned really even an advantage? I'd say if a customer attempts to restore a project, they'd expect the resources to come back if that's a possibility

I'm open to using terraform destroy if it really does provide any benefits, but it still seems to me that deleting the project directly is easier and cleaner

bleything: It's true that there'll be problems if the tfstate file disappears. I haven't had any problems with losing stuff from cloud shell homedirs, but if that's really a concern then fair enough.

Installing terraform is dead simple and fast. [Here's how we do it for the inspec workshop](https://gitlab.com/binamov/infrastructure-testing/blob/master/setup.sh#L22-24).

Resource persistence is arguable either way. I was thinking of the scenario where one person deletes the sandbox project and then another person comes along and restores it without knowing what it is and now there's resources that people are potentially being billed for? But maybe that's not a real thing?

I also think using terraform makes this script wayyyyyy simpler and cleaner, but that really just comes down to how much maintenance burden you're willing to accept.

Honestly, it probably doesn't matter that much. Like I said before, using terraform to tear down the sandbox was part of the original design but if that doesn't make sense now then I'm totally fine doing it another way. I've just had so many bad experiences with gcloud that I don't trust it any more.

bleything:
Retracting my request for changes; I trust @DanSanche's judgment on everything that's outstanding.

terraform/delete.sh
```diff
@@ -17,39 +17,48 @@
...
     log "project $PROJECT_ID not deleted"
     exit 1
 fi
 
 # remove tfstate file so a new project id will be generated next time
 log "removing tfstate file"
-if [[ -f "./terraform.tfstate"  ]]; then
-    rm terraform.tfstate
-fi
+rm -rf terraform.tfstate
```
bleything: It'd be safer to just do `-f` rather than `-rf`, but that's minor

DanSanche: good catch

DanSanche:
@bleything I made some changes to prompt the user for which project to delete

@simonz130 I updated the README

bleything:
heck yes, looks awesome!

PR #6, opened 2019-04-04
Success message

DanSanche:
Display useful information when deploy is complete

Let me know if there's anything that would be useful to add

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -73,9 +73,15 @@ displaySuccessMessage() {
     if [[ -n "${created_project}" ]]; then
         gcp_path="$gcp_path?project=$created_project"
     fi
-    loadgen_ip=$(gcloud compute instances list --project "$created_project" \
-                                               --filter="name:loadgenerator*" \
-                                               --format="value(networkInterfaces[0].accessConfigs.natIP)")
+    TRIES=0
+    while [[ -z "${loadgen_ip}" && "${TRIES}" -lt 20  ]]; do
+        sleep 1
+        loadgen_ip=$(gcloud compute instances list --project "$created_project" \
+                                                   --filter="name:loadgenerator*" \
+                                                   --format="value(networkInterfaces[0].accessConfigs.natIP)")
+        TRIES=$((TRIES + 1))
+    done
+
     if [[ -z "${loadgen_ip}" ]]; then
         loadgen_ip='[not found]'
     fi
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -81,6 +81,8 @@ displaySuccessMessage() {
     else
         loadgen_ip='[not found]'
     fi
+    GREEN='\033[0;32m'
+    log $GREEN
     log ""
     log ""
     log "--------------------------------------------------------------"
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -75,7 +75,9 @@ displaySuccessMessage() {
         TRIES=$((TRIES + 1))
     done
 
-    if [[ -z "${loadgen_ip}" ]]; then
+    if [[ -n "${loadgen_ip}" ]]; then
+        loadgen_ip='http://$loadgen_ip'
+    else
         loadgen_ip='[not found]'
     fi
     log ""
@@ -83,9 +85,9 @@ displaySuccessMessage() {
     log "--------------------------------------------------------------"
     log "Stackdriver Sandbox deployed successfully"
     log ""
-    log "Stackdriver Dashboard: https://app.google.stackdriver.com/gke"
-    log "Google Cloud Console Dashboard: http://$gcp_path"
-    log "'Hipstershop' web app address: $external_ip"
+    log "Stackdriver Dashboard: https://app.google.stackdriver.com/accounts/create"
+    log "Google Cloud Console Dashboard: $gcp_path"
+    log "'Hipstershop' web app address: http://$external_ip"
     log "'Locust' load generator web interface: $loadgen_ip"
 }
 
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -84,12 +84,12 @@ displaySuccessMessage() {
     log ""
     log ""
     log "--------------------------------------------------------------"
-    log "Stackdriver Sandbox deployed successfully"
+    log "Stackdriver Sandbox deployed successfully!"
     log ""
-    log "Stackdriver Dashboard: https://app.google.stackdriver.com/accounts/create"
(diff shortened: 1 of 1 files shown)


terraform/install.sh
```diff
@@ -39,18 +43,12 @@ applyTerraform() {
...
 getExternalIp() {
   external_ip=""; 
-  while [ -z $external_ip ]; 
-  do
+  while [ -z $external_ip ]; do
      log "Waiting for Hipster Shop endpoint..."; 
      external_ip=$(kubectl get svc frontend-external --template="{{range .status.loadBalancer.ingress}}{{.ip}}{{end}}"); 
      [ -z "$external_ip" ] && sleep 10; 
   done; 
-
```
simonz130: I see you remove the curl verification for the external IP. Why?
kubectl get service won't provide this.

terraform/install.sh
```diff
@@ -59,6 +57,45 @@ loadGen() {
...
+        gcp_path="$gcp_path?project=$created_project"
+    fi
+    TRIES=0
+    while [[ -z "${loadgen_ip}" && "${TRIES}" -lt 20  ]]; do
+        sleep 1
+        loadgen_ip=$(gcloud compute instances list --project "$created_project" \
+                                                   --filter="name:loadgenerator*" \
+                                                   --format="value(networkInterfaces[0].accessConfigs.natIP)")
+        TRIES=$((TRIES + 1))
+    done
```
simonz130: This block doesn't belong to showing display message function. This is load generation.

terraform/install.sh
```diff
@@ -59,6 +57,45 @@ loadGen() {
...
+
+    gcp_path="https://console.cloud.google.com/kubernetes/workload"
+    if [[ -n "${created_project}" ]]; then
+        gcp_path="$gcp_path?project=$created_project"
+    fi
+    TRIES=0
+    while [[ -z "${loadgen_ip}" && "${TRIES}" -lt 20  ]]; do
+        sleep 1
+        loadgen_ip=$(gcloud compute instances list --project "$created_project" \
+                                                   --filter="name:loadgenerator*" \
```
simonz130: Worth adding log messages - mainly for potential troubleshooting if something will happen in the future

terraform/install.sh
```diff
@@ -59,6 +57,45 @@ loadGen() {
...
+    echo -e $GREEN
+    log ""
+    log ""
+    log "--------------------------------------------------------------"
+    log "Stackdriver Sandbox deployed successfully!"
+    log ""
+    log "     Stackdriver Dashboard: https://app.google.stackdriver.com/accounts/create"
+    log "     Google Cloud Console Dashboard: $gcp_path"
+    log "     Hipstershop web app address: http://$external_ip"
+    log "     Locust load generator web interface: $loadgen_ip"
```
simonz130: Remove "locust". It's implementation detail. If we decide to switch to a different tool loadgen wrapper tool shouldn't change.

terraform/install.sh
DanSanche: If I'm reading it right, the old code would only display Hipstershop's service IP if it was reachable, and just silently continue it if it's not. Is that a useful check? I think we should display the URL to the user matter what, so the check seems a bit unnecessary.

terraform/install.sh
DanSanche: This block is trying to retrieve the url for the load generator's web interface so we can display it back to the user. Are you saying we should move the block into the loadGen function?

terraform/install.sh
simonz130: The idea is to check that the address is accessible and if it is - display "success" and the address to the user. If it's not accessible - we should indicate the failure.

terraform/install.sh
simonz130: yep

terraform/install.sh
DanSanche: Do you think we should completely cancel deployment if the curl verification fails (after a number of retries of course)? Or is there anything we can do to fix any possible issues as part of the script? As it is, I don't think anyone even would even notice the success message among the logs, and nothing happens on error

terraform/install.sh
DanSanche: ok, I'll move it there

terraform/install.sh
simonz130: No need to cancel the deployment.
We can just log it and in subsequent flow indicate to the user that external IP of the hipster shop cluster is not available. Even though there might be more logs that will follow this, if the user will want to troubleshoot the problem and share what happened with us  - log like that can be useful.

terraform/install.sh
```diff
@@ -39,24 +43,59 @@ applyTerraform() {
...
      external_ip=$(kubectl get svc frontend-external --template="{{range .status.loadBalancer.ingress}}{{.ip}}{{end}}"); 
      [ -z "$external_ip" ] && sleep 10; 
   done; 
-
-  log "Verifying that Hipster Shop Frontend is accessible"
-  if [[ $(curl -sL -w "%{http code}\\n" "http://$external_ip" -o /dev/null) -eq 200 ]]
-  then
-    log "Stackdriver Sandbox cluster provisioning has completed successfully! Access it at http://$external_ip"
-  fi
+  log "Hipster Shop Frontend is available at $external_ip"
```
simonz130: http://$external_ip?

terraform/install.sh
```diff
@@ -39,24 +43,59 @@ applyTerraform() {
...
 }
 
 # Install Load Generator service and start generating synthetic traffic to Sandbox
 loadGen() {
   log "Running load generator"
   ../loadgenerator-tool startup --zone us-west2-a $external_ip
+  # find the IP of the load generator web interface
+  TRIES=0
+  while [[ -z "${loadgen_ip}" && "${TRIES}" -lt 20  ]]; do
```
simonz130: Is it taking so long that you decided on 20 retries?
Can we reduce it to, say, 10?

DanSanche: We could. I thought since it's sleeping for 1 second each loop, 20 seconds would be a good upper limit, but I can change it to 10 if you prefer

terraform/install.sh
DanSanche: ok, I added it back with logs for either result

DanSanche:
@simonz130 The success message in this PR is in a good state, but the load generator isn't currently accessible. I suggest reviewing and merging David's changes (staged in the fix-loadgenerator branch) before this one

terraform/install.sh
simonz130: Up to you.

PR #7, opened 2019-04-05
Multi billing acct

DanSanche:
The current install script assumes one billing account, and will have issues if there are multiple. This PR presents the user with a selection if more than one account is found

Uses similar logic as the project selector in the delete script https://github.com/GoogleCloudPlatform/stackdriver-sandbox/pull/4

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -23,8 +23,8 @@ log() { echo "$1" >&2; }
 
 getBillingAccount() {
   log "Checking for billing accounts"
-  billingAccounts=$(gcloud beta billing accounts list --format="value(displayName)" --filter open=true)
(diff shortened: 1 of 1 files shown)


terraform/install.sh
```diff
@@ -21,8 +21,39 @@ set -o errexit  # Exit on error
 
 log() { echo "$1" >&2; }
 
-installTerraform()
-{
+getBillingAccount() {
+  log "Checking for billing accounts"
+  found_accounts=$(gcloud beta billing accounts list --format="value(displayName)" --filter open=true)
+  if [ -z "$found_accounts" ] || [[ ${#found_accounts[@]} -eq 0 ]]; then
+    log "No active billing accounts were detected. In order to create a project, Sandbox needs to have at least one billing account"
```
simonz130: Rephrase this: "In order to create sandboxed environment, the script needs to create a new GCP project and associate it with an active billing account"

PR #38, opened 2019-10-23
added new stackdriver monitoring/logging to gke cluster

yuriatgoogle:
Added "new" stackdriver monitoring and logging spec to gke cluster, rather than using the legacy feature.


simonz130:
Merging as the output of the build has nothing to do with the proposed changes.

PR #122, opened 2020-07-14
User experience dashboard

nicklin37:
#WHAT: Terraform support to create example dashboard for User Experience, contains metrics that are relevant to user interaction with the application

#WHY: We want to provide dashboard that reveals metrics about the demo application as whole

#HOW: Additional terraform config file created alongside initial provisioning terraform

### DIFF
--- a/terraform/monitoring/00_provider.tf
+++ b/terraform/monitoring/00_provider.tf
@@ -1,0 +1,33 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# the `provider` block contains the configuration for the provider, including
+# credentials and region config. It can be configured via environment variables
+# or via passing a path to a credentials JSON file.
+
+# In this demo we're using Application Default Creds instead, see this for
+# details: https://cloud.google.com/docs/authentication/production
+#
+# TODO:  we can consider configuring it via env vars
+# that were populated appropriately at runtime.
+
+provider "google" {
+  # pin provider to 3.23.0
+  version = ">=3.23.0"
+
+  # credentials = "/path/to/creds.json"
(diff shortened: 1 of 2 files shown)


terraform/monitoring/00_provider.tf
```diff
@@ -0,0 +1,33 @@
...
+# that were populated appropriately at runtime.
+
+provider "google" {
+  # pin provider to 3.23.0
+  version = ">=3.23.0"
+
+  # credentials = "/path/to/creds.json"
+  # project = "project-id"
+  # region = "default-region"
+  # zone = "default-zone"
```
simonz130, line 32: Should we come up with defaults?

terraform/monitoring/02_dashboards.tf
```diff
@@ -0,0 +1,117 @@
...
+    "columns": "2",
+    "widgets": [
+      {
+        "title": "HTTP Request Count by Method",
+        "xyChart": {
+          "dataSets": [
+            {
+              "timeSeriesQuery": {
+                "timeSeriesFilter": {
+                  "filter": "metric.type=\"custom.googleapis.com/opencensus/opencensus.io/http/server/request_count_by_method\"",
```
simonz130, line 32: FYi @gzhao408, if Nic's PR is merged before your work on OC->OT, we will need to update the name of the metric after we move to OT

simonz130: There are a few more places like that in terraform monitoring

simonz130:
@nicklin37 - nit remark about defaults but otherwise it's ready to go. Please rebase and merge.

terraform/monitoring/00_provider.tf
nicklin37: This file will eventually get merged with the one in /sandbox/terraform so I don't think it makes sense to create defaults here. I think the comments are here just to demonstrate the option.

nicklin37: If anything, we might need to add defaults to the provisioning script provider but that will be for @jiang95-dev

gzhao408:
It looks like you merged master into your branch instead of rebasing your branch on top of master. I recommend:
`git reset --hard 107bdda` to get HEAD back to your top commit.
`git rebase master` to rebase your branch on top of recent master. This puts Dan's commits before yours (you want your updates to be on top of master).
`git push -f` to update remote (here). You need to force push because the parent commit of your base commit is different and thus it's not a simple push.
In general, don't merge anything except into master.

terraform/monitoring/02_dashboards.tf
gzhao408: Got it, will remember to replace these lines when I change the metrics

nicklin37:
Rebased and now merging.

Issue #133, opened 2020-07-16
Remove OpenCensus Jaeger Exporters


gzhao408:
Part of #132 . 
Remove Jaeger exporters from the services: checkoutservice, frontend, AdService, productcatalogservice. These are not related to Ops Management tools and are implemented with sunsetted OpenCensus libraries.

Issue #132, opened 2020-07-16
Convert OpenCensus instrumentation to OpenTelemetry (Main Issue)


gzhao408:
Design doc: [go/ss-opentelemetry-dd](go/ss-opentelemetry-dd)

This is the main issue used to track the OpenCensus to OpenTelemetry project. I'll link all subissues to this one so it's easy to track progress from here.

gzhao408:
# Progress Report

OpenTelemetry Tracing is instrumented + exporting to Cloud Trace:

- [x] Frontend (go)
- [x] Checkout (go)
- [x] Productcatalog (go)
- [x] Shipping (go)
- [x] Email (python)
- [x] Recommendation (python)
- [x] Currency (nodejs)
- [x] Payment (nodejs)
- [ ] Ads (Java) - awaiting [openelemetry-operations-java](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java) trace exporter release
- [ ] Cart (C#)

Sample trace from HipsterShop checkout:
![image](https://user-images.githubusercontent.com/25183001/90453423-74154c80-e0a5-11ea-9e31-b6fb965a100d.png)

PR #135, opened 2020-07-16
Remove opencensus jaeger exporter

gzhao408:
Part of #133, subtask of #132  

Remove jaeger exporter from checkoutservice, shippingservice, frontend, and adservice because it isn't necessary to export to anything other than GCP and OpenCensus is being deprecated.

### DIFF
--- a/src/checkoutservice/main.go
+++ b/src/checkoutservice/main.go
@@ -25,7 +25,6 @@ import (
 	"contrib.go.opencensus.io/exporter/stackdriver"
 	"github.com/google/uuid"
 	"github.com/sirupsen/logrus"
-	"go.opencensus.io/exporter/jaeger"
 	"go.opencensus.io/plugin/ocgrpc"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/trace"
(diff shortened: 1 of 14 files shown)


src/checkoutservice/main.go
```diff
@@ -192,6 +168,10 @@ func (cs *checkoutService) Check(ctx context.Context, req *healthpb.HealthCheckR
 	return &healthpb.HealthCheckResponse{Status: healthpb.HealthCheckResponse_SERVING}, nil
 }
 
+func (cs *checkoutService) Watch(req *healthpb.HealthCheckRequest, srv healthpb.Health_WatchServer) error {
+	return nil
+}
```
simonz130, line 173: Is this related to removing jaeger?

gzhao408: sorry no, i was getting a build error about it so I added it in there and it worked as a bandaid but is probably not the issue. I'm going to rebase and try again without it, will open a separate pr if it's actually necessary

gzhao408: I was clearly confused 😅  just removed all of them

src/checkoutservice/main.go
```diff
@@ -154,7 +131,6 @@ func initStackDriverTracing() {
 }
 
 func initTracing() {
-	initJaegerTracing()
```
Daniel-Sanche, line 157: nit: this function should probably be removed if it contains a single line now

src/shippingservice/main.go
```diff
@@ -185,7 +162,6 @@ func initStackDriverTracing() {
 }
 
 func initTracing() {
-	initJaegerTracing()
```
Daniel-Sanche, line 188: same as above

src/checkoutservice/main.go
gzhao408: done and done

Issue #50, opened 2020-07-07
Add Istio to a hipstershop deployment on a Sandbox 


simonz130:
We need to add istio support, but likely having Anthos Service Mesh is even better option to illustrate how it helps observe the apps.

Daniel-Sanche:
ASM looks like it requires an Anthos subscription. I think we'd be better off using OSS Istio for now. We can use this as a template: https://github.com/GoogleCloudPlatform/bank-of-anthos-scripts/blob/master/install/istio/install_istio.sh

Daniel-Sanche:
Discussed this with Nick, he's going to try handling this, and we'll check back in next week.

Nick:
- [Here's how](https://github.com/GoogleCloudPlatform/microservices-demo#option-4-deploying-on-a-istio-enabled-gke-cluster) we added Istio upstream, but that method is stuck at 1.4, so we're better off using the OSS directions instead
- [This is a script](https://github.com/GoogleCloudPlatform/bank-of-anthos-scripts/blob/master/install/istio/install_istio.sh) we used to enable Istio 1.6 in a different project
- [This looks like](https://github.com/GoogleCloudPlatform/gke-anthos-holistic-demo/tree/master/holistic-demo/istio) another attempt to provision istio using terraform, but I'm not familiar enough with Terraform to say if it's useful for us

Let me know if you run into any issues

nicklin37:
This has been resolved in #139

PR #139, opened 2020-07-16
Install Istio 1.6.2 for Stackdriver Sandbox

nicklin37:
#WHAT: Installs Istio version 1.6 onto the Stackdriver Sandbox cluster during provisioning script. Bash script is run through terraform resource where 1) Istio 1.6.2 will be downloaded and installed and 2) services will be redeployed so that Istio sidecar envoy will take effect.

#WHY: We want to enable Istio as the service mesh to provide valuable metrics to Cloud Monitoring. 

#HOW: Created bash script to download the proper version of Istio and manage the installation. istio_operator.yaml contains the Operator file which dictates the parts of Istio that are configured. 

#TESTING: Makefile passes.

Addresses Issue #50


terraform/install_istio.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Download Istio
+
+echo "### "
+echo "### Begin install istio control plane - ${CONTEXT}"
+echo "### "
+
+# Set vars for DIRs
+ISTIO_VERSION="${ISTIO_VERSION:-1.6.2}"
+export WORK_DIR=`pwd`
```
Daniel-Sanche, line 26: maybe we should be more specific about where we're installing this stuff. What directory will this be running in?

terraform/install_istio.sh
```diff
@@ -0,0 +1,66 @@
...
+    --from-file=samples/certs/root-cert.pem \
+    --from-file=samples/certs/cert-chain.pem
+
+# cd back into install/
+cd ../
+
+kubectl label namespace default istio-injection=enabled
+kubectl create clusterrolebinding cluster-admin-binding \
+    --clusterrole=cluster-admin \
+    --user=$(gcloud config get-value core/account)
```
Daniel-Sanche, line 57: are we sure we can rely on gcloud here? Is the user account stored somewhere else in terraform already?

nicklin37: The user account is stored in terraform as the billing account ID. We could pass this in as an argument to the shell script. But this also means the user wouldn't be able to run the script alone anymore too (if there was ever a use case for that). Is there a reason we can't rely on the gcloud command?

terraform/install_istio.sh
nicklin37: Added comment to address this.

terraform/install_istio.sh
Daniel-Sanche: If terraform was already holding a copy the user id somewhere, I'd rather use that instead of pulling it out of the environment, just to reduce the reliance on other tools wherever we can. But it sounds like that's not the case, so that's fine.

nicklin37: Okay that makes sense. In addition, if one of our long term goals is to remove the requirement of a billing account from creating the sandbox, then the Terraform wouldn't contain any references to an account.

terraform/install_istio.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# install using operator config - https://istio.io/docs/setup/install/istioctl/#customizing-the-configuration
+INSTALL_PROFILE="./istio_operator.yaml"
+${WORK_DIR}/istioctl manifest apply -f ${INSTALL_PROFILE}
+
+# redeploy the pods
+cd ../release
+kubectl delete -f kubernetes-manifests.yaml
+kubectl apply -f kubernetes-manifests.yaml
```
Daniel-Sanche, line 64: We should actually put this Istio step before the original deploy_services step. Two reasons:
- that would remove the need for this delete and re-deploy
- the original deploy_services was followed by a delay to "wait a few sec to ensure kubectl apply completes". This re-deploy has no delay, so the user will see a cluster still being spun up

simonz130:
We also need to update the CI with installing istio step

Daniel-Sanche:
> We need to update the CI with installing istio

Do you think that should be worked on as a separate PR? I don't want this one to end up blocked for too long, and I don't think this work necessarily depends on a CI upgrade

nicklin37:
> > We need to update the CI with installing istio
> 
> Do you think that should be worked on as a separate PR? I don't want this one to end up blocked for too long, and I don't think this work necessarily depends on a CI upgrade
> 
> We might have to do some planning if we want to add Istio to the CI system. We currently spin up a fresh local cluster each time. If we and an Istio installation step, that may add 2-10 minutes to each run (I haven't tested). If it makes the tests too slow, we may want to consider spinning up persistent clusters on GKE instead.
> 
> I think these decisions probably go beyond the scope of this PR

Sounds good. The Istio installation should only take ~ 2 minutes from my experience running it. I've held off on adding it to the CI for now, but I addressed all the previous todos.

terraform/install_istio.sh
```diff
@@ -0,0 +1,63 @@
...
+echo "### "
+
+# Set vars for DIRs
+ISTIO_VERSION="${ISTIO_VERSION:-1.6.2}"
+
+# Set the working directory to our current directory (/sandbox/terraform)
+export WORK_DIR=`pwd`
+
+# Install Istio on ${CONTEXT}
+kubectx ${CONTEXT}
```
simonz130, line 31: where do we install kubectx?

terraform/install_istio.sh
```diff
@@ -0,0 +1,63 @@
...
+# Install Istio on the single cluster for Stackdriver Sandbox
+
+# Download Istio
+
+echo "### "
+echo "### Begin install istio control plane - ${CONTEXT}"
+echo "### "
+
+# Set vars for DIRs
+ISTIO_VERSION="${ISTIO_VERSION:-1.6.2}"
```
simonz130, line 25: There should be also an update in general terraform script or install.sh - wherever we determine cluster's external IP - now that it's istio that defines external IP

terraform/install_istio.sh
Daniel-Sanche: It's actually a part of the standard Cloud Shell image (I just tested with gcr.io/cloudshell-images/cloudshell to verify)

nicklin37: We don't actually need kubectx though since we're single cluster. I'm going to remove it soon.

terraform/install_istio.sh
nicklin37: This sounds good. I've made this change.

terraform/istio_operator.yaml
```diff
@@ -0,0 +1,435 @@
+# Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
nicklin37: This is a good idea. I've moved it into an istio folder inside terraform.

terraform/install_istio.sh
nicklin37: I've updated the install script to use the istio exposed external IP.

nicklin37:
> Hmm, I'm still seeing a couple issues when I run it on my GCP account:
> 
> * It doesn't give me a load generator url (it says" Load generator web interface: [not found]")
> * the cluster is unreachable when the scripts end. It says "The cluster is being updated" on GCP, and kubectl commands say "the connection to the server was refused"
> * when the cluster does eventualy spin up, the cartservice is in a crashloop. It looks like it's having trouble talking to redis. I wasn't seeing this last time. I noticed you deleted some of the contents of the istio-manifests directiory, any specific reason?
> * you might want to change the getExternalIp function of install.sh to use the istio url instead of the loadgenerator

Thanks Dan. I've tried to address these as best I can. I've put responses to these issues.

1. I think the load generator issue is separate from this PR - running both with and without Istio (occasionally) produces the load generator as being unavailable. 
2. I've slept at the end of istio install and increased the sleep at the end of deploying the services to give more time for it to come up. At the end of the script the External IP is available for me now.
3. Cartservice is no longer in a crashloop for me. I replaced the istio-manifests since it was recommended to utilize the subset for the newer version of Istio we are using.
4. Changed :)

Please let me know if this is working better.

Daniel-Sanche:
A couple more comments:
- **We need to increase the size of the cluster now that we've added istio**. We spin up 5 by default, but we need at least 6 to hold everything. We [usually recommend](https://cloud.google.com/istio/docs/istio-on-gke/installing#command-line) at least 8 vCPUs for Istio. I think a lot of flakiness will go away if we start with a large enough cluster
- **try to avoid sleep statements whenever possible**. Do you know which resources we need to wait on before proceeding? Is it possible to do a `kubectl wait` or ping a host before moving on to the next step? Arbitrary waits are a recipe for flakiness. If you can tell me which resources need to stabalize, I can try to help you pin down how to check against them. It's also possible fixing point 1 will remove the need for the long sleeps though

terraform/03_gke_cluster.tf
```diff
@@ -157,7 +157,7 @@ resource "null_resource" "deploy_services" {
 # For the purposes of Sandbox, we can mitigate by waiting a few sec to ensure kubectl apply completes
```
Daniel-Sanche, line 157: @simonz130 do you have the context for this comment? Is there a reason we can't `kubectl wait`?

nicklin37:
> A couple more comments:
> 
> * **We need to increase the size of the cluster now that we've added istio**. We spin up 5 by default, but we need at least 6 to hold everything. We [usually recommend](https://cloud.google.com/istio/docs/istio-on-gke/installing#command-line) at least 8 vCPUs for Istio. I think a lot of flakiness will go away if we start with a large enough cluster
> * **try to avoid sleep statements whenever possible**. Do you know which resources we need to wait on before proceeding? Is it possible to do a `kubectl wait` or ping a host before moving on to the next step? Arbitrary waits are a recipe for flakiness. If you can tell me which resources need to stabalize, I can try to help you pin down how to check against them. It's also possible fixing point 1 will remove the need for the long sleeps though

1. Sounds good. I"ll move the number of nodes to 8.
2. I inserted a sleep statements after applying the istio manifests. I think it helps the Istio configuration apply the manifests properly before moving to deploy. I'm not 100% certain, but I did notice the crash loop issue disappeared when I did this.

nicklin37:
I updated the nodes to be 7 since the loadgenerator needs 1 for itself. This removed the need for the additional sleep.

Daniel-Sanche:
Try replacing that sleep block I commented on above with:

```
        kubectl wait --for=condition=available --timeout=600s deployment/adservice
        kubectl wait --for=condition=available --timeout=600s deployment/cartservice
        kubectl wait --for=condition=available --timeout=600s deployment/checkoutservice
        kubectl wait --for=condition=available --timeout=600s deployment/currencyservice
        kubectl wait --for=condition=available --timeout=600s deployment/emailservice
        kubectl wait --for=condition=available --timeout=600s deployment/frontend
        kubectl wait --for=condition=available --timeout=600s deployment/paymentservice
        kubectl wait --for=condition=available --timeout=600s deployment/productcatalogservice
        kubectl wait --for=condition=available --timeout=600s deployment/recommendationse
```

That worked for me, and fixed all loadgenerator flakiness. After that, I think it should be good!

terraform/istio/install_istio.sh
```diff
@@ -0,0 +1,60 @@
...
+
+cd ./istio-${ISTIO_VERSION}/
+kubectl create secret generic cacerts -n istio-system \
+    --from-file=samples/certs/ca-cert.pem \
+    --from-file=samples/certs/ca-key.pem \
+    --from-file=samples/certs/root-cert.pem \
+    --from-file=samples/certs/cert-chain.pem
+
+# cd back into istio/
+cd ../
```
Daniel-Sanche, line 47: nit: these relative cd commands make me nervous. You should try to avoid them in scripts. Is this the same as `cd $WORK_DIR`?

terraform/istio/install_istio.sh
```diff
@@ -0,0 +1,60 @@
...
+kubectl create clusterrolebinding cluster-admin-binding \
+    --clusterrole=cluster-admin \
+    --user=$(gcloud config get-value core/account)
+
+# install using operator config - https://istio.io/docs/setup/install/istioctl/#customizing-the-configuration
+INSTALL_PROFILE="./istio_operator.yaml"
+${WORK_DIR}/istioctl manifest apply -f ${INSTALL_PROFILE}
+
+# apply manifests
+cd ../../
```
Daniel-Sanche, line 59: same as above, can this be made an explicit path? Or maybe just `kubectl apply -f $WORK_DIR/../istio-manifests`?

terraform/istio/install_istio.sh
Daniel-Sanche: This could also be a good place to use pushd and popd. Whatever is more readable

terraform/istio/install_istio.sh
```diff
@@ -0,0 +1,60 @@
...
+# Download Istio
+
+echo "### "
+echo "### Begin install istio control plane"
+echo "### "
+
+# Set vars for DIRs
+ISTIO_VERSION="${ISTIO_VERSION:-1.6.2}"
+
+# Set the working directory to our current directory (/sandbox/terraform)
```
Daniel-Sanche, line 27: is this comment still correct? It looks like the script is now in `/terraform/istio`

terraform/istio/install_istio.sh
```diff
@@ -0,0 +1,60 @@
...
+export WORK_DIR=`pwd`
```
Daniel-Sanche, line 28: Scripts can be run from anywhere, we may not be in the directory you expect to be. It's a good practice to capture the location of the script itself, and then move into that directory to do your work.

There's some discussion about the best way to do that [here](https://stackoverflow.com/questions/4774054/reliable-way-for-a-bash-script-to-get-the-full-path-to-itself/4774063), but I believe this should work

```
SCRIPTPATH=$(dirname $(realpath $0))
cd $SCRIPTPATH
export WORKDIR = `pwd`
```

terraform/istio/install_istio.sh
nicklin37: Thanks! I'm using WORK_DIR as much as possible.

terraform/istio/install_istio.sh
nicklin37: Changed :)

terraform/istio/install_istio.sh
nicklin37: Changed :)

terraform/istio/install_istio.sh
nicklin37: Nice this makes the script a lot better. Changed :)

nicklin37:
> Try replacing that sleep block I commented on above with:
> 
> ```
>         kubectl wait --for=condition=available --timeout=600s deployment/adservice
>         kubectl wait --for=condition=available --timeout=600s deployment/cartservice
>         kubectl wait --for=condition=available --timeout=600s deployment/checkoutservice
>         kubectl wait --for=condition=available --timeout=600s deployment/currencyservice
>         kubectl wait --for=condition=available --timeout=600s deployment/emailservice
>         kubectl wait --for=condition=available --timeout=600s deployment/frontend
>         kubectl wait --for=condition=available --timeout=600s deployment/paymentservice
>         kubectl wait --for=condition=available --timeout=600s deployment/productcatalogservice
>         kubectl wait --for=condition=available --timeout=600s deployment/recommendationse
> ```
> 
> That worked for me, and fixed all loadgenerator flakiness. After that, I think it should be good!

Thanks! This fixed the flakiness and got rid of those sleep commands! Thanks for the help!! :)

terraform/03_gke_cluster.tf
```diff
@@ -135,20 +135,28 @@ resource "null_resource" "set_gke_context" {
...
+  depends_on = ["null_resource.install_istio"]
 }
 
-# There is no reliable way to do deployment verification with kubernetes
-# For the purposes of Sandbox, we can mitigate by waiting a few sec to ensure kubectl apply completes
+# We wait for all of our microservices to become available on kubernetes
 resource "null_resource" "delay" {
   provisioner "local-exec" {
-    command = "sleep 5"
+    command = "kubectl wait --for=condition=available --timeout=600s deployment/adservice && kubectl wait --for=condition=available --timeout=600s deployment/cartservice && kubectl wait --for=condition=available --timeout=600s deployment/checkoutservice && kubectl wait --for=condition=available --timeout=600s deployment/currencyservice && kubectl wait --for=condition=available --timeout=600s deployment/emailservice && kubectl wait --for=condition=available --timeout=600s deployment/frontend && kubectl wait --for=condition=available --timeout=600s deployment/paymentservice && kubectl wait --for=condition=available --timeout=600s deployment/productcatalogservice && kubectl wait --for=condition=available --timeout=600s deployment/recommendationservice"
```
simonz130, line 159: Can this be split into multiple lines for clarity?

terraform/istio/install_istio.sh
```diff
@@ -24,8 +24,12 @@ echo "### "
 # Set vars for DIRs
 ISTIO_VERSION="${ISTIO_VERSION:-1.6.2}"
 
-# Set the working directory to our current directory (/sandbox/terraform)
-export WORK_DIR=`pwd`
+# Set the working directory to our current directory (/sandbox/terraform/istio)
+export SCRIPTPATH=$(dirname $(realpath $0))
+cd $SCRIPTPATH
+# WORKDIR and SCRIPTPATH should be the same in our case, 
+# but you may find it more readable to keep both variables around
```
Daniel-Sanche, line 31: Not a big deal, but I'd probably leave this comment out. I put that in for you, not necessarily for the script

terraform/istio/install_istio.sh
```diff
@@ -44,17 +48,16 @@ kubectl create secret generic cacerts -n istio-system \
     --from-file=samples/certs/cert-chain.pem
 
 # cd back into istio/
-cd ../
+cd ${WORD_DIR}
```
Daniel-Sanche, line 51: looks like a typo on WORK_DIR

terraform/istio/install_istio.sh
```diff
@@ -44,17 +48,16 @@ kubectl create secret generic cacerts -n istio-system \
...
 
 kubectl label namespace default istio-injection=enabled
 kubectl create clusterrolebinding cluster-admin-binding \
     --clusterrole=cluster-admin \
     --user=$(gcloud config get-value core/account)
 
 # install using operator config - https://istio.io/docs/setup/install/istioctl/#customizing-the-configuration
-INSTALL_PROFILE="./istio_operator.yaml"
+INSTALL_PROFILE="${WORK_DIR}/istio_operator.yaml"
 ${WORK_DIR}/istioctl manifest apply -f ${INSTALL_PROFILE}
```
Daniel-Sanche, line 60: nit, but you probably don't need INSTALL_PROFILE as a variable. You could just do:

`${WORK_DIR}/istioctl manifest apply -f ${WORK_DIR}/istio_operator.yaml`

terraform/03_gke_cluster.tf
Daniel-Sanche: @nicklin37 I just checked and this is also valid to cut down on the size:

`kubectl wait deployment --for=condition=available --timeout=600s adservice cartservice checkoutservice currencyservice emailservice frontend paymentservice productcatalogservice recommendationservice shippingservice`

That's still pretty long though, so breaking it up would be ideal if that's possible

terraform/istio/install_istio.sh
nicklin37: Good idea. Changed :)

terraform/istio/install_istio.sh
nicklin37: Thanks for catching that - Technically this line is unneeded but it feels safer to come back to our workdirectory after leaving.

terraform/istio/install_istio.sh
nicklin37: Haha woops I didn't realize that. Changed :)

terraform/03_gke_cluster.tf
nicklin37: Thanks Terraform really only likes single lines for quoted strings so I'm using an EOT  'heredoc' style. It causes added delimiters but it breaks it up. Let me know if that's better.

terraform/03_gke_cluster.tf
simonz130: I think we should kubectl wait :)

PR #167, opened 2020-07-21
Refactor: remove customize script

Daniel-Sanche:
First part of repo clean up in advance of adding release process (https://github.com/GoogleCloudPlatform/stackdriver-sandbox/issues/114)

Deploy kubernetes manifests directly, instead of relying on extra script to re-write the images. We now rely on [skaffold's image rewriting functionality](https://skaffold.dev/docs/environment/image-registries/) instead.

Benefits of this change:
- users can now `kubectl apply` the kubernetes manifests directly to test the HEAD of the repo without building
- official container paths/versions are more obvious. They can be found directly in the manifests, instead of being hidden away in an obscure script

Tested manually in my personal project


kubernetes-manifests/adservice.yaml
```diff
@@ -28,7 +28,7 @@ spec:
       terminationGracePeriodSeconds: 5
       containers:
       - name: server
-        image: adservice
+        image: gcr.io/stackdriver-sandbox-230822/sandbox/adservice:latest
```
simonz130, line 31: Hardcoding the image (even if we use latest tag here) means that if we want to test things locally, we'll have to update the image path manually in all the microservices yaml files. Why not using skaffold profiles feature where we could just specify the image prefix once (--default-repo)  and not in 10 places?

Daniel-Sanche: This is actually the way skaffold is intended to work ([source](https://skaffold.dev/docs/environment/image-registries/)). It will rewrite gcr.io/stackdriver-sandbox-230822/sandbox/adservice:latest to $DEFAULT_REPO/gcr.io/stackdriver-sandbox-230822/sandbox/adservice:$hash

The re-written path is a bit ugly, but I think we're better off following skaffold's best practices instead of writing custom scripts to do rewriting ourselves

PR #173, opened 2020-07-23
upgrade terraform version

nicklin37:
#WHAT: Update the version of terraform that's installed to be the latest version

#WHY: The current version is slightly too old and doesn't support the dashboard resource

#HOW: Changed which version of terraform we install.

#TESTING: install.sh script still works the same, terraform apply will now work in monitoring directory without manual reinstall of proper version.

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -60,7 +60,7 @@ getBillingAccount() {
 
 installTerraform() {
   sudo apt-get install unzip
-  wget -q https://releases.hashicorp.com/terraform/0.11.11/terraform_0.11.11_linux_amd64.zip -O ./terraform.zip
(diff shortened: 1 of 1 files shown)


terraform/install.sh
```diff
@@ -60,7 +60,7 @@ getBillingAccount() {
 
 installTerraform() {
   sudo apt-get install unzip
-  wget -q https://releases.hashicorp.com/terraform/0.11.11/terraform_0.11.11_linux_amd64.zip -O ./terraform.zip
+  wget -q https://releases.hashicorp.com/terraform/0.12.29/terraform_0.12.29_linux_amd64.zip -O ./terraform.zip
```
simonz130, line 63: @eaberman, FYI - we might want to move installing terraform into your custom cloud shell image.

PR #175, opened 2020-07-23
frontend dashboards

### LINKED ISSUE #61: Create Terraform for frontend service custom dashboard
#WHAT: Creates dashboard representing four golden signals of monitoring (latency, traffic, errors, and saturation) for the frontend service

#WHY: Creates opinionated monitoring of this service

#HOW: Terraform resources added to initialization terraform scripts

nicklin37:
#WHAT: Frontend Service Dashboards using Istoi

#WHY: We want to use Istio to demonstrate the four golden signals for the frontend service

#HOW: Terraform resource

#TESTING: Integration test for resource creation passes.

Closes #61

### DIFF
--- a/terraform/monitoring/02_dashboards.tf
+++ b/terraform/monitoring/02_dashboards.tf
@@ -14,104 +14,15 @@
 
 # Here we create a dashboard for the general user experience - it contains metrics
 # that reflect how users are interacting with the demo application such as latency
-# of requests, distribution of types of requests, and response types.
+# of requests, distribution of types of requests, and response types. The JSON object
+# containing the exact details of the dashboard can be found in the 'dashboards' folder.
 resource "google_monitoring_dashboard" "userexp_dashboard" {
-	dashboard_json = <<EOF
-	{
-  "displayName": "User Experience Dashboard",
-  "gridLayout": {
-    "columns": "2",
-    "widgets": [
-      {
-        "title": "HTTP Request Count by Method",
-        "xyChart": {
-          "dataSets": [
-            {
-              "timeSeriesQuery": {
-                "timeSeriesFilter": {
-                  "filter": "metric.type=\"custom.googleapis.com/opencensus/opencensus.io/http/server/request_count_by_method\"",
-                  "aggregation": {
-                    "perSeriesAligner": "ALIGN_RATE"
-                  },
-                  "secondaryAggregation": {}
-                },
-                "unitOverride": "1"
-              },
-              "plotType": "LINE",
-              "minAlignmentPeriod": "60s"
-            }
-          ],
-          "timeshiftDuration": "0s",
-          "yAxis": {
-            "label": "y1Axis",
-            "scale": "LINEAR"
-          },
-          "chartOptions": {
-            "mode": "COLOR"
-          }
-        }
-      },
-      {
-        "title": "HTTP Response Errors",
-        "xyChart": {
-          "dataSets": [
-            {
-              "timeSeriesQuery": {
-                "timeSeriesFilter": {
-                  "filter": "metric.type=\"custom.googleapis.com/opencensus/opencensus.io/http/server/response_count_by_status_code\" metric.label.\"http_status\"!=\"200\"",
-                  "aggregation": {
-                    "perSeriesAligner": "ALIGN_RATE"
-                  },
-                  "secondaryAggregation": {}
-                },
-                "unitOverride": "1"
-              },
-              "plotType": "LINE",
-              "minAlignmentPeriod": "60s"
-            }
-          ],
-          "timeshiftDuration": "0s",
-          "yAxis": {
-            "label": "y1Axis",
-            "scale": "LINEAR"
-          },
-          "chartOptions": {
-            "mode": "COLOR"
-          }
-        }
-      },
-      {
-        "title": "HTTP Request Latency",
-        "xyChart": {
-          "dataSets": [
-            {
-              "timeSeriesQuery": {
-                "timeSeriesFilter": {
-                  "filter": "metric.type=\"custom.googleapis.com/opencensus/opencensus.io/http/server/latency\"",
-                  "aggregation": {
-                    "perSeriesAligner": "ALIGN_DELTA",
-                    "crossSeriesReducer": "REDUCE_PERCENTILE_99"
-                  },
-                  "secondaryAggregation": {}
-                },
-                "unitOverride": "ms"
-              },
-              "plotType": "LINE",
-              "minAlignmentPeriod": "60s"
-            }
-          ],
-          "timeshiftDuration": "0s",
-          "yAxis": {
-            "label": "y1Axis",
-            "scale": "LINEAR"
-          },
-          "chartOptions": {
-            "mode": "COLOR"
-          }
-        }
-      }
-    ]
-  }
+	dashboard_json = file("./dashboards/userexp_dashboard.json")
 }
-EOF
-}
+
+# This resource creates dashboards for the frontend service of the
+# Stackdriver Sandbox. The JSON object containing the exact details
+# of the dashboard can be found in the 'dashboards' folder.
+resource "google_monitoring_dashboard" "frontend_dashboard" {
+  dashboard_json = file("./dashboards/frontend_dashboard.json")
+} 
--- a/terraform/monitoring/dashboards/frontend_dashboard.json
+++ b/terraform/monitoring/dashboards/frontend_dashboard.json
@@ -1,0 +1,116 @@
+{
+  "displayName": "Frontend Service Dashboard",
+  "gridLayout": {
+    "columns": "2",
+    "widgets": [
+      {
+        "title": "CPU Limit Utilization - Frontend Service",
+        "xyChart": {
+          "dataSets": [
+            {
+              "timeSeriesQuery": {
+                "timeSeriesFilter": {
+                  "filter": "metric.type=\"kubernetes.io/container/cpu/limit_utilization\" resource.type=\"k8s_container\" resource.label.\"cluster_name\"=\"stackdriver-sandbox\" metadata.user_labels.\"app\"=\"frontend\"",
+                  "aggregation": {
+                    "perSeriesAligner": "ALIGN_MEAN"
+                  }
+                },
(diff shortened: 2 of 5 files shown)


terraform/monitoring/dashboards/frontend_dashboard.json
```diff
@@ -0,0 +1,153 @@
...
+        "xyChart": {
+          "dataSets": [
+            {
+              "timeSeriesQuery": {
+                "timeSeriesFilter": {
+                  "filter": "metric.type=\"kubernetes.io/container/cpu/limit_utilization\" resource.type=\"k8s_container\" resource.label.\"cluster_name\"=\"stackdriver-sandbox\" metadata.user_labels.\"app\"=\"frontend\"",
+                  "aggregation": {
+                    "perSeriesAligner": "ALIGN_MEAN"
+                  },
+                  "secondaryAggregation": {}
```
eliotw, line 17: Did you create these dashboards via the UI, and then download the JSON? There is an unfortunate bug where the UI is writing empty secondaryAggregation objects and empty timeshiftDuration fields.

Since this is an example dashboard I think we should try to keep it as clean as possible. I would recommend removing all empty secondaryAggregation objects and timeshiftDuration set to "0s".

terraform/monitoring/dashboards/frontend_dashboard.json
```diff
@@ -0,0 +1,153 @@
...
+                },
+                "unitOverride": "1"
+              },
+              "plotType": "LINE",
+              "minAlignmentPeriod": "60s"
+            }
+          ],
+          "timeshiftDuration": "0s",
```
eliotw, line 25: See previous comment about removing all default ("0s") instances of this.

terraform/monitoring/variables.tf
```diff
@@ -19,12 +19,12 @@
...
 variable "external_ip" {
   type        = "string"
-  description = "The external IP of the kubernetes cluster. Can be revealed by running kubectl get service frontend-external in Cloud CLI."
+  description = "The external IP of the kubernetes cluster. Can be revealed by running \"kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'\" in the Google Cloud CLI."
 }
 
 variable "project_id" {
   type        = "string"
-  description = "The project id that was created by Stackdriver Sandbox."
+  description = "The project id that was created by Stackdriver Sandbox.  Can be revealed by running \"gcloud config get-value project\" in the Google Cloud CLI."
```
eliotw, line 27: Nit: Extra space here.

terraform/monitoring/dashboards/frontend_dashboard.json
nicklin37: Yeah I retrieved them with the API. Good idea - I pruned away all the empty secondaryAggregations and timeShiftDuration="0s". Thanks for catching that.

terraform/monitoring/dashboards/frontend_dashboard.json
nicklin37: Changed :)

terraform/monitoring/variables.tf
nicklin37: Oops! Thanks, changed :)

terraform/monitoring/dashboards/frontend_dashboard.json
```diff
@@ -0,0 +1,143 @@
...
+            "label": "y1Axis",
+            "scale": "LINEAR"
+          },
+          "chartOptions": {
+            "mode": "COLOR"
+          }
+        }
+      },
+      {
+        "title": "Server Request Count for stackdriver-sandbox, frontend",
```
simonz130, line 34: remove "for stackdriver-sandbox".

terraform/monitoring/dashboards/frontend_dashboard.json
```diff
@@ -0,0 +1,143 @@
...
+            "label": "y1Axis",
+            "scale": "LINEAR"
+          },
+          "chartOptions": {
+            "mode": "COLOR"
+          }
+        }
+      },
+      {
+        "title": "Server Request Bytes for stackdriver-sandbox, frontend",
```
simonz130, line 61: remove "for stackdriver-sandbox" from everywhere.
Instead, the title should mention the microservice name and the metric, like this:

"Server Requests Bytes - Frontend"

tests/monitoring_integration_test.py
```diff
@@ -95,16 +95,23 @@ def testUptimeCheckAlertingPolicyNotificationChannel(self):
...
+
+	def testUserExpDashboard(self):
+		""" Test that the User Experience Dashboard gets created. """
+		found_dash = self.checkForDashboard('User Experience Dashboard')
+		self.assertTrue(found_dash)
+
+	def testFrontendServiceDashboard(self):
+		""" Test that the Frontend Service Dashboard gets created. """
+		found_dash = self.checkForDashboard('Frontend Service Dashboard')
+		self.assertTrue(found_dash)
```
simonz130, line 114: Are these tests added to run as part of each CI run?

terraform/monitoring/dashboards/frontend_dashboard.json
nicklin37: Changed :)

terraform/monitoring/dashboards/frontend_dashboard.json
nicklin37: Changed :)

tests/monitoring_integration_test.py
nicklin37: They're not - not sure if you saw the conversation with Dan in the chat room but basically we think that setting up the testing environment will take significant changes that might be better to be added in the next sprint/release. The tests are still run manually but they are passing.

terraform/monitoring/02_dashboards.tf
```diff
@@ -114,4 +114,11 @@ resource "google_monitoring_dashboard" "userexp_dashboard" {
...
 EOF
-}
\ No newline at end of file
+}
+
+# This resource creates dashboards for the frontend service of the
+# Stackdriver Sandbox. The JSON object containing the exact details
+# of the dashboard can be found in the 'dashboards' folder.
+resource "google_monitoring_dashboard" "frontend_dashboard" {
+  dashboard_json = file("./dashboards/frontend_dashboard.json")
```
briancorwin, line 123: I might suggest moving the userexp_dashboard to it's own file as well.  You should be consistent on either doing it inline or with files.

terraform/monitoring/dashboards/frontend_dashboard.json
```diff
@@ -0,0 +1,143 @@
...
+            "label": "y1Axis",
+            "scale": "LINEAR"
+          },
+          "chartOptions": {
+            "mode": "COLOR"
+          }
+        }
+      },
+      {
+        "title": "Server Request Bytes - Frontend Service",
```
briancorwin, line 61: Generally I've found this is not an incredibly useful thing to use especially if the requests are consistent in size.  If the endpoint accepts potentially large objects of highly varying size (like images) then request counts is not good enough.  In can't imagine that's the case here.

terraform/monitoring/dashboards/frontend_dashboard.json
```diff
@@ -0,0 +1,143 @@
...
+            "label": "y1Axis",
+            "scale": "LINEAR"
+          },
+          "chartOptions": {
+            "mode": "COLOR"
+          }
+        }
+      },
+      {
+        "title": "Server Request Count filtered by non-200 Responses - Frontend Service",
```
briancorwin, line 88: We just call this "errors" not "non-200".  Everyone generally knows what the means wrt response codes.

terraform/monitoring/dashboards/frontend_dashboard.json
```diff
@@ -0,0 +1,143 @@
...
+        "xyChart": {
+          "dataSets": [
+            {
+              "timeSeriesQuery": {
+                "timeSeriesFilter": {
+                  "filter": "metric.type=\"istio.io/service/server/request_count\" resource.type=\"k8s_container\" resource.label.\"cluster_name\"=\"stackdriver-sandbox\" metadata.user_labels.\"app\"=\"frontend\" metric.label.\"response_code\"!=\"200\"",
```
briancorwin, line 94: One thing that might be interesting here is to use MQL to do a ratio of the errors to the full request count to get an error ratio.  This would be a good bonus to add afterwards.

terraform/monitoring/dashboards/frontend_dashboard.json
```diff
@@ -0,0 +1,143 @@
...
+      {
+        "title": "Server Response Latencies - Frontend Service",
+        "xyChart": {
+          "dataSets": [
+            {
+              "timeSeriesQuery": {
+                "timeSeriesFilter": {
+                  "filter": "metric.type=\"istio.io/service/server/response_latencies\" resource.type=\"k8s_container\" resource.label.\"cluster_name\"=\"stackdriver-sandbox\" metadata.user_labels.\"app\"=\"frontend\"",
+                  "aggregation": {
+                    "perSeriesAligner": "ALIGN_PERCENTILE_99"
```
briancorwin, line 123: You should probably put the percentile in the title.

nicklin37: Changed :)

terraform/monitoring/dashboards/frontend_dashboard.json
nicklin37: Gotcha - Changed :)

terraform/monitoring/dashboards/frontend_dashboard.json
nicklin37: Yeah I agree here - I've removed the chart since it's not adding any additional information.

terraform/monitoring/02_dashboards.tf
nicklin37: Changed :)

terraform/monitoring/dashboards/frontend_dashboard.json
nicklin37: This is a good idea. I'll make note of it to add later.

PR #186, opened 2020-07-24
Adding google.com accounts support and terraform stateful provisioning

jiang95-dev:
1. Implement the project creation process: create a project and a bucket and link the billing account.
2. Implement the project choosing process: let user to choose an existing project or create a new one.
3. Switch the Terraform backend from local to remote: Interpolate the bucket name in the backend config.

Update:
1. Add support to google.com account: projects should now be in /experimental-gke.
2. Change the way to get the billing ID: use a map to store (name:id) to avoid fuzzy search supported by the gcloud filter flag.
3. Remove code remarks and make code clean.

Update2:
1. Change the way to ensure bucket is created: put it in a retry loop.
2. Now it only fetches the projects under certain billing account.
3. Shows project creation time on the prompt.

### DIFF
--- a/terraform/00_state.tf
+++ b/terraform/00_state.tf
@@ -16,22 +16,28 @@
 # backend which stores everything in a local directory. The config you see below
 # is functionally equivalent to the defaults.
 
-terraform {
-  backend "local" {
-    path = "terraform.tfstate"
-  }
-}
+#terraform {
+#  backend "local" {
+#    path = "terraform.tfstate"
+#  }
+#}
 
 # TODO - Consider storing state remotely on a per-instance basis. 
 # In that case we'd generate a backend config just before
 # running terraform that would look something like this:
 
-# terraform {
-#   backend "gcs" {
-#     bucket = "stackdriver-sandbox"
-#     prefix = "<project-id>"
-#   }
-# }
+terraform {
+   backend "gcs" {
+       bucket="stackdriver-sandbox-1882647097-bucket"
+   }
+}
+
+data "terraform_remote_state" "state" {
+    backend = "gcs"
+    config = {
+        bucket = "${var.bucket_name}"
+    }
+}
 
 # Interpolations are not supported in backend configs so we'd have to generate
 # the file rather than rely on env vars like we can do almost everywhere else.
--- a/terraform/00_state.tf
+++ b/terraform/00_state.tf
@@ -27,9 +27,7 @@
 # running terraform that would look something like this:
 
 terraform {
-   backend "gcs" {
-       bucket="stackdriver-sandbox-1882647097-bucket"
-   }
+   backend "gcs" {}
 }
 
 data "terraform_remote_state" "state" {
--- a/terraform/00_state.tf
+++ b/terraform/00_state.tf
@@ -12,6 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+# We are now using gcs as our backend, so the state file will be stored
+# in a storage bucket. Since the bucket must preexists, we will create 
+# the project and bucket outside Terraform. Also since the configuration
+# of bucket can't be a variable, we create an empty config and modify it
+# in the data section.
+
 terraform {
    backend "gcs" {}
 }
--- a/terraform/00_state.tf
+++ b/terraform/00_state.tf
@@ -12,20 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Here we configure the state backend. For this demo we're using the "local"
-# backend which stores everything in a local directory. The config you see below
-# is functionally equivalent to the defaults.
-
-#terraform {
-#  backend "local" {
-#    path = "terraform.tfstate"
-#  }
-#}
-
-# TODO - Consider storing state remotely on a per-instance basis. 
-# In that case we'd generate a backend config just before
-# running terraform that would look something like this:
-
 terraform {
    backend "gcs" {}
 }
@@ -36,6 +22,3 @@ data "terraform_remote_state" "state" {
         bucket = "${var.bucket_name}"
     }
 }
-
-# Interpolations are not supported in backend configs so we'd have to generate
-# the file rather than rely on env vars like we can do almost everywhere else.
--- a/terraform/00_state.tf
+++ b/terraform/00_state.tf
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 
-# We are now using gcs as our backend, so the state file will be stored
+# We use gcs as our backend, so the state file will be stored
 # in a storage bucket. Since the bucket must preexists, we will create 
 # the project and bucket outside Terraform. Also since the configuration
 # of bucket can't be a variable, we create an empty config and modify it
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -86,7 +86,7 @@ getProject() {
       found_projects+=("$bill_proj | $create_time")
     fi
   done
-  
+
   if [ -z "$found_projects" ] || [[ ${#found_projects[@]} -eq 0 ]]; then
     createProject;
   else
@@ -123,11 +123,12 @@ createProject() {
     if [[ $acct == *"google.com"* ]];
     then
       log ""
-      log "Are you sure you have requested access to /experimental-gke folder?"
-      log "If not, please make a request at https://sphinx.corp.google.com/sphinx/#accessChangeRequest:systemName=internal_google_cloud_platform_usage"
+      log "Note: your project will be created in the /experimental-gke folder."
+      log "If you don't have access to this folder, please make sure to request at:"
+      log "https://sphinx.corp.google.com/sphinx/#accessChangeRequest:systemName=internal_google_cloud_platform_usage"
       log ""
-      select opt in "yes" "no"; do
-        if [[ "$opt" == "yes" ]]; then
+      select opt in "continue" "cancel"; do
+        if [[ "$opt" == "continue" ]]; then
           break;
         else
           exit 0;
@@ -235,7 +236,7 @@ log "Checking Prerequisites..."
 getBillingAccount;
 
 log "Install current version of Terraform"
-#installTerraform
+installTerraform
 
 # Make sure we use Application Default Credentials for authentication
 # For that we need to unset GOOGLE_APPLICATION_CREDENTIALS and generate
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -23,8 +23,6 @@ set -o errexit  # Exit on error
 SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
 
-create_flag=false # whether user wants to create a new project
-
 log() { echo "$1" >&2; }
 
 getBillingAccount() {
@@ -78,19 +76,19 @@ getBillingAccount() {
 
 getProject() {
   log "Checking for project list"
-  found_projects=$(gcloud projects list --filter="name:stackdriver-sandbox-*" --format="value(projectId)")
+  found_projects=$(gcloud projects list --filter="id:stackdriver-sandbox-* AND name='Stackdriver Sandbox Demo'" --format="value(projectId)")
   if [ -z "$found_projects" ] || [[ ${#found_projects[@]} -eq 0 ]]; then
-    create_flag=true
+    createProject;
   else
       log "Which project would you like to use?:"
       IFS_bak=$IFS
       IFS=$'\n'
-      select opt in ${found_projects} "create a new Sandbox" "cancel"; do
+      select opt in "create a new Sandbox" ${found_projects} "cancel"; do
         if [[ "${opt}" == "cancel" ]]; then
           exit 0
         elif [[ "${opt}" == "create a new Sandbox" ]]; then
-          create_flag=true
           log "create a new Sandbox!"
+          createProject;
           break
         elif [[ -z "${opt}" ]]; then
           log "invalid response"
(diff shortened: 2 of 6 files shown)


terraform/00_state.tf
```diff
@@ -16,22 +16,28 @@
...
-terraform {
-  backend "local" {
-    path = "terraform.tfstate"
-  }
-}
+#terraform {
+#  backend "local" {
+#    path = "terraform.tfstate"
+#  }
+#}
```
simonz130, line 23: Don't leave code remarks in your changelists

terraform/00_state.tf
```diff
@@ -16,22 +16,28 @@
...
 
-# terraform {
-#   backend "gcs" {
-#     bucket = "stackdriver-sandbox"
-#     prefix = "<project-id>"
-#   }
-# }
+terraform {
+   backend "gcs" {
+       bucket="stackdriver-sandbox-1882647097-bucket"
```
simonz130, line 31: What is the idea behind **1882647097** in the name of the bucket?

terraform/install.sh
```diff
@@ -23,6 +23,8 @@ set -o errexit  # Exit on error
 SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
 
+create_flag=false
```
simonz130, line 26: What is this? Please document the purpose of this variable

terraform/install.sh
```diff
@@ -58,6 +60,51 @@ getBillingAccount() {
...
 }
 
+getProject() {
+  log "Checking for project list"
+  found_projects=$(gcloud projects list --filter="name:stackdriver-sandbox-*" --format="value(projectId)")
+  if [ -z "$found_projects" ] || [[ ${#found_projects[@]} -eq 0 ]]; then
+    create_flag=true
+  else
+      log "Which project would you like to use?:"
+      IFS_bak=$IFS
```
simonz130, line 70: What is IFS?

terraform/install.sh
```diff
@@ -58,6 +60,51 @@ getBillingAccount() {
...
+          break
+        fi
+      done
+      IFS=$IFS_bak
+  fi
+}
+
+createProject() {
+  if [ "$create_flag" = true ]; then
+    # generate random id
```
simonz130, line 93: The degree of randomness should not exceed maximum allowed name length for project name. Did you check the restrictions on the project id?

terraform/install.sh
```diff
@@ -58,6 +60,51 @@ getBillingAccount() {
...
+    project_id="stackdriver-sandbox-$(od -N 4 -t uL -An /dev/urandom | tr -d " ")"
+    bucket_name="$project_id-bucket"
```
simonz130, line 95: I suggest that we name it so it's clear what's the purpose of the bucket. You don't need to include the projectid in the name since each bucket belongs to a project.
You can name it stackdriver-sandbox-state or something along those lines.

terraform/install.sh
```diff
@@ -58,6 +60,51 @@ getBillingAccount() {
...
+    # create project
+    gcloud projects create "$project_id" --name="Stackdriver Sandbox Demo"
+    # link billing account
+    billing_id=$(gcloud beta billing accounts list --format="value(name)" --filter="displayName:$billing_acct")
```
simonz130, line 99: What happens if the user has two or more billing accounts?

terraform/00_state.tf
jiang95-dev: Ok.

terraform/00_state.tf
jiang95-dev: This is a mistake. I removed it.

terraform/install.sh
jiang95-dev: It indicates whether user wants to create a project.

terraform/install.sh
jiang95-dev: It's a delimiter in shell script.

terraform/install.sh
jiang95-dev: The random number is 4 bytes, the same as previous.

terraform/install.sh
jiang95-dev: Ok.

terraform/install.sh
jiang95-dev: The filter is the account name user chose.

terraform/install.sh
```diff
@@ -37,25 +39,93 @@ getBillingAccount() {
...
       IFS=$'\n'
       select opt in ${found_accounts} "cancel"; do
         if [[ "${opt}" == "cancel" ]]; then
           exit 0
         elif [[ -z "${opt}" ]]; then
           log "invalid response"
         else
-          billing_acct=$opt
+          billing_acct=${opt}
+          billing_id=${map[$billing_acct]}
```
simonz130, line 68: will this always find an item in the map?

terraform/install.sh
```diff
@@ -150,15 +215,15 @@ installTerraform
 #export GOOGLE_APPLICATION_CREDENTIALS=""
 #gcloud auth application-default login
 
-# Ensure no google.com accounts early - they are not supported!
 acct=$(gcloud info --format="value(config.account)")
 if [[ $acct == *"google.com"* ]];
 then
-  log "Google.com accounts are currently not supported by Stackdriver Sandbox.";
-  exit;
+  folder_id="262044416022"
```
simonz130, line 221: We should also add instructions for Googlers:  
we should show them a prompt that asks them to apply for permissions (one off) to create projects under /experimental folder. In the prompt they need to acknowledge that they applied or cancel. If they acknowledged that they applied, we run the rest of the flow that creates the project under /experimental for google.com accounts.

terraform/install.sh
```diff
@@ -37,25 +39,93 @@ getBillingAccount() {
...
+    else
+      gcloud projects create "$project_id" --name="Stackdriver Sandbox Demo" --folder="$folder_id"
+    fi;
+    # link billing account
+    gcloud beta billing projects link "$project_id" --billing-account="$billing_id"
+    # confirm billing account linked
+    check_billing=$(gcloud beta billing projects list --billing-account="$billing_id" --format="value(project_id)" --filter="project_id:$project_id")
+    while [ -z "$check_billing" ]; do
+      log "waiting for billing account to be linked"
+      sleep 1
```
Daniel-Sanche, line 124: I'm a bit nervoud about leaving this potentially infinite loop here. Maybe you should add a TRIES variable like we use here: https://github.com/GoogleCloudPlatform/stackdriver-sandbox/blob/dab7ab519a04c874058f3b2dc0d552b345590d6e/terraform/install.sh#L100-L102

terraform/install.sh
Daniel-Sanche, line 221: add a comment explaining that this is the /experimental-gke folder

terraform/variables.tf
```diff
@@ -20,3 +20,13 @@ variable "billing_account" {
...
 }
+
+variable "project_id" {
+  type        = "string"
+  description = "The id of your project. Case-sensitive."
+}
+
+variable "bucket_name" {
+  type        = "string"
+  description = "The name of your bucket. Case-sensitive."
```
Daniel-Sanche, line 31: this could be improved. What is the bucket for?

terraform/install.sh
```diff
@@ -150,15 +215,15 @@ installTerraform
...
 if [[ $acct == *"google.com"* ]];
 then
-  log "Google.com accounts are currently not supported by Stackdriver Sandbox.";
-  exit;
+  folder_id="262044416022"            
 fi;
 
 # Provision Stackdriver Sandbox cluster
+getProject;
+createProject;
```
Daniel-Sanche, line 226: Having createProject here is a bit misleading, since a project is only created sometimes. Consider removing `create_flag` and putting `createProject` inside the logic of `getProject` instead

terraform/install.sh
jiang95-dev: Ok.

terraform/install.sh
```diff
@@ -37,25 +39,93 @@ getBillingAccount() {
...
   else
     billing_acct=${found_accounts}
+    billing_id=${map[$found_accounts]}
+  fi
+  IFS=$IFS_bak
+}
+
+getProject() {
+  log "Checking for project list"
+  found_projects=$(gcloud projects list --filter="name:stackdriver-sandbox-*" --format="value(projectId)")
```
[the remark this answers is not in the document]
jiang95-dev: Good idea!

terraform/install.sh
```diff
@@ -37,25 +39,93 @@ getBillingAccount() {
...
+  if [ -z "$found_projects" ] || [[ ${#found_projects[@]} -eq 0 ]]; then
+    create_flag=true
+  else
+      log "Which project would you like to use?:"
+      IFS_bak=$IFS
+      IFS=$'\n'
+      select opt in ${found_projects} "create a new Sandbox" "cancel"; do
```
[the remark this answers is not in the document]
jiang95-dev: Ok.

terraform/00_state.tf
```diff
@@ -12,26 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Here we configure the state backend. For this demo we're using the "local"
-# backend which stores everything in a local directory. The config you see below
-# is functionally equivalent to the defaults.
-
```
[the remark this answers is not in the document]
jiang95-dev: Ok.

terraform/install.sh
jiang95-dev: Good idea!

terraform/install.sh
jiang95-dev: Sure.

terraform/variables.tf
jiang95-dev: When you initialize the backend, you need to give it a bucket.

terraform/install.sh
jiang95-dev: It should be.

terraform/variables.tf
Daniel-Sanche: It might be helpful to say what the bucket is used for in the description I mean. I assume it's to hold the tfstate?

terraform/install.sh
simonz130: add a comment

terraform/install.sh
```diff
@@ -150,15 +215,15 @@ installTerraform
...
 #gcloud auth application-default login
 
-# Ensure no google.com accounts early - they are not supported!
 acct=$(gcloud info --format="value(config.account)")
 if [[ $acct == *"google.com"* ]];
 then
-  log "Google.com accounts are currently not supported by Stackdriver Sandbox.";
-  exit;
+  folder_id="262044416022"            
 fi;
```
Daniel-Sanche, line 222: I don't really like having this logic out here on its own. I think this should be moved into createProject, maybe like this:

```
acct=$(gcloud info --format="value(config.account)")
if [ $acct == *"google.com"* ]; then
  gcloud projects create "$project_id" --name="Stackdriver Sandbox Demo" --folder=262044416022
else
  gcloud projects create "$project_id" --name="Stackdriver Sandbox Demo"
fi;
```

Daniel-Sanche: Additionally, it would be a good idea to prefix Googler-created sandboxes with the user's ldap, so we can keep track of ownership:

```
ldap=$(echo $acct |  cut -f1 -d"@")
project_id=$ldap-$project_id
gcloud projects create "$project_id" --name="Stackdriver Sandbox Demo" --folder=262044416022
```

Do you think something like that is doable?

terraform/install.sh
jiang95-dev: Ok. I'll do that.

terraform/install.sh
jiang95-dev: Ok.

terraform/install.sh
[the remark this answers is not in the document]
Daniel-Sanche: You should also add the same filter here for consistency: https://github.com/GoogleCloudPlatform/stackdriver-sandbox/blob/dab7ab519a04c874058f3b2dc0d552b345590d6e/terraform/destroy.sh#L29

terraform/install.sh
jiang95-dev: I don't think we can add idap. The id will exceed its maximum length.

terraform/install.sh
jiang95-dev: The problem is that the while loop doesn't help waiting the project to be linked.

terraform/install.sh
```diff
@@ -140,7 +235,7 @@ log "Checking Prerequisites..."
 getBillingAccount;
 
 log "Install current version of Terraform"
```
Daniel-Sanche, line 237: This log should be removed

terraform/00_state.tf
```diff
@@ -12,26 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Here we configure the state backend. For this demo we're using the "local"
-# backend which stores everything in a local directory. The config you see below
-# is functionally equivalent to the defaults.
+
+# We are now using gcs as our backend, so the state file will be stored
```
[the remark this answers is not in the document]
jiang95-dev: Should I put my idap at the end?

terraform/install.sh
```diff
@@ -37,25 +37,111 @@ getBillingAccount() {
...
+createProject() {
+    # generate random id
+    project_id="stackdriver-sandbox-$(od -N 4 -t uL -An /dev/urandom | tr -d " ")"
+    bucket_name="$project_id-bucket" # bucket name should be globally unique
+    # create project
+    acct=$(gcloud info --format="value(config.account)")
+    if [[ $acct == *"google.com"* ]];
+    then
+      log ""
+      log "Are you sure you have requested access to /experimental-gke folder?"
```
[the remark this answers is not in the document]
jiang95-dev: Ok. Will do that.

terraform/install.sh
```diff
@@ -140,7 +223,7 @@ log "Checking Prerequisites..."
 getBillingAccount;
 
 log "Install current version of Terraform"
-installTerraform
+#installTerraform
```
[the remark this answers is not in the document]
jiang95-dev: This should be kept. I commented it out just for convenience when testing.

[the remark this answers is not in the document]
Daniel-Sanche: ok cool. Try to avoid pushing those changes to your branch if you can

PR #222, opened 2020-07-31
Added code to push image to GCR

eaberman:


Daniel-Sanche:
We talked about this in chat, for posterity:

I'm not sure this is the right place for this. The Cloud Shell image isn't being deployed with Kubernetes, so it doesn't have to be listed with the other artifacts.

We could consider making a separate profile for building/pushing the cloud shell image on its own. But I'm not sure if there's much of an argument for doing that rather than just doing a regular `docker build` and `docker push`

PR #239, opened 2020-08-03
Allow install.sh to be driven by the command line

Daniel-Sanche:
related issue: https://github.com/GoogleCloudPlatform/stackdriver-sandbox/issues/200

This PR adds three optional arguments arguments to install.sh (`--project-id`, `--billing-id`, `--skip-workspace-prompt`), allowing scripts to bypass the human prompts and deploy directly to an existing project. 

I also did some light refactoring:
- I moved bucket creation into its own function, so all bucket code is in one place
- bucket existence checking is more robust, allowing long-lived projects to have other buckets in the project
- I renamed user input functions to make their purpose more clear; renamed cluster authentication function to be more clear
- `gcloud billing accounts list` will sort based on name instead of random each time
- I lowered the sleep time in checkAuthorization so it's more responsive

Test link:
[Open in Cloudshell](https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/Daniel-Sanche/stackdriver-sandbox.git&cloudshell_git_branch=install-automation&cloudshell_working_dir=terraform&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image:v0.2.0)


terraform/install.sh
```diff
@@ -232,22 +245,53 @@ displaySuccessMessage() {
...
 
-log "Checking Prerequisites..."
-getBillingAccount;
+parseArguments() {
+while (( "$#" )); do
+  case "$1" in
+    -p|--project|--project-id)
+      if [ -n "$2" ] && [ ${2:0:1} != "-" ]; then
+        project_id=$2
+        shift 2
```
Daniel-Sanche, line 254: gcloud config set project "$project_id"

terraform/install.sh
```diff
@@ -204,7 +214,9 @@ installMonitoring() {
 
   gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
   log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
```
simonz130, line 216: This line should also go into if in line 217.

terraform/install.sh
```diff
@@ -278,32 +290,72 @@ checkAuthentication() {
         log "Authentication failed"
         log "Please allow gcloud and Cloud Shell to access your GCP account"
     fi
-    while [[ -z $AUTH_ACCT  && "${TRIES}" -lt 10  ]]; do
+    while [[ -z $AUTH_ACCT  && "${TRIES}" -lt 30  ]]; do
         AUTH_ACCT=$(gcloud auth list --format="value(account)")
-        sleep 3;
+        sleep 1;
```
simonz130, line 295: This might be not enough time for something to change, so second retry can still fail with high probability if it's a transient error

terraform/install.sh
```diff
@@ -278,32 +290,72 @@ checkAuthentication() {
...
     fi
 }
 
-log "Checking Prerequisites..."
+parseArguments() {
+  while (( "$#" )); do
+    case "$1" in
+    -p|--project|--project-id)
+      if [ -n "$2" ] && [ ${2:0:1} != "-" ]; then
+        project_id=$2
```
simonz130, line 308: we should also check that projectid is not empty

terraform/install.sh
Daniel-Sanche: In this case, we're waiting for the user to accept the dialog box. Since we're just doing a poll and wait here, there's no real downside to checking often, and a quicker poll time makes the script respond faster after the user finishes their task. I didn't want the user staring at the `Please allow gcloud and Cloud Shell to access your GCP account` error message for 3 seconds wondering what they did wrong 

But I'm fine with reverting too, because a 3 second wait isn't much of an issue

terraform/install.sh
Daniel-Sanche: This is already captured in `[ -n "$2" ]`. If the value is empty, it will print "Error: Argument for --project-id is missing"

terraform/install.sh
Daniel-Sanche: Good catch!

PR #240, opened 2020-08-04
add monitoring as part of installation

### LINKED ISSUE #231: Update provisioning to include terraform monitoring provisioning
As part of the provisioning, we should:

1) Ask the user to visit workspaces page (to manually create a workspace)
2) After the user confirms the workspace was created, we should provision monitoring objects: **terraform apply** on the /monitoring folder.

nicklin37:
#WHAT: Runs the terraform for monitoring examples as part of the provisioning install.sh script

#WHY: We want customers to be able to have monitoring examples automatically created during Sandbox creation

#HOW: Added it to the install script since this is the easiest way to retrieve the external IP and project ID for creating the monitoring examples. We will want to move this to be terraform in the future.

#TESTING: None :( did a manual test to verify it works correctly

Closes #231

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -174,6 +174,32 @@ applyTerraform() {
   terraform apply -auto-approve -var="billing_account=${billing_acct}" -var="project_id=${project_id}" -var="bucket_name=${bucket_name}"
 }
 
+installMonitoring() {
+  log "Retrieving the external IP address of the application..."
+  TRIES=0
+  external_ip="";
+  while [[ -z $external_ip && "${TRIES}" -lt 20 ]]; do
+     external_ip=$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'); 
+     [ -z "$external_ip" ] && sleep 5; 
+     TRIES=$((TRIES + 1))
+  done;
+
+  if [[ -z $external_ip ]]; then
+    log "Could not retrieve external IP... skipping monitoring configuration."
+    return 1
+  fi
+
+  gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
+  log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
+  read -p "When you are done, please press enter to continue"
+
+  log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."
+  pushd monitoring/
+  terraform init -lock=false
+  terraform apply --auto-approve -var="project_id=${project_id}" -var="external_ip=${external_ip}"
+  popd
+}
+
 getExternalIp() {
   external_ip=""; 
   while [ -z $external_ip ]; do
@@ -249,7 +275,8 @@ installTerraform
 # Provision Stackdriver Sandbox cluster
 getProject;
 applyTerraform;
+# || true to prevent errors during monitoring setup from stopping the installation script
+installMonitoring || true;
 getExternalIp;
 loadGen;
 displaySuccessMessage;
-


terraform/install.sh
```diff
@@ -174,6 +174,26 @@ applyTerraform() {
   terraform apply -auto-approve -var="billing_account=${billing_acct}" -var="project_id=${project_id}" -var="bucket_name=${bucket_name}"
 }
 
+installMonitoring() {
+  external_ip="";
+  while [ -z $external_ip ]; do
+     external_ip=$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'); 
+     [ -z "$external_ip" ] && sleep 10;
```
simonz130, line 181: increase the timeout - current interval is not enough to make a difference.

terraform/install.sh
```diff
@@ -174,6 +174,26 @@ applyTerraform() {
...
+
+  gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
+  log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
+  read -p "When you are done, please press enter to continue"
+
+  log "Initializing terraform state for monitoring examples"
+  cd ./monitoring
+  terraform init -lock=false
+
+  log "Apply Terraform automation for monitoring examples"
```
simonz130, line 192: Update to: "Creating dashboards and uptime checks for monitoring..."

From user's perspective it's not important that this is done with Terraform or other tool.

terraform/install.sh
nicklin37: Modified the timeout to 30 seconds. Let me know if you think it needs more.

terraform/install.sh
nicklin37: Changed the earlier comment to reflect this :)

terraform/install.sh
```diff
@@ -174,6 +174,24 @@ applyTerraform() {
...
+  external_ip="";
+  while [ -z $external_ip ]; do
+     external_ip=$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'); 
+     [ -z "$external_ip" ] && sleep 30; 
+  done;
+
+  gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
+  log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
+  read -p "When you are done, please press enter to continue"
+
```
Daniel-Sanche, line 187: Can we add some kind of check here to make sure it was enabled>

terraform/install.sh
```diff
@@ -174,6 +174,24 @@ applyTerraform() {
   terraform apply -auto-approve -var="billing_account=${billing_acct}" -var="project_id=${project_id}" -var="bucket_name=${bucket_name}"
 }
 
+installMonitoring() {
+  external_ip="";
+  while [ -z $external_ip ]; do
```
Daniel-Sanche, line 179: I think you should use a `TRIES` variable and exit with a useful error message if it is never set up. I don't want the script to get stuck forever

terraform/install.sh
```diff
@@ -174,6 +174,24 @@ applyTerraform() {
...
+     external_ip=$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'); 
+     [ -z "$external_ip" ] && sleep 30;
```
Daniel-Sanche, line 181: 30 seems like a long wait time

terraform/install.sh
```diff
@@ -174,6 +174,24 @@ applyTerraform() {
   terraform apply -auto-approve -var="billing_account=${billing_acct}" -var="project_id=${project_id}" -var="bucket_name=${bucket_name}"
 }
 
+installMonitoring() {
+  external_ip="";
```
Daniel-Sanche, line 178: you should display a message to the user so they know what they're waiting on

terraform/install.sh
```diff
@@ -174,6 +174,24 @@ applyTerraform() {
...
+  done;
+
+  gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
+  log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
+  read -p "When you are done, please press enter to continue"
+
+  log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."
+  cd ./monitoring
+  terraform init -lock=false
+  terraform apply --auto-approve -var="project_id=${project_id}" -var="external_ip=${external_ip}"
```
Daniel-Sanche, line 191: should this store the tfstate in the GCS bucket like the other terraform command?

terraform/install.sh
```diff
@@ -174,6 +174,24 @@ applyTerraform() {
...
+  cd ..
```
Daniel-Sanche, line 192: I don't like seeing relative cd commands like this, it can cause hard to catch bugs. `pushd` and `popd` could work well here

terraform/install.sh
simonz130: Great point! It should, however more importantly, it is supposed to be part of larger provisioning terraform script as opposed to be a separate run. We decided to incorporate monitoring into larger provisioning script after doing the release.

terraform/install.sh
simonz130: We should add a tracking task for this for next iterations. Currently the API is in alpha. We won't be able to check without having the project whitelisted, which is impossible in our scenario

terraform/install.sh
simonz130: Change it to 5 seconds

nicklin37: Changed :)

terraform/install.sh
nicklin37: I will open a task for this - the workaround is to use the Dashboards API since Workspaces API is in EAP. We'll add this in after the release.

terraform/install.sh
nicklin37: good idea - I've added a TRIES variable to try 20 times, retrieving the IP every 5 seconds so it'll spend at most 100 seconds on this block

terraform/install.sh
nicklin37: decreased to 5

terraform/install.sh
nicklin37: Added logging message :)

terraform/install.sh
nicklin37: Will hold off on adding it to tfstate until after the release since we want to merge things to the main terraform anyways

terraform/install.sh
nicklin37: Changed to pushd/popd.

terraform/install.sh
```diff
@@ -174,6 +174,32 @@ applyTerraform() {
...
+  log "Retrieving the external IP address of the application..."
+  TRIES=0
+  external_ip="";
+  while [[ -z $external_ip && "${TRIES}" -lt 20 ]]; do
+     external_ip=$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'); 
+     [ -z "$external_ip" ] && sleep 5; 
+     TRIES=$((TRIES + 1))
+  done;
+
+  if [[ TRIES -eq 20 ]]; then
```
Daniel-Sanche, line 187: To be safe I'd change this to `z $external_ip`, just in case it found it in the last iteration (or code changes).

It's good to check against the variable you care about, not derivative variable

terraform/install.sh
Daniel-Sanche: Sounds good. Is there an open issue around this to make sure it isn't lost?

terraform/install.sh
```diff
@@ -249,7 +275,7 @@ installTerraform
 # Provision Stackdriver Sandbox cluster
 getProject;
 applyTerraform;
+installMonitoring || true;
```
Daniel-Sanche, line 278: What's the reason for the `|| true`? Won't the script stop early within the function anyway if it encounters an error?

nicklin37: If there's an error with the Terraform apply like an API returns a 403 then the entire script will stop - this lets the rest of the script continue even if the monitoring isn't created.

terraform/install.sh
nicklin37: Gotcha. Changed :)

terraform/install.sh
simonz130: Please add a comment on this above the line, but I agree with the logic for the upcoming release.

Daniel-Sanche: Have you tested that it works? My assumption would have been that the script would exit right at line 199 if it encountered an error. I'd think you'd have to add `|| true` there

nicklin37: Yep it works! I tested it with a project w/o a workspace with and without the || true. This screenshot is with it[](url)
Screenshot 2020-08-04 at 1 41 22 PM - Display 2

nicklin37:
test passing now so merging :)

PR #245, opened 2020-08-04
Add owner permission check and credential check

jiang95-dev:
1. Add auth login when terraform init fails.
2. Add check on owner permission of a project.
3. Remove terraform warnings.
4. Adjust log wordings.

### DIFF
--- a/terraform/00_state.tf
+++ b/terraform/00_state.tf
@@ -26,6 +26,6 @@ terraform {
 data "terraform_remote_state" "state" {
     backend = "gcs"
     config = {
-        bucket = "${var.bucket_name}"
+        bucket = var.bucket_name
     }
 }
--- a/terraform/03_gke_cluster.tf
+++ b/terraform/03_gke_cluster.tf
@@ -23,7 +23,7 @@ resource "random_shuffle" "zone" {
   # found that it only ever picked `us-central-1c` unless we seeded it. Here
   # we're using the ID of the project as a seed because it is unique to the
   # project but will not change, thereby guaranteeing stability of the results.
-  seed = "${data.google_project.project.id}"
+  seed = data.google_project.project.id
 }
 
 # First we create the cluster. If you're wondering where all the sizing details
@@ -40,7 +40,7 @@ resource "random_shuffle" "zone" {
 # replicates what the Hipster Shop README creates. If you want to see what else
 # is possible, check out the docs: https://www.terraform.io/docs/providers/google/r/container_cluster.html
 resource "google_container_cluster" "gke" {
-  project = "${data.google_project.project.id}"
+  project = data.google_project.project.id
 
   # Here's how you specify the name
   name = "stackdriver-sandbox"
@@ -48,7 +48,7 @@ resource "google_container_cluster" "gke" {
   # Set the zone by grabbing the result of the random_shuffle above. It
   # returns a list so we have to pull the first element off. If you're looking
   # at this and thinking "huh terraform syntax looks a clunky" you are NOT WRONG
-  zone = "${element(random_shuffle.zone.result, 0)}"
+  zone = element(random_shuffle.zone.result, 0)
 
   # Using an embedded resource to define the node pool. Another
   # option would be to create the node pool as a separate resource and link it
@@ -106,7 +106,7 @@ resource "google_container_cluster" "gke" {
   # be enabled) before the cluster can be created. This will not address the
   # eventual consistency problems we have with the API but it will make sure
   # that we're at least trying to do things in the right order.
-  depends_on = ["google_project_service.gke"]
+  depends_on = [google_project_service.gke]
 }
 
 
@@ -122,7 +122,7 @@ resource "null_resource" "current_project" {
 #      command = "sleep 60 >./stdout.log 2>./stderr.log & echo \"sleeping in PID\" $!"
 #  }
 #
-#  depends_on = ["google_container_cluster.gke"]
+#  depends_on = [google_container_cluster.gke]
 #}
 
 # Setting kubectl context to currently deployed GKE cluster
@@ -132,8 +132,8 @@ resource "null_resource" "set_gke_context" {
   }
 
   depends_on = [
-    "google_container_cluster.gke", 
-    "null_resource.current_project"
+    google_container_cluster.gke, 
+    null_resource.current_project
   ]
 }
 
@@ -143,7 +143,7 @@ resource "null_resource" "install_istio" {
     command = "./istio/install_istio.sh"
   }
 
-  depends_on = ["null_resource.set_gke_context"]
+  depends_on = [null_resource.set_gke_context]
 }
 
 # Deploy microservices into GKE cluster 
@@ -152,7 +152,7 @@ resource "null_resource" "deploy_services" {
     command = "kubectl apply -f ../kubernetes-manifests"
   }
 
-  depends_on = ["null_resource.install_istio"]
+  depends_on = [null_resource.install_istio]
 }
 
 # We wait for all of our microservices to become available on kubernetes
@@ -172,7 +172,7 @@ resource "null_resource" "delay" {
   }
 
   triggers = {
-    "before" = "${null_resource.deploy_services.id}"
+    "before" = null_resource.deploy_services.id
   }
 }
 
--- a/terraform/destroy.sh
+++ b/terraform/destroy.sh
@@ -26,7 +26,7 @@ SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
 
 # find the stackdriver sandbox project id
-found=$(gcloud projects list --filter="name:stackdriver-sandbox-*" --format="value(projectId)")
+found=$(gcloud projects list --filter="id:stackdriver-sandbox-* AND name='Stackdriver Sandbox Demo'" --format="value(projectId)")
 if [[ -z "${found}" ]]; then
     log "error: no Stackdriver Sandbox projects found"
     exit 1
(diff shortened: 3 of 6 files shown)


terraform/03_gke_cluster.tf
```diff
@@ -67,8 +67,6 @@ resource "google_container_cluster" "gke" {
   # interesting things.
   node_pool {
     node_config {
-      machine_type = "n1-standard-2"
```
simonz130, line 70: Don't we need to specify the machine_type anymore?

simonz130: @Daniel-Sanche - was there any change related to this?

terraform/install.sh
```diff
@@ -54,7 +54,7 @@ getBillingAccount() {
   done
 
   if [[ $(echo "${found_accounts}" | wc -l) -gt 1 ]]; then
-      log "Which billing account would you like to use?:"
+      log "Which billing account would you like to use? Enter the number next to the billing account:"
```
simonz130, line 57: Enter the number representing a billing account you want to use

terraform/install.sh
```diff
@@ -75,22 +75,35 @@ getBillingAccount() {
 }
 
 getProject() {
-  log "Checking for project list"
+  log "Checking for project list..."
+  acct=$(gcloud info --format="value(config.account)")
+  # get projects with user account
+  owner_projects=$(gcloud projects list --filter="id:stackdriver-sandbox-* AND name='Stackdriver Sandbox Demo'" --format="value(projectId)")
```
simonz130, line 81: Where is the filter for owned projects? I see you filter the projects that have Stackdriver Sandbox Demo in their name

simonz130: I see that you make the determination of owner later on. The name owner_projects here is then misleading.

terraform/install.sh
```diff
@@ -75,22 +75,35 @@ getBillingAccount() {
...
+      fi
     fi
   done
-
+  # make user choose projects
   if [ -z "$found_projects" ] || [[ ${#found_projects[@]} -eq 0 ]]; then
     createProject;
   else
-      log "Which project would you like to use?:"
+      log "Which project would you like to use? Enter the number next to the project:"
```
simonz130, line 106: Enter the number next to the project you would like to use:

terraform/install.sh
```diff
@@ -236,7 +255,7 @@ log "Checking Prerequisites..."
 getBillingAccount;
 
 log "Install current version of Terraform"
-installTerraform
+#installTerraform
```
simonz130, line 258: You can remove this given that now it's part of the Cloud Shell container image

terraform/03_gke_cluster.tf
jiang95-dev: It didn't modify this line.

terraform/install.sh
jiang95-dev: Ok.

terraform/03_gke_cluster.tf
Daniel-Sanche: We do need this line. Please make sure you keep changes from upstream

terraform/03_gke_cluster.tf
```diff
@@ -78,8 +76,8 @@ resource "google_container_cluster" "gke" {
         cluster = "stackdriver-sandbox-main"   
       }
     }
-
-    initial_node_count = 4
+    
+    initial_node_count = 7
```
Daniel-Sanche, line 80: Revert this

terraform/destroy.sh
```diff
@@ -26,7 +26,7 @@ SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
 
 # find the stackdriver sandbox project id
-found=$(gcloud projects list --filter="name:stackdriver-sandbox-*" --format="value(projectId)")
+found=$(gcloud projects list --filter="id:stackdriver-sandbox-* AND name='Stackdriver Sandbox Demo'" --format="value(projectId)")
```
Daniel-Sanche, line 29: revert this

jiang95-dev: Oh, why is that? Shouldn't we also filter the name for demo?

terraform/00_state.tf
```diff
@@ -26,6 +26,6 @@ terraform {
 data "terraform_remote_state" "state" {
     backend = "gcs"
     config = {
-        bucket = "${var.bucket_name}"
+        bucket = var.bucket_name
```
Daniel-Sanche, line 29: Are all of these quote changes on purpose? What's the difference?

jiang95-dev: To remove the warnings. Quotes are deprecated.

terraform/install.sh
```diff
@@ -75,22 +75,35 @@ getBillingAccount() {
...
+  # get projects with user account
+  owner_projects=$(gcloud projects list --filter="id:stackdriver-sandbox-* AND name='Stackdriver Sandbox Demo'" --format="value(projectId)")
+  # get projects with billing account
   billed_projects=$(gcloud beta billing projects list --billing-account="$billing_id" --filter="project_id:stackdriver-sandbox-*" --format="value(projectId)")
-  # only keep projects with name "Stackdriver Sandbox Demo"
+  declare -A proj_map
+  for proj in ${billed_projects[@]}
+  do
+    proj_map["$proj"]="yes"
+  done
```
Daniel-Sanche, line 88: I don't think I understand the purpose of proj_map

terraform/install.sh
```diff
@@ -75,22 +75,35 @@ getBillingAccount() {
 }
 
 getProject() {
-  log "Checking for project list"
+  log "Checking for project list..."
+  acct=$(gcloud info --format="value(config.account)")
+  # get projects with user account
+  owner_projects=$(gcloud projects list --filter="id:stackdriver-sandbox-* AND name='Stackdriver Sandbox Demo'" --format="value(projectId)")
+  # get projects with billing account
   billed_projects=$(gcloud beta billing projects list --billing-account="$billing_id" --filter="project_id:stackdriver-sandbox-*" --format="value(projectId)")
```
Daniel-Sanche, line 83: Why do you need "owner_projects" and "billed_projects"?

terraform/install.sh
jiang95-dev: It's easier to check if a variable exists in a map than a list.

terraform/install.sh
jiang95-dev: Because "gcloud projects list" won't tell you anything about the billing info. And not all billing project lets you get IAM if you don't have at least browser permission. So I have to combine them.

terraform/destroy.sh
jiang95-dev: @Daniel-Sanche

terraform/install.sh
Daniel-Sanche: Since we only care about projects with owner permission, do we need the ones without browser permission?

terraform/destroy.sh
Daniel-Sanche: sorry, I thought this was another accidental overwrite. This can be kept

terraform/00_state.tf
Daniel-Sanche: ok sounds good

terraform/install.sh
Daniel-Sanche: It would be more readable if we only had one list of projects, and we could just iterate through the list. Something like this: https://stackoverflow.com/questions/2696055/intersection-of-two-lists-in-bash

But I'm skeptical that we really need both `billed_projects` and `owned_projects`

terraform/install.sh
jiang95-dev: I mean if you use "gcloud projects get-iam-policy" on projects that you don't have at least browser permission, it will throw an error. And there are many projects under the billing account that you may not have permission for the current user account.

terraform/install.sh
Daniel-Sanche: @simonz130 aren't we leaving the cloud shell container out for this release though?

terraform/install.sh
Daniel-Sanche: We can ignore projects that throw an error though, right? We know those won't make it onto the list we display to the user

terraform/install.sh
```diff
@@ -165,7 +165,8 @@ applyTerraform() {
   rm -f .terraform/terraform.tfstate
 
   log "Initialize terraform backend with bucket ${bucket_name}"  
-  if [[ $(terraform init -backend-config "bucket=${bucket_name}" -lock=false) ]]; then
+  
+  if [[ $(terraform init -backend-config "bucket=${bucket_name}" -lock=false 2> /dev/null | cat) ]]; then
```
Daniel-Sanche, line 169: Won't this always return true now?

terraform/install.sh
simonz130: No, we are adding cloud shell custom image in upcoming release - the Cloud Shell team did fix the blocker on their side.

terraform/install.sh
Daniel-Sanche: `If [[ ` statements expect some kind of comparitor, this line as is may give us unexpected results

jiang95-dev:
Reopen it after rebasing.

terraform/install.sh
```diff
@@ -157,49 +161,24 @@ createProject() {
...
-  log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
-  read -p "When you are done, please press enter to continue"
-
-  log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."
-  pushd monitoring/
-  terraform init -lock=false
-  terraform apply --auto-approve -var="project_id=${project_id}" -var="external_ip=${external_ip}"
-  popd
-}
-
```
simonz130, line 202: why is this removed?

PR #249, opened 2020-08-04
OpenTelemetry Tracing for ProductCatalogService

### LINKED ISSUE #248: OpenTelemetry Tracing for ProductCatalogService
Subtask of #132 
Use OpenTelemetry for tracing in the productcatalog service.

gzhao408:
Resolves #248 . Subtask of #132 .

To test, deploy the sandbox and you should see this in Cloud Trace (specifically, look for Span.{server, client}-hipstershop.ProductCatalogService/{ListProducts, GetProduct}):

image

image

### DIFF
--- a/src/productcatalogservice/server.go
+++ b/src/productcatalogservice/server.go
@@ -173,21 +173,35 @@ func initOpenCensusStats() {
 func initTraceProvider() {
 	// Initialize exporter OTel Trace -> Google Cloud Trace
 	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
-	exporter, err := texporter.NewExporter(texporter.WithProjectID(projectID))
-	if err != nil {
-		log.Fatalf("failed to initialize exporter: %v", err)
+	if len(projectID) == 0 {
+		log.Warn("GOOGLE_CLOUD_PROJECT not set")
 	}
-
-	// Create trace provider with the exporter.
-	tp, err := sdktrace.NewProvider(sdktrace.WithConfig(
-		sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}),
-		sdktrace.WithSyncer(exporter),
-		// TODO: replace with predefined constant for GKE or autodetection when available
-		sdktrace.WithResource(resource.New(standard.ServiceNameKey.String("GKE"))))
-	if err != nil {
-		log.Fatal("failed to initialize trace provider: %v", err)
+	for i := 1; i <= 3; i++ {
+		exporter, err := texporter.NewExporter(texporter.WithProjectID(projectID))
+		if err != nil {
+			log.Infof("failed to initialize exporter: %v", err)
+		} else {
+			// Create trace provider with the exporter.
+			// The AlwaysSample sampling policy is used here for demonstration
+			// purposes and should not be used in production environments.
+			tp, err := sdktrace.NewProvider(sdktrace.WithConfig(
+				sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}),
+				sdktrace.WithSyncer(exporter),
+				// TODO: replace with predefined constant for GKE or autodetection when available
+				sdktrace.WithResource(resource.New(standard.ServiceNameKey.String("GKE"))))
+			if err == nil {
+				log.Info("initialized trace provider")
+				global.SetTraceProvider(tp)
+				return
+			} else {
+				d := time.Second * 10 * time.Duration(i)
+				log.Infof("sleeping %v to retry initializing trace provider", d)
+				time.Sleep(d)
+			}
+		}
+		
 	}
-	global.SetTraceProvider(tp)
+	log.Warn("failed to initialize trace provider")
 }
 
 func initProfiling(service, version string) {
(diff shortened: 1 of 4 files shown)


src/frontend/main.go
```diff
@@ -120,12 +120,12 @@ func main() {
 	go initTelemetry(log)
 
 	// TODO: register views when OpenTelemetry Views API is available
-	meter := controller.Provider().Meter("hipstershop/frontend")
+	meter := controller.Provider().Meter("hipstershop")
```
simonz130, line 123: Is this change related to product catalog service?

src/productcatalogservice/server.go
```diff
@@ -78,7 +85,8 @@ func init() {
 }
 
 func main() {
-	initStackDriverTracing()
+	initOpenCensus()
```
simonz130, line 88: Maybe rename OpenCensus method to be more concrete: initOpenCensusMetrics
This way it's more clear what is part of OpenTelemetry vs OpenCensus instrumentation as we work towards getting rid of OpenCensus completely.

src/productcatalogservice/server.go
```diff
@@ -119,14 +127,20 @@ func run(port int) string {
...
 	svc := &productCatalog{}
 	pb.RegisterProductCatalogServiceServer(srv, svc)
 	healthpb.RegisterHealthServer(srv, svc)
 	go srv.Serve(l)
 	return l.Addr().String()
 }
 
+// Initialize OpenCensus Stats (Metrics)
+// TODO: remove this after full conversion to OpenTelemetry
 func initStats(exporter *stackdriver.Exporter) {
```
simonz130, line 144: Make it clear it's OpenCensus in the method name

src/frontend/main.go
gzhao408: removed this change

src/productcatalogservice/server.go
gzhao408: I just moved the helper function in there and called it initOpenCensusStats, and commented that this would be replaced with OpenTelemetry Metrics.

gzhao408:
@simonz130 did a [refactor](https://github.com/GoogleCloudPlatform/stackdriver-sandbox/compare/c5bf1ff936c873780074a7705681f7971e3b98e9..95db4d2c0be67909d36ee8ac71a894184f6d5501)

Daniel-Sanche:
For some reason I'm not seeing those traces, only BatchWriteSpans and health checks. I'm not too familiar with Cloud Trace though, so it could just be a me problem

gzhao408:
> For some reason I'm not seeing those traces

:O no span.server-frontend whatsoever? Oh, maybe you need to open hipstershop and click around a few times, or have the loadgenerator hit it with some requests, otherwise there might not be anything to trace?

src/productcatalogservice/server.go
```diff
@@ -164,6 +169,27 @@ func initStackDriverTracing() {
 	log.Warn("could not initialize stackdriver exporter after retrying, giving up")
 }
 
+// Initialize OTel trace provider that exports to Cloud Trace
+func initTraceProvider() {
+	// Initialize exporter OTel Trace -> Google Cloud Trace
+	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
+	exporter, err := texporter.NewExporter(texporter.WithProjectID(projectID))
```
simonz130, line 176: Make sure the projectId is not null or empty as well as check the presense of the GOOGLE_APPLICATION_CREDENTIALS env variable to make sure that auth will work.

src/productcatalogservice/server.go
```diff
@@ -164,6 +169,27 @@ func initStackDriverTracing() {
...
+	}
+
+	// Create trace provider with the exporter.
+	tp, err := sdktrace.NewProvider(sdktrace.WithConfig(
+		sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}),
+		sdktrace.WithSyncer(exporter),
+		// TODO: replace with predefined constant for GKE or autodetection when available
+		sdktrace.WithResource(resource.New(standard.ServiceNameKey.String("GKE"))))
+	if err != nil {
+		log.Fatal("failed to initialize trace provider: %v", err)
```
simonz130, line 188: I don't think we should fail the app if tracing doesn't work, but we should log it as critical error.

Daniel-Sanche:
> :O no span.server-frontend whatsoever? Oh, maybe you need to open hipstershop and click around a few times, or have the loadgenerator hit it with some requests, otherwise there might not be anything to trace?

Nevermind, I tried again today and it's working!

src/productcatalogservice/server.go
gzhao408: checking that projectID isn't empty, I don't think it's necessary to check credentials, would be the responsibility of the [exporter](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/4bf57be619cd46e7dfce46107fc7baa368ebc9ef/exporter/trace/README.md#authentication)

src/productcatalogservice/server.go
gzhao408: yep, got it :)

gzhao408:
@simonz130 Please take another look, I'm requiring PROJECT_ID since it's a required parameter, but the auth/credentials are the [responsibility of the exporter](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/4bf57be619cd46e7dfce46107fc7baa368ebc9ef/exporter/trace/README.md#authentication). I'm having it retry up to 3 times and just warn if it fails. Thanks!

PR #251, opened 2020-08-04
Add owner permission check and credential check

Daniel-Sanche:
Since Levi is sick, I made some additions to his PR so we can get it merged for the release tomorrow

more details in Levi's PR: https://github.com/GoogleCloudPlatform/stackdriver-sandbox/pull/245

### DIFF
--- a/terraform/00_state.tf
+++ b/terraform/00_state.tf
@@ -16,22 +16,28 @@
 # backend which stores everything in a local directory. The config you see below
 # is functionally equivalent to the defaults.
 
-terraform {
-  backend "local" {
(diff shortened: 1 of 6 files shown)


Daniel-Sanche:
I've noticed that there's now quite a bit of lag when fetching the list of projects (~20 seconds). It might be worth removing the creation time field, that would cut the load time in half. But this will only really impact internal users, so not sure if it really matters

Daniel-Sanche:
Levi's back at work. I merged my changes into his fork, so we can keep working from there

PR #259, opened 2020-08-05
add project to monitoring terraform provider

### LINKED ISSUE #258: Creating monitoring objects fails with missing info
When running ./install.sh from /terraform folder in master branch, I get to the monitoring provisioning part, and then see this:

```bash
  on 02_dashboards.tf line 37, in resource "google_monitoring_dashboard" "recommendationservice_dashboard":
  37: resource "google_monitoring_dashboard" "recommendationservice_dashboard" {
Error: project: required field is not set
  on 02_dashboards.tf line 81, in resource "google_monitoring_dashboard" "service_dashboards":
  81: resource "google_monitoring_dashboard" "service_dashboards" {
Error: project: required field is not set
  on 02_dashboards.tf line 81, in resource "google_monitoring_dashboard" "service_dashboards":
  81: resource "google_monitoring_dashboard" "service_dashboards" {
Error: project: required field is not set
  on 02_dashboards.tf line 81, in resource "google_monitoring_dashboard" "service_dashboards":
  81: resource "google_monitoring_dashboard" "service_dashboards" {
Error: project: required field is not set
  on 03_service_slo.tf line 22, in resource "google_monitoring_custom_service" "frontend":
  22: resource "google_monitoring_custom_service" "frontend" {
```

nicklin37:
We're running into an issue where Terraform seems to think there is no configured project for the monitored resource. This fix sets the project for the provider so that won't be an issue.

Not sure why the issue exists for some projects and not for others, this should fix it universally.

Also updating the types for the variables to suppress the warnings.

Closes #258

### DIFF
--- a/terraform/monitoring/00_provider.tf
+++ b/terraform/monitoring/00_provider.tf
@@ -27,7 +27,7 @@ provider "google" {
   version = ">=3.23.0"
 
   # credentials = "/path/to/creds.json"
-  # project = "project-id"
+  project = "var.project_id"
   # region = "default-region"
   # zone = "default-zone"
 } 
--- a/terraform/monitoring/02_dashboards.tf
+++ b/terraform/monitoring/02_dashboards.tf
@@ -69,7 +69,7 @@ variable "services" {
 # the template file defined in the 'dashboards' folder.
 data "template_file" "dash_json" {
   template = "${file("./dashboards/generic_dashboard.tmpl")}"
-  count    = "${length(var.services)}"
+  count    = "length(var.services)"
   vars     = {
     service_name = "${var.services[count.index].service_name}"
     service_id   = "${var.services[count.index].service_id}"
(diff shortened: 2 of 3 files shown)


terraform/monitoring/00_provider.tf
```diff
@@ -27,7 +27,7 @@ provider "google" {
   version = ">=3.23.0"
 
   # credentials = "/path/to/creds.json"
-  # project = "project-id"
+  project = "${var.project_id}"
```
simonz130, line 30: I think Terraform doesn't like using quotes anymore. Can you try it without quotes?

nicklin37: Yeah it actually doesn't like the ${} syntax anymore - kept the quotes but removed the interpolations to suppress the warnings. Will merge once tests pass unless there are other fixes.

Issue #257, opened 2020-08-05
Email should be autodetected from username in monitoring provisioning


simonz130:
![image](https://user-images.githubusercontent.com/32641282/89377952-ad15f000-d6a7-11ea-8db9-5b33d5314041.png)

simonz130:
Fixed by @nicklin37

PR #265, opened 2020-08-05
fix email and deprecations

### LINKED ISSUE #256: Using deprecated terraform field in provisioning
When provisioning a new cluster, here is what I get:
```
Warning: "zone": [DEPRECATED] Use location instead
  on 03_gke_cluster.tf line 42, in resource "google_container_cluster" "gke":
  42: resource "google_container_cluster" "gke" {
```

nicklin37:
1) Configures the user email for monitoring terraform

2) Gets rid of deprecated Zone for GKE cluster

3) Add back in the interpolation for the zone.

Closes #256 and #257

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -194,6 +194,8 @@ installMonitoring() {
     return 1
   fi
 
+  acct=$(gcloud info --format="value(config.account)")
+
   gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
   log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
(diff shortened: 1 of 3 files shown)


nicklin37:
interpolation for the variable length of services**

terraform/install.sh
```diff
@@ -194,14 +194,16 @@ installMonitoring() {
     return 1
   fi
 
+  acct=$(gcloud info --format="value(config.account)")
```
jiang95-dev, line 197: This line should be removed.  Acct is set in getProject().

nicklin37: We need to call it again to retrieve the email for the Monitoring Terraform configuration

PR #271, opened 2020-08-05
set current project

jiang95-dev:
Set current environment to use the property of selected project.

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -112,6 +112,7 @@ getProject() {
           opt=($opt)
           project_id=${opt[0]}
           bucket_name="$project_id-bucket"
+          gcloud config set project "$project_id"
           break
         fi
       done
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -112,12 +112,12 @@ getProject() {
           opt=($opt)
           project_id=${opt[0]}
           bucket_name="$project_id-bucket"
-          gcloud config set project "$project_id"
           break
         fi
       done
       IFS=$IFS_bak
   fi
+  gcloud config set project "$project_id"
 }
 
 createProject() {


terraform/install.sh
```diff
@@ -112,6 +112,7 @@ getProject() {
           opt=($opt)
           project_id=${opt[0]}
           bucket_name="$project_id-bucket"
+          gcloud config set project "$project_id"
```
Daniel-Sanche, line 115: I'd suggest putting this at the end of the function, so we know it'll be run no matter what

nicklin37: @Dan Not sure how bash variable scoping works but gcloud config set project will be called within createProject already

jiang95-dev: It's ok to double set the config.
And bash variable is always global.

Daniel-Sanche: Leaving this important function inside 3 layers of if/else/select logic is setting us up for bugs later. If we want to make sure this line will apply to all projects, we should run it at the bottom of the function instead of trying to replicate it in each logic branch.

Daniel-Sanche: The same idea applies to the bucket_name. It's a bad idea to define custom names in two different branches in the code, But I have a refactor staged for that [here](https://github.com/GoogleCloudPlatform/stackdriver-sandbox/pull/239) after this release

nicklin37: Gotcha - I agree, I was just worried project_id might not have a valid value outside of the if/else branches but it seems like that might not be the case.

Daniel-Sanche: I see, yes, project_id is global and used across the script

jiang95-dev: @Daniel-Sanche "gsutil ls" needs the project config, so that was why I config project in createProject(). Removing it will cause problems.

Daniel-Sanche: I'm fine with leaving it in createProject too, I just want you to move this one to the bottom of the function so we can be sure it's always called

PR #277, opened 2020-08-05
OpenTelemetry Tracing for Checkout Service + Reuse gRPC Connections

### LINKED ISSUE #275: OpenTelemetry tracing for checkout service
Subtask of #132

### LINKED ISSUE #276: Reuse gRPC connections in checkout service
Mentioned in the commment [here](https://github.com/GoogleCloudPlatform/stackdriver-sandbox/blob/b27627325b8f5ec699a262d08eec100411afc23e/src/checkoutservice/main.go#L379).

The checkout service creates a new gRPC client connection every time it calls another service, and it has 6 services that it talks to. This is inefficient and quite redundant.

gzhao408:
Resolves #275 and subtask of #132. 

I decided to fix #276 too because otherwise it was introducing a lot of redundant code. Previously, we opened+closed a brand new connection every time we needed to call another service; checkout is a client to 6 services and a single ShipOrder has 10+ calls. I've changed it:
(1) Added the gRPC connections to the checkout service data structure
(2) Open the connections at the start, with OC and OT telemetry interceptors/handler wrappers. 
(3) Reuse the connections for subsequent calls (they're methods that belong to the checkout service and thus have access to the connections directly).
This happens to greatly simplify the code as well.

To test, deploy hipster shop, add things to your cart, and then check out. You should see the following trace with `g.co/agent:opentelemetry0.9.0`. I added events to the checkout span so you can see the relative latencies for each task in the check out process :) 
image

Note: Merging this earlier than the others would be best, because the checkout service acts as a client to 6 different services and is needed to propagate trace contexts.

### DIFF
--- a/src/checkoutservice/main.go
+++ b/src/checkoutservice/main.go
@@ -68,7 +68,7 @@ type checkoutService struct {
 }
 
 func main() {
-	go initStackDriverTracing()
+	go initOpenCensusStats()
 	go initProfiling("checkoutservice", "1.0.0")
 
 	port := listenPort
@@ -98,19 +98,9 @@ func main() {
 	log.Fatal(err)
 }
 
-func initStats(exporter *stackdriver.Exporter) {
-	view.SetReportingPeriod(60 * time.Second)
-	view.RegisterExporter(exporter)
-	if err := view.Register(ocgrpc.DefaultServerViews...); err != nil {
-		log.Warn("Error registering default server views")
-	} else {
-		log.Info("Registered default server views")
-	}
-}
-
-func initStackDriverTracing() {
-	// TODO(ahmetb) this method is duplicated in other microservices using Go
-	// since they are not sharing packages.
+// Initialize Stats collection using OpenCensus
+// TOOD: remove this after conversion to OpenTelemetry Metrics
+func initOpenCensusStats() {
 	for i := 1; i <= 3; i++ {
 		exporter, err := stackdriver.NewExporter(stackdriver.Options{})
 		if err != nil {
@@ -120,7 +110,13 @@ func initStackDriverTracing() {
 			log.Info("registered stackdriver tracing")
 
 			// Register the views to collect server stats.
-			initStats(exporter)
+			view.SetReportingPeriod(60 * time.Second)
+			view.RegisterExporter(exporter)
+			if err := view.Register(ocgrpc.DefaultServerViews...); err != nil {
+				log.Warn("Error registering default server views")
+			} else {
+				log.Info("Registered default server views")
+			}
 			return
 		}
 		d := time.Second * 10 * time.Duration(i)
@@ -164,6 +160,7 @@ func (cs *checkoutService) Check(ctx context.Context, req *healthpb.HealthCheckR
 	return &healthpb.HealthCheckResponse{Status: healthpb.HealthCheckResponse_SERVING}, nil
 }
 
+// Acting as gRPC server, handling PlaceOrder request from Frontend.
 func (cs *checkoutService) PlaceOrder(ctx context.Context, req *pb.PlaceOrderRequest) (*pb.PlaceOrderResponse, error) {
 	log.Infof("[PlaceOrder] user_id=%q user_currency=%q", req.UserId, req.UserCurrency)
 
@@ -172,6 +169,7 @@ func (cs *checkoutService) PlaceOrder(ctx context.Context, req *pb.PlaceOrderReq
 		return nil, status.Errorf(codes.Internal, "failed to generate order uuid")
 	}
 
+	// Dispatch subtasks (as gRPC client)
 	prep, err := cs.prepareOrderItemsAndShippingQuoteFromCart(ctx, req.UserId, req.UserCurrency, req.Address)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, err.Error())
@@ -246,6 +244,7 @@ func (cs *checkoutService) prepareOrderItemsAndShippingQuoteFromCart(ctx context
 	return out, nil
 }
 
+// Acting as gRPC client
 func (cs *checkoutService) quoteShipping(ctx context.Context, address *pb.Address, items []*pb.CartItem) (*pb.Money, error) {
 	conn, err := grpc.DialContext(ctx, cs.shippingSvcAddr,
 		grpc.WithInsecure(),
@@ -265,6 +264,7 @@ func (cs *checkoutService) quoteShipping(ctx context.Context, address *pb.Addres
 	return shippingQuote.GetCostUsd(), nil
 }
 
+// Acting as gPRC client
 func (cs *checkoutService) getUserCart(ctx context.Context, userID string) ([]*pb.CartItem, error) {
 	conn, err := grpc.DialContext(ctx, cs.cartSvcAddr, grpc.WithInsecure(), grpc.WithStatsHandler(&ocgrpc.ClientHandler{}))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


src/checkoutservice/main.go
```diff
@@ -84,43 +106,47 @@ func main() {
...
-
 			// Register the views to collect server stats.
-			initStats(exporter)
+			view.SetReportingPeriod(60 * time.Second)
+			view.RegisterExporter(exporter)
+			if err := view.Register(ocgrpc.DefaultServerViews...); err != nil {
+				log.Warn("Error registering default server views")
+			} else {
+				log.Info("Registered default server views")
+			}
```
eaberman, line 149: What makes it necessary to put this block of code here as opposed to above the if/else clause for NewExporter()?

src/checkoutservice/main.go
```diff
@@ -265,42 +313,28 @@ func (cs *checkoutService) quoteShipping(ctx context.Context, address *pb.Addres
...
+// Acting as gPRC client
 func (cs *checkoutService) getUserCart(ctx context.Context, userID string) ([]*pb.CartItem, error) {
-	conn, err := grpc.DialContext(ctx, cs.cartSvcAddr, grpc.WithInsecure(), grpc.WithStatsHandler(&ocgrpc.ClientHandler{}))
-	if err != nil {
-		return nil, fmt.Errorf("could not connect cart service: %+v", err)
-	}
-	defer conn.Close()
-
-	cart, err := pb.NewCartServiceClient(conn).GetCart(ctx, &pb.GetCartRequest{UserId: userID})
+	cart, err := pb.NewCartServiceClient(cs.cartSvcConn).GetCart(ctx, &pb.GetCartRequest{UserId: userID})
```
eaberman, line 318: Nice way to reduce redundancies in the code!

src/checkoutservice/main.go
```diff
@@ -265,42 +313,28 @@ func (cs *checkoutService) quoteShipping(ctx context.Context, address *pb.Addres
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to get user cart during checkout: %+v", err)
```
eaberman, line 320: Why log.fatal() errors in some places, but return nil in others? (I'm not familiar with Golang--is there a difference?)

gzhao408: not 100% sure what you're referring to, but return nil would mean no error
when to return depends on the severity of the error, some thing we want to simply log, others are unrecoverable and should halt execution immediately

src/checkoutservice/main.go
gzhao408: it wasn't reused, and i wanted it to be easy to delete

src/checkoutservice/main.go
```diff
@@ -130,6 +156,26 @@ func initStackDriverTracing() {
 	log.Warn("could not initialize stackdriver exporter after retrying, giving up")
 }
 
+// Initialize OTel trace provider that exports to Cloud Trace
+func InitTraceProvider() {
```
nicklin37, line 160: nit: camelCase function doesn't match the rest, maybe prefer initTraceProvider?

gzhao408: woops, addressed.

Daniel-Sanche:
Can you sync up to master? It looks like your current install.sh doesn't support internal accounts

gzhao408:
@Daniel-Sanche rebased on current master

src/checkoutservice/main.go
```diff
@@ -84,43 +106,47 @@ func main() {
 	mustMapEnv(&svc.emailSvcAddr, "EMAIL_SERVICE_ADDR")
 	mustMapEnv(&svc.paymentSvcAddr, "PAYMENT_SERVICE_ADDR")
 
+	mustConnGRPC(ctx, &svc.shippingSvcConn, svc.shippingSvcAddr)
+	mustConnGRPC(ctx, &svc.productCatalogSvcConn, svc.productCatalogSvcAddr)
+	mustConnGRPC(ctx, &svc.cartSvcConn, svc.cartSvcAddr)
+	mustConnGRPC(ctx, &svc.currencySvcConn, svc.currencySvcAddr)
+	mustConnGRPC(ctx, &svc.emailSvcConn, svc.emailSvcAddr)
+	mustConnGRPC(ctx, &svc.paymentSvcConn, svc.paymentSvcAddr)
```
Daniel-Sanche, line 114: Can you explain this part a bit more? What happens if a service isn't available? Or if the address changes?

gzhao408: Yeah sure, `mustMapEnv` retrieves the addresses from environment variables. Then, `mustConnGRPC` attempts to make the connection using the address, and [errors](https://github.com/GoogleCloudPlatform/stackdriver-sandbox/pull/277/files#diff-48d0b5ff7f7a67a009a11c584b29d422R404) if it fails.

The addresses are set as environment variables in the kubernetes-manifests, and if an address changes, it's retrieved at the start. I'm assuming this means that it'll have the right information.

src/checkoutservice/main.go
```diff
@@ -130,6 +156,26 @@ func initStackDriverTracing() {
 	log.Warn("could not initialize stackdriver exporter after retrying, giving up")
 }
 
+// Initialize OTel trace provider that exports to Cloud Trace
+func initTraceProvider() {
+	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
```
simonz130, line 161: ensure that the value is not empty or null

src/checkoutservice/main.go
```diff
@@ -130,6 +156,26 @@ func initStackDriverTracing() {
...
+	exporter, err := texporter.NewExporter(texporter.WithProjectID(projectID))
+	if err != nil {
+		log.Fatalf("failed to initialize exporter: %v", err)
```
simonz130, line 164: Not sure if we should exit the app when trace exporter fails to initialize. It makes our app more fragile and dependent on exporter implementation

simonz130: How the auth to Cloud Trace is handled? I assume it's likely using environment variable that is supposed to be set (GOOGLE_APPLICATION_CREDENTIALS). So worth ensuring this variable is set correctly

src/checkoutservice/main.go
```diff
@@ -130,6 +156,26 @@ func initStackDriverTracing() {
...
+	}
+
+	// Create trace provider with the exporter.
+	tp, err := sdktrace.NewProvider(sdktrace.WithConfig(
+		sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}),
```
simonz130, line 169: add a note that AlwaysSample is not a good strategy for production due to scale concerns, but we use it for demo purposes

src/checkoutservice/main.go
Daniel-Sanche: What happens if one of the other pods goes down? Will the connection re-establish with a new one? Do you know if this connection is load balanced if we scale up the other services? Or is it always going to the same Pod?

I want to make sure we don't become too tightly coupled with other services. Ideally, they should be free to scale or move or crash without impacting this one

Daniel-Sanche: never mind, I did some testing and deleting and recreating pods seems fine!

src/checkoutservice/main.go
gzhao408: good idea, done

src/checkoutservice/main.go
gzhao408: yep it pulls from environment variables for auth
if running on gcp, I believe these are set automatically. otherwise it needs to be set manually or be provided a credentials file

src/checkoutservice/main.go
gzhao408: got it, done

src/checkoutservice/main.go
gzhao408: I'm having it retry up to 3 times to be a little less brittle. Info if it fails once, Warn if it fails three times.

gzhao408:
@simonz130 Please take another look, I'm requiring PROJECT_ID since it's a required parameter, but the auth/credentials are the [responsibility of the exporter](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/4bf57be619cd46e7dfce46107fc7baa368ebc9ef/exporter/trace/README.md#authentication). I'm having it retry up to 3 times and just warn if it fails. Thanks!

PR #279, opened 2020-08-05
OpenTelemetry Tracing for Shipping Service

### LINKED ISSUE #278: OpenTelemetry Tracing for Shipping Service
Subtask of #132

gzhao408:
Resolves #278, subtask of #132 

To test, deploy hipster shop and either (1) view your cart (which gives you a shipping quote) or (2) check out (which calls the shipping service to "ship" your items). You should see the following trace in your Cloud Trace console with `g.co/agent:opentelemetry`:

image

### DIFF
--- a/src/shippingservice/main.go
+++ b/src/shippingservice/main.go
@@ -25,7 +25,6 @@ import (
 	"github.com/sirupsen/logrus"
 	"go.opencensus.io/plugin/ocgrpc"
 	"go.opencensus.io/stats/view"
-	"go.opencensus.io/trace"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/reflection"
@@ -81,7 +80,7 @@ func main() {
 	}
 	// TOOD: replace ocgrpc with automatic OpenTelemetry grpc metrics collector
 	srv := grpc.NewServer(grpc.StatsHandler(
-		&ocgrpc.ServerHandler{}),
+		&ocgrpc.ServerHandler{}), // TODO: replace with automatic OTel grpc metrics collector when available
 		grpc.UnaryInterceptor(grpctrace.UnaryServerInterceptor(global.TraceProvider().Tracer("shipping"))),
 		grpc.StreamInterceptor(grpctrace.StreamServerInterceptor(global.TraceProvider().Tracer("shipping"))),
 	)
@@ -156,10 +155,6 @@ func initOpenCensusStats() {
 		if err != nil {
 			log.Warnf("failed to initialize stackdriver exporter: %+v", err)
 		} else {
-			trace.RegisterExporter(exporter)
-			trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
-			log.Info("registered stackdriver tracing")
-
 			// Register the views to collect server stats.
 			view.SetReportingPeriod(60 * time.Second)
 			view.RegisterExporter(exporter)
--- a/src/shippingservice/main.go
+++ b/src/shippingservice/main.go
@@ -29,6 +29,16 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/reflection"
+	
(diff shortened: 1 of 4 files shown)


src/shippingservice/main.go
```diff
@@ -68,7 +78,12 @@ func main() {
 	if err != nil {
 		log.Fatalf("failed to listen: %v", err)
 	}
-	srv := grpc.NewServer(grpc.StatsHandler(&ocgrpc.ServerHandler{}))
+	// TOOD: replace ocgrpc with automatic OpenTelemetry grpc metrics collector
```
eaberman, line 81: Is this the automatic OpenTelemetry metrics collector?

src/shippingservice/main.go
```diff
@@ -68,7 +78,12 @@ func main() {
...
+	srv := grpc.NewServer(grpc.StatsHandler(
+		&ocgrpc.ServerHandler{}), // TODO: replace with automatic OTel grpc metrics collector when available
+		grpc.UnaryInterceptor(grpctrace.UnaryServerInterceptor(global.TraceProvider().Tracer("shipping"))),
+		grpc.StreamInterceptor(grpctrace.StreamServerInterceptor(global.TraceProvider().Tracer("shipping"))),
+	)
```
eaberman, line 86: What do the additional arguments do?

src/shippingservice/main.go
```diff
@@ -116,6 +133,8 @@ func (s *server) GetQuote(ctx context.Context, in *pb.GetQuoteRequest) (*pb.GetQ
 // ShipOrder mocks that the requested items will be shipped.
 // It supplies a tracking ID for notional lookup of shipment delivery status.
 func (s *server) ShipOrder(ctx context.Context, in *pb.ShipOrderRequest) (*pb.ShipOrderResponse, error) {
+	span := apitrace.SpanFromContext(ctx)
+	span.AddEvent(ctx, "Ship Order")
 	log.Info("[ShipOrder] received request")
```
eaberman, line 138: What does this "span" variable do?

src/shippingservice/main.go
```diff
@@ -128,30 +147,22 @@ func (s *server) ShipOrder(ctx context.Context, in *pb.ShipOrderRequest) (*pb.Sh
...
-		log.Info("Registered default server views")
-	}
-}
-
-func initStackDriverTracing() {
-	// TODO(ahmetb) this method is duplicated in other microservices using Go
-	// since they are not sharing packages.
+// Initialize Stats using OpenCensus
+// TODO: remove this after conversion to OpenTelemetry Metrics
+func initOpenCensusStats() {
```
eaberman, line 152: This function seems very similar to a function in the checkout service. Is there any way to simplify this to one function across the services (if it is the case that they're similar enough)?

src/shippingservice/main.go
gzhao408: nope, it doesn't exist yet :P that's why TODO

src/shippingservice/main.go
gzhao408: they intercept all grpc messages to collect information, in this case the trace context so we can link calls between services

src/shippingservice/main.go
gzhao408: span = one unit of execution, in this case it represents the ShipOrder call
seeing multiple spans for one trace, e.g. in the "waterfall view" like in GCP Trace console helps us see the "story" of a call across all microservices and see relative latencies/bottlenecks.

src/shippingservice/main.go
gzhao408: they're identical, yes :) the problem is these services are isolated and don't share packages, which is intentional. it's possible to reuse, but it's a small amount of code and we're planning on deleting it anyway

gzhao408:
> were you ever able to get the testing to work using the API?

There's an auth issue with the Cloud Trace API, in the process of asking them for help on that. Manual only for now, unfortunately :) but maybe it's fun to click through HipsterShop and see your activity displayed? haha

PR #289, opened 2020-08-07
Added logic to push-master.yml to create cloud image

### LINKED ISSUE #227: Cloud Shell images aren't updated on each commit
I believe this is being worked on, but adding an issue just in case

We [build and push new tagged images](https://github.com/GoogleCloudPlatform/stackdriver-sandbox/blob/master/.github/workflows/push-master.yml) for all services on each push to master. We should do the same with the new Cloud Shell image

eaberman:
Fixes #227 

Instead of creating new Cloud Image only on tags, a Custom Cloud Shell Image is also created on updates to master.

Essentially, the only change here is that the code written to Push Cloud Shell Images to GCR in `push-tags.yml` is also added to `push-master.yml`.

### DIFF
--- a/.github/workflows/push-master.yml
+++ b/.github/workflows/push-master.yml
@@ -36,3 +36,14 @@ jobs:
                        --tag=latest
       env:
         PROJECT_ID: ${{ secrets.PROJECT_ID }}
+        
+     - name: Push Cloud Shell Image to GCR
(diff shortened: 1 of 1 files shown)


Daniel-Sanche:
It looks like this PR broke the workflow. It's now giving an error due to the extra empty line

PR #301, opened 2020-08-10
emphasize workspace creation in install.sh

nicklin37:
Some users skip over the Monitoring Workspace creation because they aren't able to read the part in the script that asks them to create a monitoring workspace. 

Makes Workspace creation more visible to the user when running the install script and notifies them that there is something they need to do.

Hopefully this prevents users from accidentally skipping the Workspace creation.

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -214,8 +214,14 @@ installMonitoring() {
 
   gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
   if [[ -z $skip_workspace_prompt ]]; then
-    log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
-    read -p "When you are done, please press enter to continue"
+    YELLOW=`tput setaf 3`
+    log ""
+    log ""
+    log "${YELLOW}********************************************************************************"
+    log ""
+    log "${YELLOW}⚠️ Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
+    log ""
+    read -p "${YELLOW}When you are done, please PRESS ENTER TO CONTINUE"
   fi
 
   log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."


terraform/install.sh
```diff
@@ -214,7 +214,14 @@ installMonitoring() {
 
   gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
   if [[ -z $skip_workspace_prompt ]]; then
+    log ""
+    log ""
+    log "********************************************************************************"
+    log ""
+    log "ATTENTION! ACTION REQUIRED!"
+    log ""
     log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
```
simonz130, line 223: Can we change the color of the text here to yellow?

terraform/install.sh
```diff
@@ -214,7 +214,14 @@ installMonitoring() {
 
   gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
   if [[ -z $skip_workspace_prompt ]]; then
+    log ""
+    log ""
+    log "********************************************************************************"
+    log ""
+    log "ATTENTION! ACTION REQUIRED!"
```
Daniel-Sanche, line 221: I'd probably prefer an emoji instead of this: ⚠️

Or a mouse, to indicate action is required: 🖱️

Maybe in-line with the log text?

nicklin37: Added emoji instead!

terraform/install.sh
nicklin37: Text is now all yellow!

terraform/install.sh
```diff
@@ -214,8 +214,14 @@ installMonitoring() {
...
   gcp_monitoring_path="https://console.cloud.google.com/monitoring?project=$project_id"
   if [[ -z $skip_workspace_prompt ]]; then
-    log "Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
-    read -p "When you are done, please press enter to continue"
+    YELLOW=`tput setaf 3`
+    log ""
+    log ""
+    log "${YELLOW}********************************************************************************"
+    log ""
+    log "${YELLOW}⚠️Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
```
Daniel-Sanche, line 222: nit: I'd add an extra space between the emoji and the text

nicklin37: Added extra space :)

PR #302, opened 2020-08-10
Add integration test for Sandbox provisioning

jiang95-dev:
1. Added tests for GKE cluster
        a. Check if machine type is correct.
        b. Check if node number is as expected.
        c. Check if all services are ready.
        d. Check if the Hipster Shop returns 200 at request.
2. Added tests for load generator
        a. Check if there's only one load generator instance.
        b. Check if load generator and GKE cluster are in different zone.
        c. Check if the load generator returns 200 at request.
3. Added tests for resource provisioning
        a. Check if all APIs requested are available.
        b. Check if we can use Error Reporting API to report errors.


simonz130:
Your tests are more of integration tests than E2E tests.

E2E test simulate what customers would be doing. Current PR is testing provisioning (which is very important in itself)

tests/provision_test.py
```diff
@@ -0,0 +1,75 @@
...
+        client = cluster_manager.ClusterManagerClient()
+        cluster = client.get_cluster(name=self.__class__.name)
+        node_config = cluster.node_config
+        self.assertEqual(node_config.machine_type, 'n1-standard-2')
+
+    def testNumberOfNode(self):
+        """" Test if the number of nodes in the node pool is in the specified range """
+        client = cluster_manager.ClusterManagerClient()
+        cluster = client.get_cluster(name=self.__class__.name)
+        self.assertTrue(3 < cluster.current_node_count < 10)
```
simonz130, line 50: This is not a definitive test.
We should assert on a specific number of nodes, not a range.

tests/provision_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+import os
+import unittest
+import subprocess
+from shlex import split
+import json
+
+from google.cloud.container_v1.services import cluster_manager
+
+class TestGKECluster(unittest.TestCase):
```
simonz130, line 25: what will invoke this test?

Daniel-Sanche:
Is it possible to test the endpoints to make sure they're returning 200s?

tests/provision_test.py
jiang95-dev: There's an autoscaling in the config, so I guess the node number may not be definite in some circumstances?

tests/provision_test.py
jiang95-dev: "python3 provision_test.py" will invoke the tests.

jiang95-dev:
> Is it possible to test the endpoints to make sure they're returning 200s?

Only a few endpoints are external. How can I reach them?

tests/provision_test.py
simonz130: you should document it in readme.md.
Also please create a "provisioning" folder under "tests" and move your tests there

tests/provision_test.py
simonz130: Good point. When running provisioning tests on CI, I don't expect loadgen to create synthetic traffic to the cluster, so autoscaling rules won't apply

simonz130:
> > Is it possible to test the endpoints to make sure they're returning 200s?
> 
> Only a few endpoints are external. How can I reach them?

You can test that publicly facing endpoint of the hipstershop app returns 200 when you query it.
For loadgen, you can test that locust web UI is accessible

Daniel-Sanche:
> Only a few endpoints are external. How can I reach them?

You can find them using commands like this: https://github.com/GoogleCloudPlatform/stackdriver-sandbox/blob/8c0263cb0998e7a71612becd57b71f99a6d8da50/terraform/install.sh#L234

If you want to avoid doing that in Python code, you could make your script require them as arguments or something

tests/provisioning/README.md
```diff
@@ -0,0 +1,13 @@
...
+To run the integration test for the provisioning, follow the steps below.
+
+1. Install the dependencies
+```bash
+pip3 install -r requirements.txt
+```
+
+2. Run the tests
+```bash
+python3 provision_test.py
```
simonz130, line 12: This should be part of our CI @jiang95-dev

Daniel-Sanche: Yes, you should follow the patter Nick uses here: https://github.com/GoogleCloudPlatform/stackdriver-sandbox/pull/335

tests/provisioning/README.md
```diff
@@ -0,0 +1,13 @@
+# Tests for Stackdriver Sandbox Provisioning
```
simonz130, line 1: Tests for Cloud Ops Sandbox Provisioning

.github/workflows/e2e.yml
```diff
@@ -61,6 +62,15 @@ jobs:
           test-cloud-shell:$GITHUB_SHA
       env:
         PROJECT_ID: ${{ secrets.E2E_PROJECT_ID  }}
+    - name: Run Provisioning Test
+      run: |
+        set -x
+        # install dependencies
+        python3 -m pip install -r tests/provisioning/requirements.txt
+        # run tests
+        python3 tests/provisioning/provision_test.py $PROJECT_ID
```
simonz130, line 71: Consider running the tests in docker. This way when the finish, you don't need to  cleanup the state of the github worker VM - you just remove the docker container.

tests/provisioning/provision_test.py
```diff
@@ -0,0 +1,119 @@
...
+class TestGKECluster(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        """ Get the location of GKE cluster for later queries """
+        command = ('gcloud container clusters list --format="value(location)"')
+        result = subprocess.run(split(command), encoding='utf-8', capture_output=True)
+        cls.location = str(result.stdout.replace('\n', ''))
+        if cls.location == '':
+            raise Exception('GKE cluster is not provisioned!')
+        cls.name = 'projects/' + getProjectId() + '/locations/' + cls.location + '/clusters/stackdriver-sandbox'
```
simonz130, line 37: cloud-ops-sandbox

tests/provisioning/provision_test.py
```diff
@@ -0,0 +1,119 @@
...
+        """ Test if all the service deployments are ready """
+        command = ("kubectl get deployment --all-namespaces -o json")
+        result = subprocess.run(split(command), encoding='utf-8', capture_output=True)
+        services = json.loads(result.stdout)
+        for service in services['items']:
+            self.assertTrue(int(service['status']['readyReplicas']) >= 1)
+    
+    def testReachOfFrontend(self):
+        """ Test if query frontend returns 200 """
+        command_ip = ("kubectl get svc frontend-external -o json")
```
simonz130, line 62: We install Istio on the cluster, so getting the external IP of the cluster is not by "kubectl get svc frontend-external". Look at Nic's PRs for adding istio - there is a different command that returns external IP

tests/provisioning/provision_test.py
```diff
@@ -0,0 +1,119 @@
...
+        command_ip = ('gcloud compute instances list --filter="name:loadgenerator*" --format="value(networkInterfaces[0].accessConfigs.natIP)"')
+        result_ip = subprocess.run(split(command_ip), encoding='utf-8', capture_output=True)
+        url = "".join(['http://', result_ip.stdout.split('\n')[0], ':8080'])
+        self.assertTrue(urllib.request.urlopen(url).getcode() == 200)
+
+class TestProjectResources(unittest.TestCase):
+    def testCloudTrace(self):
+        """ Test if Cloud Trace is provisioned """
+        try:
+            command = ('gcloud alpha trace sinks list --project={0}'.format(getProjectId()))
```
simonz130, line 82: What does this command return? Is it returning the traces that were captured in a project?
Also, if this is not available in gcloud beta or gcloud (ga), make sure you do
```
gcloud components alpha install
```
before the tests start

simonz130:
Your e2e test fails with:
```
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'tests/provisioning/requirements.txt'

```

tests/provisioning/provision_test.py
nicklin37: Here is the command for that:
kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'

simonz130:
@jiang95-dev, I suggest that you address the CR feedback iteratively in order to accelerate the feedback loop.
For example, you can address rebranding remarks first, while taking more time to address running the tests in a container.

tests/provisioning/provision_test.py
jiang95-dev: How should I test the reachablity? I can't open the url for istio, but I can open the url for the frontend.

PR #304, opened 2020-08-10
End-to-End CI test

Daniel-Sanche:
This PR adds a new end-to-end phase to our CI tests. The test does the following steps:
- clean up test project to remove any leftover resources
- build a fresh set of containers (including cloud shell)
- re-write manifests to deploy new containers instead of last pushed
- run install.sh inside our custom cloud shell container environment
- clean up test project to remove any leftover resources

From here, we can add tests that check the state of the project after running install.sh, to make sure all the resources we expected are properly set up

Open questions:
- How often should this test run? (I currently set it to run on each commit to master and each release branch. But it is a long test (~30 mins end-to-end), so it may be better to do periodically)
- Am I missing anything from `clean.sh`? I'm not sure if I have a full grasp on all the resources being spun up (@nicklin37)

I also had to make a couple changes to the terraform/install scripts to make the billing id optional (Thanks @nicklin37  for your help here!)


.github/workflows/e2e_scripts/run_install.sh
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# This file sets up and executes an installation inside the
+# custom Cloud Shell container
+
```
eaberman, line 19: LGTM! Thanks for adding this!

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,70 @@
...
+        set -x
+        # build cloud shell image
+        docker build -t custom-cloud-shell:$GITHUB_SHA ./cloud-shell
+        # run install script
+        docker run --rm \
+          -v ~/.config:/root/.config \
+          -v `pwd`:/sandbox-shared \
+          --entrypoint /sandbox-shared/.github/workflows/e2e_scripts/run_install.sh \
+          custom-cloud-shell:$GITHUB_SHA \
+          -p $PROJECT_ID --skip-workspace-prompt
```
eaberman, line 63: LGTM.

.github/workflows/e2e_scripts/clean.sh
```diff
@@ -0,0 +1,101 @@
...
+# clear notification channels
+CHANNEL_LIST="first_run"
+while [ -n "$CHANNEL_LIST" ]; do
+  CHANNEL_LIST=$(gcloud alpha monitoring channels list --format="value(name)")
+  if [ -n "$CHANNEL_LIST" ]; then
+    for CHANNEL in $CHANNEL_LIST; do
+      gcloud alpha monitoring channels delete $CHANNEL --quiet
+    done
+  fi
+done
```
eaberman, line 101: Is there anything in the `clean.sh` script that needs to be done to exit out of the custom image and delete other resources made by running `./install.sh`?

.github/workflows/e2e_scripts/clean.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# clear logs
+for LOG in $(gcloud logging logs list --project $PROJECT_ID --format="value(NAME)"); do
+    echo "deleting $LOG..."
+    gcloud logging logs delete $LOG --project $PROJECT_ID --quiet
+done
+
+# clear dashboards
```
nicklin37, line 70: During Monitoring provisioning I also create the following
1) Custom Services
2) Uptime Check
I think these will also need to be cleaned up?

.github/workflows/e2e_scripts/clean.sh
Daniel-Sanche: clean.sh isn't actually run inside the custom cloud shell container, it just runs straight on the worker. Let me know if you see anything the script is missing

.github/workflows/e2e_scripts/clean.sh
Daniel-Sanche: Do you know if there are gcloud commands to manage those?

nicklin37: Just checked - there isn't gcloud integration for either :(

nicklin37: There is REST and client library support for both though

Daniel-Sanche: Hmm ok, maybe we can add a curl command or something to handle that instead. Do you want to look into that as part of your monitoring tests later?

nicklin37: Yeah that sounds good!

.github/workflows/README.md
```diff
@@ -53,11 +63,27 @@ Some actions require authentication to GCP to push container images. These worke
...
+
+### E2E.yaml
+
+#### Triggers
+- on each commit to master
+- on each commit to a release branch
+- on each PR into master
+
+#### Actions
+- ensure end-to-end test project has deleted all test resources
```
simonz130, line 84: The E2E test should run once or twice daily but not on each PR or each commit to master. If we run it on each PR, we'll need to handle multiple runs in parallel and also reduce velocity.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,70 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "End to End Tests"
+on:
+  push:
+    # run on pushes to master or release/*
+    branches:
+      - master
```
simonz130, line 20: see my other comment on this

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,70 @@
...
+jobs:
+  end-to-end:
+    runs-on: [self-hosted, e2e-worker]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Clean Project State
+      run: |
+        # clean GCP resources
+        .github/workflows/e2e_scripts/clean.sh
```
simonz130, line 34: can we reuse destroy.sh?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,70 @@
...
+        # tag with git hash
+        skaffold build --default-repo=gcr.io/$PROJECT_ID \
+                       --tag=$GITHUB_SHA
+      env:
+        PROJECT_ID: ${{ secrets.E2E_PROJECT_ID  }}
+    - name: Rewrite Manifests
+      run: |
+        set -x
+        OLD_PATH=gcr.io/stackdriver-sandbox-230822
+        NEW_PATH=gcr.io/stackdriver-sandbox-e2e/$OLD_PATH
```
simonz130, line 48: please indicate "Tests" in the name

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,70 @@
...
+        # use newly built containers instead of latest pushed
+        find "./kubernetes-manifests" -name '*.yaml' -exec sed -i -e "s#$OLD_PATH#$NEW_PATH#g" {} \;
+        find "./kubernetes-manifests" -name '*.yaml' -exec sed -i -e "s#:latest#:$GITHUB_SHA#g" {} \;
+    - name: Run Install Script
+      run: |
+        set -x
+        # build cloud shell image
+        docker build -t custom-cloud-shell:$GITHUB_SHA ./cloud-shell
```
simonz130, line 56: indicate "tests" in name

.github/workflows/e2e_scripts/run_install.sh
```diff
@@ -0,0 +1,25 @@
...
+# copy sandbox files into a new directory outside the
+# docker volume, to avoid persisting tfstate between runs
+mkdir /sandbox
+cp -r /sandbox-shared/. /sandbox
+# run install script
+bash -x /sandbox/terraform/install.sh $*
```
simonz130, line 25: This way we don't exercise the customization we added for running install.sh 
We should run the install.sh from the custom container image using the customization we added to cloudshell-open command.

.github/workflows/README.md
Daniel-Sanche: I changed it to run on each commit to master, as we discussed in the meeting. I think that's a good first step, and if that's causing issues we can move it to periodically

.github/workflows/e2e.yml
Daniel-Sanche: No, destroy.sh is simple because it wipes the entire project. We need to keep the project, but individually delete resources

.github/workflows/e2e.yml
Daniel-Sanche: `gcr.io/stackdriver-sandbox-e2e/tests/gcr.io/stackdriver-sandbox-230822`?

Note that this is in a different project than the canonical images, so we don't have to worry about them getting mixed up

.github/workflows/e2e.yml
Daniel-Sanche: This image isn't being pused, it's just for local use. Do you still want me to change the name?

.github/workflows/e2e_scripts/run_install.sh
Daniel-Sanche: Hmm my first thought was that we can't run it directly through `cloudshell_open_cp.sh`, because we need to run install.sh with specific arguments. But on second thought it may work fine as long as we set the environment variables correctly. I'll see if I can get it to work

Daniel-Sanche: Ok, it should be executed through cloudshell_open now

.github/workflows/e2e.yml
simonz130: yeah, this name works

.github/workflows/e2e.yml
simonz130: yes

.github/workflows/e2e_scripts/clean.sh
```diff
@@ -0,0 +1,101 @@
...
+      echo "deleting cluster"
+      gcloud container clusters delete stackdriver-sandbox \
+          --project $PROJECT_ID --zone $CLUSTER_ZONE --quiet
+      sleep 20
+  fi
+done
+
+
+# delete loadgenerator
+LOADGENRATOR="first_run"
```
simonz130, line 40: nit: typo - LOADGENERATOR

.github/workflows/e2e_scripts/clean.sh
```diff
@@ -0,0 +1,101 @@
...
+while [ -n "$LOADGENRATOR" ]; do
```
simonz130, line 41: Maybe worth moving this script to be in the same place where destroy.sh is. This way we might even offer users the ability to clean the project rather than destroy it in the future.

simonz130: Also, making changes in provisioning will be easier - less change we'll miss appropriate updates that also need to be done in clean.sh

.github/workflows/e2e_scripts/clean.sh
simonz130: Lack of gcloud for these is a big DevX gap, however it should be possible to destroy with ***terraform destroy***

.github/workflows/e2e_scripts/clean.sh
```diff
@@ -0,0 +1,101 @@
...
+    for DASHBOARD in $DASHBOARD_LIST; do
+      gcloud monitoring dashboards delete $DASHBOARD --quiet
+    done
+  fi
+done
+
+# clear policies
+POLICY_LIST="first_run"
+while [ -n "$POLICY_LIST" ]; do
+  POLICY_LIST=$(gcloud alpha monitoring policies list --format="value(name)")
```
simonz130, line 84: is it available only in alpha feed?

.github/workflows/e2e_scripts/clean.sh
Daniel-Sanche: 1. Will terraform destroy destroy the project as well?
2. I was hoping to avoid terraform destroy, because that relies on the existence of the tfstate file. I want to make sure `clean.sh` is reliable and idempotent, and not rely on existing state

.github/workflows/e2e_scripts/clean.sh
Daniel-Sanche: Yeah, it looks like it

.github/workflows/e2e_scripts/clean.sh
Daniel-Sanche: I wrote this for test automation in mind, it isn't currently safe for user input. Right now, it will automatically wipe the currently active project (even if it isn't a sandbox project!). If we move it to the user-facing folder, I think I should add some prompts and guardrails first. Do you think that's worth doing?

.github/workflows/e2e.yml
Daniel-Sanche: done

.github/workflows/e2e.yml
Daniel-Sanche: done

.github/workflows/e2e_scripts/clean.sh
nicklin37: 1) Currently monitoring terraform is separated so it will not destroy the project. however if we move them to be together then yes it would try to destroy the project.

simonz130: If there is no better solution currently available for removing all monitoring resources, it's better to use *terraform destroy* and not curl-based API calls but anyway, this doesn't have to be part of this PR

.github/workflows/e2e_scripts/clean.sh
simonz130: Let's keep it for now in the test automation folder but we need to be extra careful in code reviews - if we add more objects that needs to be provisioned, we also need to handle their cleanup here.

PR #312, opened 2020-08-11
OpenTelemetry Tracing for Currency Service

### LINKED ISSUE #311: Add Tracing to Currency Service
...and use OpenTelemetry for it!

Subissue of #132. Currency Service currently does not have any telemetry instrumentation, so this is not a "conversion" task but I'm still grouping it under the same project.

gzhao408:
Resolves #311, subtask of #132 

Currency service acts as a server to Frontend (GetSupportedCurrencies() and Convert()) and the Checkout service in PlaceOrder (Convert()).

To manually test this branch, you can see it from (1) home page or (2) at checkout.

Frontend uses `GetSuportedCurrencies` to show the menu of currencies; the corresponding trace should show both `client-hipstershop.currencyService` (which comes from frontend) and `grpc.hipstershop.CurrencyService` (which comes from CurrencyService itself and means the context was properly propagated through grpc). 
AND It should also show the HTTP call to the ECB website 💶  🤑  💸  💵 
image

For Checkout, you need to add an item to the cart and click check out. Checkout.PlaceOrder() dispatches 6 calls to the other services, one of which is the CurrencyService to Convert the total price. The trace should show all 6 calls from the client (checkout service)`Span.client-hipstershop`. This should already show on current master:
image

On this branch, there should also be additional information from the Currency service itself. Clicking the carrot and checking Show Events should show more detail, indicating that the CurrencyService indeed receives the trace context and starts a child span `grpc.hipstershop.CurrencyService/Convert`:
image
image

### DIFF
--- a/src/currencyservice/package.json
+++ b/src/currencyservice/package.json
@@ -16,6 +16,8 @@
     "@opentelemetry/api": "^0.10.2",
     "@opentelemetry/node": "^0.10.2",
     "@opentelemetry/plugin-grpc": "^0.10.2",
+    "@opentelemetry/plugin-http": "^0.10.2",
+    "@opentelemetry/plugin-https": "^0.10.2",
     "@opentelemetry/tracing": "^0.10.2",
     "async": "1.5.2",
     "google-protobuf": "3.12.4",
--- a/src/currencyservice/package.json
+++ b/src/currencyservice/package.json
@@ -10,8 +10,12 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@google-cloud/debug-agent": "3.2.0",
+    "@google-cloud/opentelemetry-cloud-trace-exporter": "^0.4.1",
     "@google-cloud/profiler": "4.0.2",
     "@grpc/proto-loader": "0.5.5",
+    "@opentelemetry/api": "^0.10.2",
+    "@opentelemetry/node": "^0.10.2",
+    "@opentelemetry/tracing": "^0.10.2",
     "async": "1.5.2",
     "google-protobuf": "3.12.4",
     "grpc": "1.24.3",
--- a/src/currencyservice/package.json
+++ b/src/currencyservice/package.json
@@ -15,6 +15,7 @@
     "@grpc/proto-loader": "0.5.5",
     "@opentelemetry/api": "^0.10.2",
     "@opentelemetry/node": "^0.10.2",
+    "@opentelemetry/plugin-grpc": "^0.10.2",
     "@opentelemetry/tracing": "^0.10.2",
     "async": "1.5.2",
     "google-protobuf": "3.12.4",
--- a/src/currencyservice/package.json
+++ b/src/currencyservice/package.json
@@ -11,7 +11,6 @@
   "dependencies": {
     "@google-cloud/debug-agent": "3.2.0",
     "@google-cloud/profiler": "4.0.2",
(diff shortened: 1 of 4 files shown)


gzhao408:
discovered an auth issue, making a draft while i figure it out

src/currencyservice/server.js
```diff
@@ -116,16 +138,28 @@ function _carry (amount) {
  * Lists the supported currencies
  */
 function getSupportedCurrencies (call, callback) {
+  const currentSpan = tracer.getCurrentSpan();
+  const span = tracer.startSpan('currencyservice:GetSupportedCurrencies()', {
+    parent: currentSpan,
+    kind: 1, //server
```
simonz130, line 144: why it's server?

src/currencyservice/server.js
```diff
@@ -116,16 +138,28 @@ function _carry (amount) {
...
 }
 
 /**
  * Converts between currencies
  */
 function convert (call, callback) {
+  const currentSpan = tracer.getCurrentSpan();
+  const span = tracer.startSpan('currencyservice:Convert()', {
+    parent: currentSpan,
+    kind: 1, //server
```
simonz130, line 161: ditto

src/currencyservice/server.js
gzhao408: Currency Service is the server and whoever is calling this function (i.e. frontend or checkout) is the client

Daniel-Sanche:
It looks like the Ci is failing because emailservice and recommendationservice are crashing without GCP credentials. I'm not sure why that would be impacting this PR and not master though...

gzhao408:
Rebased after #320 wooooo

gzhao408:
I removed the @google-cloud/trace-agent because it was messing with the opentelemetry tracing, I think the exporter gets confused about which traces it's supposed to use, or one overrides the other or something. It was also the source of the "PROJECT_ID not found" warning. We don't need it 🤷

src/currencyservice/package.json
```diff
@@ -10,9 +10,14 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@google-cloud/debug-agent": "3.2.0",
+    "@google-cloud/opentelemetry-cloud-trace-exporter": "^0.4.1",
     "@google-cloud/profiler": "4.0.2",
-    "@google-cloud/trace-agent": "2.11.0",
+    "@google-cloud/trace-agent": "5.1.0",
```
simonz130, line 15: is this needed now that we use opentelemetry tracing?

gzhao408: :'((((( yeah i tried to delete it... and suddenly thousands of `grpc.google.devtools.cloudtrace.v2.TraceService/BatchWriteSpans` traces were coming out of nowhere, which caused the quota exceeded and probably the CI error. still trying to figure out what's going on, so i'm looking to see what variables i can turn on/off

gzhao408:
Sorry for the many many force pushes. I think I've resolved the spammy batchwrite issue by switching out the processor and flushing the exporter, in doing so, accidentally enabled a new feature - the traces are so amazing that we can now track the calls to the ECB website:

image

Also I got rid of @google-cloud/trace-agent.

src/currencyservice/package.json
gzhao408: alright it's GONE

PR #313, opened 2020-08-11
Added test to CI to Build Custom Cloud Shell Image

### LINKED ISSUE #215: Create tests for Custom Cloud Image
Tests should be added to CI to ensure nothing breaks whenever changes are made to the custom cloud image.

The Custom Cloud Image is created from a Dockerfile stored in the project repo (`cloud-shell/Dockerfile`).

eaberman:
Fixes #215 
- Creates `docker build` command to `ci.yml` to build the image
- Specifies that if build does not work, it exits w/code 1

One down-side of adding this to CI is that building the image takes a long time. This is mainly due to the fact that it's being pulled from the Cloud Shell Base Image (a LARGE file). It took approximately 20 minutes for the CI pipeline to run within this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,3 +87,8 @@ jobs:
         if [[ "$ERROR_COUNT" -gt "0" ]]; then
           exit 1
         fi
+    - name: Test Custom Cloud Shell Image Build
+      run: |
+        set -x
+        # build cloud shell image
+        docker build -t custom-cloud-shell:$GITHUB_SHA ./cloud-shell
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,11 @@ jobs:
         if [[ -n $(git status -s) ]]; then
           exit 1
         fi
+    - name: Test Custom Cloud Shell Image Build
+      run: |
+        set -x
+        # build cloud shell image: it will automatically exit if there's an error in build
+        docker build -t cloud-shell ./cloud-shell
     - name: Setup Cluster
       run: |
         set -x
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -87,3 +87,8 @@ jobs:
         if [[ "$ERROR_COUNT" -gt "0" ]]; then
           exit 1
         fi
+    - name: Test Custom Cloud Shell Image Build
```
Daniel-Sanche, line 90: Short tests like this should go near the top so we can fail fast if anything goes wrong. This step should be right after the license check

.github/workflows/ci.yml
```diff
@@ -87,3 +87,8 @@ jobs:
...
+      run: |
+        set -x
+        # build cloud shell image: it will automatically exit if there's an error in build
+        docker build -t custom-cloud-shell:$GITHUB_SHA ./cloud-shell
```
Daniel-Sanche, line 94: Nit: It seems a bit misleading to label the container with `:$GITHUB_SHA`. That would make me think the container is being saved/used somewhere, but the real purpose of this step is the build process, not the artifact itself. 

Maybe use `cloud-shell`, or even `_`?

Daniel-Sanche:
> One down-side of adding this to CI is that building the image takes a long time. This is mainly due to the fact that it's being pulled from the Cloud Shell Base Image (a LARGE file). It took approximately 20 minutes for the CI pipeline to run within this PR.

Luckily, this should only happen the first time. The container layers will be cached and much faster going forward. Although we may have to re-size the worker disks soon to better handle the new giant containers

PR #316, opened 2020-08-12
SLO and Alerting Policies for Custom Services

### LINKED ISSUE #74: Create terraform for SLOs and alerting policies for checkoutservice
#WHAT: Create two SLOs for the checkoutservice (1 for availability and 1 for performance) and an alerting policy on budget burn for each one. SLO defined as 99.9% availability/performance over 30 day rolling window. Alert when 10% budget expended over 1 hour.

#WHY: Demonstrate best practice alerting on SLO error budget burn through terraform for this service

#HOW: Terraform resources added into initialization terraform scripts

nicklin37:
#WHAT: Creates Custom Services for 5 of the 10 microservices and then tether availability and latency SLOs/alerting policies onto the custom service. 
  - previously intended to use template file, but it's not suitable for our needs here so we use the count functionality to iterate through the list of services and create all necessary resources

#WHY: We want SLOs and alerting policies for each service - here we demonstrate how to use a custom service through Terraform to create these. 

#HOW: Terraform count utility lets us create repeated resources.

#TESTING: Added tests for each resource that is created. All tests pass. PR is ready for merge after test are fully integrated into CI.

Closes #74, #75, #76, #77


terraform/monitoring/03_service_slo.tf
```diff
@@ -12,31 +12,92 @@
...
+      service_id = "shippingservice"
+      availability_goal = 0.99
+      availability_burn_rate = 2
+      latency_goal = 0.99
+      latency_threshold = 500
+      latency_burn_rate = 2
+    }
+  ]
+}
+
```
eaberman, line 75: I noticed that a lot of these numbers for `latency_threshold` or `availability_burn_rate` are the same across services. Would it make sense to define constants and use that for each of the services?

terraform/monitoring/04_alerting_policies.tf
```diff
@@ -12,50 +12,51 @@
...
+    display_name = "SLO burn rate alert for availability SLO with a threshold of ${var.custom_services[count.index].availability_burn_rate}"
     condition_threshold {
 
       # This filter alerts on burn rate over the past 60 minutes
-      filter     = "select_slo_burn_rate(\"projects/${var.project_id}/services/${google_monitoring_custom_service.frontend.service_id}/serviceLevelObjectives/${google_monitoring_slo.frontend_availability_slo.slo_id}\", 60m)"
+      filter     = "select_slo_burn_rate(\"projects/${var.project_id}/services/${google_monitoring_custom_service.custom_service[count.index].service_id}/serviceLevelObjectives/${google_monitoring_slo.custom_service_availability_slo[count.index].slo_id}\", 60m)"
       
-      # The threshhold is determined by how quickly we burn through our error budget, and we alert if we will use our error budget within 15 days.
+      # The threshhold is determined by how quickly we burn through our error budget.
+      # Example: If Threshold = 2 then error budget is consumed in 15 days.
```
eaberman, line 30: Nice documentation with example!

nicklin37:
Converting to draft while I refactor to look like #322

terraform/monitoring/03_service_slo.tf
nicklin37: Yeah the numbers are pretty similar - It's definitely possible to use a constant for them but it would require data sources in terraform which would clutter the config. I think also keeping the values like this allows us to change the values later and shows customers how to choose different values for each service.

If there is strong support towards consolidating towards constants then please let me know!

terraform/monitoring/04_alerting_policies.tf
nicklin37: Thanks!!

nicklin37:
Done with refactor so ready for review again.

terraform/monitoring/03_services.tf
```diff
@@ -0,0 +1,88 @@
...
+    latency_threshold = number,
+    latency_burn_rate = number
+  }))
+  default = [
+    {
+      service_name = "Frontend Service"
+      service_id = "frontend"
+      availability_goal = 0.9		# configurable goal for the availability SLO (0.9 = 90% of requests are successsful)
+      availability_burn_rate = 2	# limit on error budget burn rate (2 indicates we alert if error budget is consumed 2x faster than it should)
+      latency_goal = 0.9			# configurable goal for the latency SLO (0.9 = 90% of requests finish in under the latency threshold)
```
simonz130, line 34: nit: probably a glitch of 1 tab here compared to adjacent lines with comments

terraform/monitoring/03_services.tf
```diff
@@ -0,0 +1,88 @@
...
+    availability_burn_rate = number,
+    latency_goal = number,
+    latency_threshold = number,
+    latency_burn_rate = number
+  }))
+  default = [
+    {
+      service_name = "Frontend Service"
+      service_id = "frontend"
+      availability_goal = 0.9		# configurable goal for the availability SLO (0.9 = 90% of requests are successsful)
```
simonz130, line 32: perhaps worth adding a note about why we set availability goal of only 1 nine. Typically it's 2 nines (99) as minimum, 99.9 as stretch goal and 99.999 as north star for extremely highly available services (not applicable to our demo)

PR #317, opened 2020-08-12
Added cart service with managed grpc in c#

simonz130:
- Updated .Net implementation for cart service to use managed grpc hosted on asp.net core. Previously it was using unmanaged grpc stack (c-based). The inputs/outputs didn't change but the underlying stack hosting this did.
- Optimized container image size for cart service 10x from over 1.8Gb to ~189Mb.

This is deployed here: http://34.82.11.133/

This work is needed because with the merge of OpenCensus to OpenTelemetry and that .net core 3.1 being the current stack developers are using, the instrumentation in c# for the OT assumes using [managed grpc stack](https://docs.microsoft.com/en-us/aspnet/core/grpc/?view=aspnetcore-3.1).


src/cartservice/Protos/Cart.proto
```diff
@@ -0,0 +1,51 @@
...
+service CartService {
+    rpc AddItem(AddItemRequest) returns (Empty) {}
+    rpc GetCart(GetCartRequest) returns (Cart) {}
+    rpc EmptyCart(EmptyCartRequest) returns (Empty) {}
+}
+
+message CartItem {
+    string product_id = 1;
+    int32  quantity = 2;
+}
```
Daniel-Sanche, line 30: These seem to be a duplicate of the [protos in `/pb`](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/master/pb/demo.proto). Shouldn't we use those instead?

Issue #267, opened 2020-08-05
Reusing existing Sandbox should check tfstate before using


nicklin37:
When a user attempts to reuse a pre-existing Sandbox

Then the install script should check the GCS bucket and verify that the tfstate file is still valid to be used (i.e. the Sandbox is not out of date)

Otherwise it should reject the request and create a new one.

jiang95-dev:
Due to the rebranding, we don't need to take care of the logic now.

PR #332, opened 2020-08-14
check if sandbox is outdated

jiang95-dev:
If there's no linked bucket to the Sandbox user chooses, then it's the old version of Sandbox, and we should ask user to choose again.

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -86,7 +86,7 @@ promptForProject() {
                  --flatten="bindings[].members" \
                  --format="table(bindings.members)" \
                  --filter="bindings.role:roles/owner" 2> /dev/null | grep $acct | cat)
-      if [[ -n "$iam_test" ]]; then
+    if [[ -n "$iam_test" ]]; then
       create_time=$(gcloud projects describe "$proj" --format="value(create_time.date(%b-%d-%Y))")
       found_projects+=("$proj | [$create_time]")
     fi
@@ -112,7 +112,14 @@ promptForProject() {
           IFS=$' |'
           opt=($opt)
           project_id=${opt[0]}
-          break
+          # check if this project has a bucket
+          bucket_name="$project_id-bucket"
+          gcloud config set project "$project_id"
+          if [[ -n "$(gsutil ls | grep gs://$bucket_name/)" ]]; then
+            break
+          else
+            log "This is an outdated project that doesn't support GCS. Please try to use a new one."
+          fi
         fi
       done
       IFS=$IFS_bak


terraform/install.sh
```diff
@@ -86,7 +86,7 @@ promptForProject() {
                  --flatten="bindings[].members" \
                  --format="table(bindings.members)" \
                  --filter="bindings.role:roles/owner" 2> /dev/null | grep $acct | cat)
-      if [[ -n "$iam_test" ]]; then
+    if [[ -n "$iam_test" ]]; then
```
Daniel-Sanche, line 89: was this intentional?

terraform/install.sh
```diff
@@ -112,7 +112,14 @@ promptForProject() {
           IFS=$' |'
           opt=($opt)
           project_id=${opt[0]}
-          break
+          # check if this project has a bucket
+          bucket_name="$project_id-bucket"
```
Daniel-Sanche, line 116: we shouldn't be setting this in multiple places in the code, in case we want to change it later.

terraform/install.sh
jiang95-dev: Oh I just want to align the indentations.

terraform/install.sh
jiang95-dev: So create a temporary var?

Daniel-Sanche:
Is there a bug/issue that motivated this? What currently happens when a user selects an outdated project?

If we we're dropping support for older projects, maybe this is a good chance to just change the name filter to "operations-sandbox-*" and ignore all legacy stackdriver sandbox projects

terraform/install.sh
```diff
@@ -112,7 +112,14 @@ promptForProject() {
           IFS=$' |'
           opt=($opt)
           project_id=${opt[0]}
-          break
```
Daniel-Sanche, line 115: is the break no longer needed?

terraform/install.sh
Daniel-Sanche: somehow refactor so that the variable is only set once in the script. Maybe even just declare it at the top?

jiang95-dev:
> Is there a bug/issue that motivated this? What currently happens when a user selects an outdated project?
> 
> If we we're dropping support for older projects, maybe this is a good chance to just change the name filter to "operations-sandbox-*" and ignore all legacy stackdriver sandbox projects

Yes. This is linked to issue #267.
Users will be asked to choose another project.
So all the names and ids will become "operations-sandbox" later on?

terraform/install.sh
jiang95-dev: it's in the if-else branch now.

terraform/install.sh
Daniel-Sanche: sounds good

terraform/install.sh
jiang95-dev: But I can only check the bucket when a project is selected, so I can't know it beforehand. Might as well create a temp var.

Daniel-Sanche: Hmm good point. What do you mean by a temp var?

simonz130: Why not creating a function that gets projectId as input and returns a bucker name back and then use it everywhere where it's needed?

terraform/install.sh
```diff
@@ -112,7 +112,14 @@ promptForProject() {
...
+          # check if this project has a bucket
+          bucket_name="$project_id-bucket"
+          gcloud config set project "$project_id"
+          if [[ -n "$(gsutil ls | grep gs://$bucket_name/)" ]]; then
+            break
+          else
+            log "This is an outdated project that doesn't support GCS. Please try to use a new one."
```
simonz130, line 121: This project uses old version of Sandbox not compatible with stateful provisioning. Please create a new project instead.

simonz130:
> If there's no linked bucket to the Sandbox user chooses, then it's the old version of Sandbox, and we should ask user to choose again.

You should start your PR description with "Fixes #267".  This way when you merge thiis PR, corresponding bug will be auto-closed by github.

simonz130:
> Is there a bug/issue that motivated this? What currently happens when a user selects an outdated project?
> 
> If we we're dropping support for older projects, maybe this is a good chance to just change the name filter to "operations-sandbox-*" and ignore all legacy stackdriver sandbox projects

I like this idea! As part of rebranding we can update the filter throughout!

jiang95-dev:
Due to rebranding, this pull request is not necessary now.

PR #335, opened 2020-08-14
Monitoring Integration Test added to E2E CI

nicklin37:
#WHAT: Integrates the monitoring integration tests into the e2e tests of the CI pipeline.
   - Provisions monitoring examples with provisioning of the sandbox
   - Checks for successful provisioning
   - cleans up all provisioned monitoring resources using the API

#WHY: We want to automatically run the monitoring integration tests.

#HOW: Added Python script to deal with resource deletion, Python script for testing linked to the CI.

#TESTING: Manual test of deletion script works and same with monitoring integration test.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -68,7 +68,9 @@ jobs:
         # install dependencies
         python3 -m pip install -r tests/requirements.txt
         # run tests
-        python3 tests/monitoring_integration_test.py ${{ secrets.E2E_PROJECT_ID }}
+        python3 tests/monitoring_integration_test.py $PROJECT_ID
+      env:
+        PROJECT_ID: ${{ secrets.E2E_PROJECT_ID }}
     - name: Clean Project State
       if: ${{ always()  }}
       run: |
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -66,9 +66,9 @@ jobs:
       run: |
         set -x
         # install dependencies
-        pip3 install -r tests/requirements.txt
+        python3 -m pip install -r tests/requirements.txt
         # run tests
-        python3 tests/monitoring_integration_test.py
+        python3 tests/monitoring_integration_test.py ${{ secrets.E2E_PROJECT_ID }}
     - name: Clean Project State
       if: ${{ always()  }}
       run: |
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -20,6 +20,7 @@ on:
       - master
       - release/*
       - e2e-test
+      - monitoring_tests_to_e2e
 jobs:
   end-to-end:
     runs-on: [self-hosted, e2e-worker]
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -20,7 +20,6 @@ on:
       - master
       - release/*
       - e2e-test
-      - monitoring_tests_to_e2e
 jobs:
   end-to-end:
     runs-on: [self-hosted, e2e-worker]
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -70,7 +70,7 @@ jobs:
         # run tests
         python3 tests/monitoring_integration_test.py $PROJECT_ID
       env:
-        PROJECT_ID: ${{ secrets.E2E_PROJECT_ID }}
+        PROJECT_ID: ${{ secrets.E2E_PROJECT_ID  }}
     - name: Clean Project State
       if: ${{ always()  }}
       run: |
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -61,6 +61,13 @@ jobs:
           test-cloud-shell:$GITHUB_SHA
       env:
         PROJECT_ID: ${{ secrets.E2E_PROJECT_ID  }}
+    - name: Run Monitoring Integration Tests
+      run: |
+        set -x
+        # install dependencies
+        pip3 install -r tests/requirements.txt
+        # run tests
+        python3 tests/monitoring_integration_test.py
     - name: Clean Project State
       if: ${{ always()  }}
       run: |
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -67,6 +67,9 @@ jobs:
         set -x
         # install dependencies
         python3 -m pip install -r tests/requirements.txt
+        # authenticate cluster
+        CLUSTER_ZONE=$(gcloud container clusters list --filter="name:cloud-ops-sandbox" --project $PROJECT_ID --format="value(zone)")
+        gcloud container clusters get-credentials cloud-ops-sandbox --zone "$CLUSTER_ZONE"
         # run tests
         python3 tests/monitoring_integration_test.py $PROJECT_ID
       env:
--- a/.github/workflows/e2e_scripts/clean.sh
+++ b/.github/workflows/e2e_scripts/clean.sh
@@ -20,6 +20,7 @@ set +e
 # Cloud Operations Sandbox project in between test runs
 
 export PROJECT_ID=$(gcloud config get-value project)
+export WORKDIR=$(dirname $(realpath $0))
 
 # delete cluster
 CLUSTER_ZONE="first_run"
@@ -67,35 +68,6 @@ for LOG in $(gcloud logging logs list --project $PROJECT_ID --format="value(NAME
     gcloud logging logs delete $LOG --project $PROJECT_ID --quiet
 done
 
-# clear dashboards
-DASHBOARD_LIST="first_run"
-while [ -n "$DASHBOARD_LIST" ]; do
-  DASHBOARD_LIST=$(gcloud monitoring dashboards list --format="value(name)")
-  if [ -n "$DASHBOARD_LIST" ]; then
-    for DASHBOARD in $DASHBOARD_LIST; do
-      gcloud monitoring dashboards delete $DASHBOARD --quiet
-    done
-  fi
-done
-
-# clear policies
-POLICY_LIST="first_run"
-while [ -n "$POLICY_LIST" ]; do
-  POLICY_LIST=$(gcloud alpha monitoring policies list --format="value(name)")
-  if [ -n "$POLICY_LIST" ]; then
-    for POLICY in $POLICY_LIST; do
-      gcloud alpha monitoring policies delete $POLICY --quiet
-    done
-  fi
-done
-
-# clear notification channels
-CHANNEL_LIST="first_run"
-while [ -n "$CHANNEL_LIST" ]; do
-  CHANNEL_LIST=$(gcloud alpha monitoring channels list --format="value(name)")
-  if [ -n "$CHANNEL_LIST" ]; then
-    for CHANNEL in $CHANNEL_LIST; do
-      gcloud alpha monitoring channels delete $CHANNEL --quiet
-    done
-  fi
-done
+# clear monitoring
+pip3 install -r ${WORKDIR}/requirements.txt
+python3 ${WORKDIR}/cleanup_monitoring.py
--- a/.github/workflows/e2e_scripts/clean.sh
+++ b/.github/workflows/e2e_scripts/clean.sh
@@ -69,5 +69,5 @@ for LOG in $(gcloud logging logs list --project $PROJECT_ID --format="value(NAME
 done
 
 # clear monitoring
-pip3 install -r ${WORKDIR}/requirements.txt
-python3 ${WORKDIR}/cleanup_monitoring.py
+python3 -m pip install -r ${WORKDIR}/requirements.txt
+python3 ${WORKDIR}/cleanup_monitoring.py $PROJECT_ID
--- a/.github/workflows/e2e_scripts/clean.sh
+++ b/.github/workflows/e2e_scripts/clean.sh
@@ -70,4 +70,4 @@ done
 
 # clear monitoring
 python3 -m pip install -r ${WORKDIR}/requirements.txt
-python3 ${WORKDIR}/cleanup_monitoring.py $PROJECT_ID
+python3 ${WORKDIR}/cleanup_monitoring.py "projects/$PROJECT_ID"
--- a/.github/workflows/e2e_scripts/cleanup_monitoring.py
+++ b/.github/workflows/e2e_scripts/cleanup_monitoring.py
@@ -1,0 +1,109 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+
+from google.cloud import monitoring_v3
+from google.cloud.monitoring_dashboard import v1
+
+project_id = ''
+project_name = 'projects/'
+
+def getProjectId():
+    """Retrieves the project id from the environment variable.
+    Raises:
+    MissingProjectIdError -- When not set.
+    Returns:
+    str -- the project name
+    """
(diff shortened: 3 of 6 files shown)


.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,109 @@
...
+
+project_id = ''
+project_name = 'projects/'
+
+def getProjectId():
+    """Retrieves the project id from the environment variable.
+    Raises:
+    MissingProjectIdError -- When not set.
+    Returns:
+    str -- the project name
```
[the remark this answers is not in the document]
nicklin37: Fixed! :)

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,109 @@
...
+            client.delete_alert_policy(policy.name)
+        except:
+            print('Could not delete alerting policy: ' + policy.name)
+
+def cleanupNotificationChannels():
+    """ Deletes all notification channels. """
+    client = monitoring_v3.NotificationChannelServiceClient()
+    channels = client.list_notification_channels(project_name)
+    for channel in channels:
+        client.delete_notification_channel(channel.name)
```
simonz130, line 62: We should add error handling here.

simonz130:
Great job!

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: Oops good catch! Took it out to do a test and forgot to add it back in. Fixed :)

.github/workflows/e2e.yml
```diff
@@ -20,6 +20,7 @@ on:
       - master
       - release/*
       - e2e-test
+      - monitoring_tests_to_e2e
```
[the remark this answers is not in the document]
nicklin37: Dan told me to add this branch to the list so it would run on this PR but I'm not sure if I did it correctly.. @Daniel-Sanche could you advise here :)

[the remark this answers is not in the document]
Daniel-Sanche: This addition is necessary to see the tests running in this PR, since usually they'd only run after each merge

I'd recommend trying to remove this after all approvals and before merging, but it's not a big deal if it accidentally sticks around

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,112 @@
...
+project_name = 'projects/'
+
+def getProjectId():
+    """Retrieves the project id from the environment variable.
+    Raises:
+        MissingProjectIdError -- When not set.
+    Returns:
+        str -- the project name
+    """
+    project_id = os.environ['GOOGLE_CLOUD_PROJECT']
```
Daniel-Sanche, line 29: This environment variable isn't part of GitHub Actions, so we ideally shouldn't be relying on it here. You should be grabbing the project through the ` ${{ secrets.E2E_PROJECT_ID  }}` injected into the system by GitHub Actions, like the other steps. 

Also, instead of reading from the environment variable, I think it would be more direct to take the project id as an argument and retrieve it with `sys.argv[1]`

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,112 @@
...
+    return project_id
+
+def cleanupDashboards():
+    """ Deletes all dashboards. """
+    client = v1.DashboardsServiceClient()
+    dashboards = client.list_dashboards(project_name)
+    for dashboard in dashboards:
+        try:
+            client.delete_dashboard(dashboard.name)
+        except:
```
Daniel-Sanche, line 44: What happens if the resources don't exist? We want to be able to run cleanup before and after each test, so it shouldn't throw an error if the resources aren't there

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,112 @@
...
+    client = monitoring_v3.UptimeCheckServiceClient()
+    uptime_checks = client.list_uptime_check_configs(project_name)
+    for check in uptime_checks:
+        try:
+            client.delete_uptime_check_config(check.name)
+        except:
+            print('Could not delete uptime check: ' + check.name)
+
+def doCleanup():
+    """ Ensures that resources are deleted in the proper order so no exceptions are thrown. """
```
Daniel-Sanche, line 101: What happens if the resources don't exist? We want to be able to run cleanup before and after each test, so it shouldn't throw an error if the resources aren't there

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,112 @@
...
+        raise MissingProjectIdError(
+            'Set the environment variable ' +
+            'GOOGLE_CLOUD_PROJECT to your Google Cloud Project Id.')
+    return project_id
+
+def cleanupDashboards():
+    """ Deletes all dashboards. """
+    client = v1.DashboardsServiceClient()
+    dashboards = client.list_dashboards(project_name)
+    for dashboard in dashboards:
```
Daniel-Sanche, line 41: You should make sure your script blocks until it sees all resources are actually deleted as expected. We want the main tests to run if and only if the project is verified clean

The other cleanup script runs each step inside a while loop for this reason

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,112 @@
...
+            client.delete_notification_channel(channel.name)
+        except:
+            print('Could not delete notification channel: ' + channel.name)
+
+def cleanupServices():
+    """ Deletes only custom services that are identified by a trailing 'srv' in the name. """
+    client = monitoring_v3.ServiceMonitoringServiceClient()
+    services = client.list_services(project_name)
+    for service in services:
+        if service.name.endswith("srv"): # Only delete the custom services
```
Daniel-Sanche, line 72: nit, but it would be more pythonic to use a list comprehension above to avoid this:

`custom_services = [s for s in client.list_services(project_name) if s.name.endswith("srv")]`

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,112 @@
...
+                client.delete_service(service.name)
+            except:
+                print('Could not delete service: ' + service.name)
+
+def cleanupSlos():
+    """ Deletes every SLO associated with every service. """
+    client = monitoring_v3.ServiceMonitoringServiceClient()
+    services = client.list_services(project_name)
+    for service in services:
+        slos = client.list_service_level_objectives(service.name)
```
Daniel-Sanche, line 83: Again, if you want to get fancy you can remove the loops and make things more compact and readable with a comprehension here. I think this should work, although I haven't tested it:
```
slos = [slo.name for slo in client.list_service_level_objectives(service.name) 
                 for service in client.list_services(project_name)]
```

Feel free to ignore this, just giving some Python inspiration

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: Gotcha - I'm reading it from the argline and passing in the GitHub Actions secret for the testing script and the gcloud result from the clean.sh script.

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: If the resources don't exist then the list of resources to delete are all empty. There isn't an error, the script does nothing and passes.

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: The script now blocks to ensure that all resources are deleted using a while loop. The only condition to break from the while loop is the resources being deleted, let me know if it needs another way to escape (aka # of tries?)

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: image
fixed :)

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: Yeah I modified this a bit - but got it working and I think it is a lot cleaner! thanks!

.github/workflows/e2e.yml
```diff
@@ -61,6 +62,13 @@ jobs:
           test-cloud-shell:$GITHUB_SHA
       env:
         PROJECT_ID: ${{ secrets.E2E_PROJECT_ID  }}
+    - name: Run Monitoring Integration Tests
+      run: |
+        set -x
+        # install dependencies
+        python3 -m pip install -r tests/requirements.txt
+        # run tests
+        python3 tests/monitoring_integration_test.py ${{ secrets.E2E_PROJECT_ID }}
```
Daniel-Sanche, line 71: You actually have to explicitly attach the secret to the step. See the step above yours:

```
      env:
        PROJECT_ID: ${{ secrets.E2E_PROJECT_ID  }}
```

You can then use PROJECT_ID as an environment variable within the step

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,121 @@
...
+project_name = 'projects/'
+
+def getProjectId():
+    """Retrieves the project id from the script arguments.
+    Returns:
+        str -- the project name
+    Exits when project id is not set
+    """
+    try:
+        project_id = sys.argv[1]
```
Daniel-Sanche, line 30: personally I'd rather just do this in __main__ since it's so simple, but this works too

.github/workflows/e2e.yml
Daniel-Sanche: Also, you don't seem to need the id in your script, only the name. if you send it in as projects/$PROJECT_ID, you can avoid storing 2 variables

.github/workflows/e2e_scripts/clean.sh
```diff
@@ -67,35 +68,6 @@ for LOG in $(gcloud logging logs list --project $PROJECT_ID --format="value(NAME
...
-  CHANNEL_LIST=$(gcloud alpha monitoring channels list --format="value(name)")
-  if [ -n "$CHANNEL_LIST" ]; then
-    for CHANNEL in $CHANNEL_LIST; do
-      gcloud alpha monitoring channels delete $CHANNEL --quiet
-    done
-  fi
-done
+# clear monitoring
+python3 -m pip install -r ${WORKDIR}/requirements.txt
+python3 ${WORKDIR}/cleanup_monitoring.py $PROJECT_ID
```
Daniel-Sanche, line 73: It looks like you only need the project name in the script, not the id. You can prefix this with "projects/", and only store the one variable

.github/workflows/e2e_scripts/cleanup_monitoring.py
Daniel-Sanche: perfect!

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,121 @@
...
+            try:
+                client.delete_service_level_objective(slo.name)
+            except:
+                print('Could not delete SLO: ' + slo.name)
+        slos = [slo for service in client.list_services(project_name) for slo in client.list_service_level_objectives(service.name)]
+
+def cleanupUptimeCheck():
+    """ Deletes every uptime check. """
+    client = monitoring_v3.UptimeCheckServiceClient()
+    uptime_checks = list(client.list_uptime_check_configs(project_name))
```
Daniel-Sanche, line 100: nit: the better Python practice would probably be to name `project_name` an argument for each function instead of a global variable. But I don't think that's too important for our purposes

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,121 @@
...
+    """ Deletes every SLO associated with every service. """
+    client = monitoring_v3.ServiceMonitoringServiceClient()
+    slos = [slo for service in client.list_services(project_name) for slo in client.list_service_level_objectives(service.name)]
+    while (len(slos) != 0):
+        for slo in slos:
+            try:
+                client.delete_service_level_objective(slo.name)
+            except:
+                print('Could not delete SLO: ' + slo.name)
+        slos = [slo for service in client.list_services(project_name) for slo in client.list_service_level_objectives(service.name)]
```
Daniel-Sanche, line 95: It isn't a great idea to do complex comprehensions like this in two places, because we may later change one without remembering to change the other. The best way to fix this is with the [walrus operator](https://www.geeksforgeeks.org/walrus-operator-in-python-3-8/), but that's only available in 3.8+ (the VM seems to use 3.5)

Instead, you may want to do something like this:
```
complete = False
while not complete:
  slos = [slo for service in client.list_services(project_name) for slo in client.list_service_level_objectives(service.name)]
  if not slos:
    complete = True
  else:
    for slo in slos: ...
```

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,121 @@
...
+                client.delete_service(service.name)
+            except:
+                print('Could not delete service: ' + service.name)
+        custom_services = [s for s in client.list_services(project_name) if s.name.endswith("srv")]
+
+def cleanupSlos():
+    """ Deletes every SLO associated with every service. """
+    client = monitoring_v3.ServiceMonitoringServiceClient()
+    slos = [slo for service in client.list_services(project_name) for slo in client.list_service_level_objectives(service.name)]
+    while (len(slos) != 0):
```
Daniel-Sanche, line 89: FYI, a list evaluates to true if it has elements, and false if it's empty. This way is [considered more Pythonic](https://www.techiedelight.com/check-list-empty-python/): `while slos:`.

But it may be slightly less readable to non-Python people. Don't feel like you need to change these

Daniel-Sanche: you should also avoid using parentheses for single-expression statements in Python

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -0,0 +1,121 @@
...
+            try:
+                client.delete_service(service.name)
+            except:
+                print('Could not delete service: ' + service.name)
+        custom_services = [s for s in client.list_services(project_name) if s.name.endswith("srv")]
+
+def cleanupSlos():
+    """ Deletes every SLO associated with every service. """
+    client = monitoring_v3.ServiceMonitoringServiceClient()
+    slos = [slo for service in client.list_services(project_name) for slo in client.list_service_level_objectives(service.name)]
```
Daniel-Sanche, line 88: I'd recommend breaking this into two lines to line up the `for`s

.github/workflows/e2e.yml
nicklin37: Fixed! thanks for pointing this out :)

.github/workflows/e2e_scripts/clean.sh
nicklin37: True - I make this modification!

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: It was a small change so I made it!

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: I preferred the second version since it seemed a bit more readable - made the change for all the functions for consistency.

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: I made the change - I don't think a lot of people will be modifying the clean script. Removed extra parentheses :)

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: Good point - moved it to the main for simplicity

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: Fixed! :)

nicklin37:
> I made one important comment about the secret environment variable, and a some more Python style advice. Feel free to ignore those

Thanks Dan! I've made all the changes I think and resolved most of the issues - feel free to let me know if I missed anything.

PR #336, opened 2020-08-15
Add OpenTelemetry Tracing for Payment service

### LINKED ISSUE #323: Add Tracing for Payment Service
using OpenTelemetry :) 
subissue of #132

gzhao408:
Resolves #323 and subtask of #132 

Payment is only called by Checkout Service. It calls `Charge()`, i.e. to "charge the credit card." You should see both the client and server side calls to Charge in Trace console:
image

Also removes @google-cloud/trace-agent because it's not needed anymore :) and it causes the PROJECT_ID NOT FOUND errors.

### DIFF
--- a/src/paymentservice/tracer.js
+++ b/src/paymentservice/tracer.js
@@ -1,0 +1,38 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const opentelemetry = require('@opentelemetry/api');
+const {NodeTracerProvider} = require('@opentelemetry/node');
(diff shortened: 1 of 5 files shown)


src/paymentservice/tracer.js
```diff
@@ -0,0 +1,38 @@
...
+        // Use grpc plugin to receive trace contexts from client
+        plugins: {
+            grpc: {
+                enabled: true,
+                path: '@opentelemetry/plugin-grpc',
+            }
+        }
+    });
+    // Cloud Trace Exporter handles credentials.
+    provider.addSpanProcessor(new BatchSpanProcessor(new TraceExporter()));
```
simonz130, line 32: How the auth works here?
Will this work when running locally w/o auth environment variable being set?
If not, please add a comment here to explain how auth is handled implicitly here.

gzhao408: According to [this GCP doc](https://cloud.google.com/trace/docs/setup/nodejs-ot):
image
It uses default credentials, which I think depends on the situation (could be `GOOGLE_APPLICATION_CREDENTIALS` when local, or some other implementation in GCP that I don't know anything about. Since I'm also not directly using any environment variables in the code I wouldn't want to assume that there's an issue if, for example, `PROJECT_ID` isn't set.
This was the case for the go services - it warns if the environment variable isn't set, but it's always been able to authenticate even when the warning is printed.

simonz130: You can copy the same remark to the code where you instantiate the TraceExporter so it doesn't feel weird for folks who just copy-paste your code as a reference code when they learn how to use TraceExporter in node for tracing

PR #338, opened 2020-08-17
Documentation for OpenTelemetry Instrumentation

gzhao408:
Adds thorough and consistent documentation about the OpenTelemetry tracing instrumentation to each service. A few key points:
- How authentication is handled in the cloud trace exporters.
- Warnings for PROJECT_ID when the environment variable isn't found, but also noting that this is not needed when running on GCP.
- Reminders when certain implementations would not be appropriate for production environments (i.e. AlwaysSample)
- Distinguishing between "client" and "server" responsibilities within the services
- And other miscellaneous things!

Also, there was an OpenCensus tracer leftover in frontend - I removed it and tested to make sure it doesn't affect the dashboards.

Part of #132 - we're close to the end!

### DIFF
--- a/src/checkoutservice/main.go
+++ b/src/checkoutservice/main.go
@@ -121,6 +121,7 @@ func main() {
 	}
 	srv := grpc.NewServer(
 		grpc.StatsHandler(&ocgrpc.ServerHandler{}), // TODO: replace with OTel grpc metrics collector
+		// OpenTelemetry gRPC server channel interceptors receive trace contexts from clients.
 		grpc.UnaryInterceptor(grpctrace.UnaryServerInterceptor(global.TraceProvider().Tracer("checkout"))),
 		grpc.StreamInterceptor(grpctrace.StreamServerInterceptor(global.TraceProvider().Tracer("checkout"))),
 	)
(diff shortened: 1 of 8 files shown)


gzhao408:
Woops, race condition with the #312 and #336 merges.
Rebasing on master...

src/checkoutservice/main.go
```diff
@@ -158,6 +159,12 @@ func initOpenCensusStats() {
 
 // Initialize OTel trace provider that exports to Cloud Trace
 func initTraceProvider() {
+	// When running on GCP, authentication is handled automatically
+	// using default credentials. This environment variable check
+	// is to help debug projects running locally. It's possible for this
+	// this warning to be printed while the exporter works normally. See
```
simonz130, line 165: nit: you say 'this' twice

src/frontend/main.go
```diff
@@ -240,15 +232,28 @@ func initTelemetry(log logrus.FieldLogger) {
 	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
 	initOpenCensus(log)
 
-	// Initialize exporter OTel Trace -> GCP Trace
+	// When running on GCP, authentication is handled automatically
+	// using default credentials. This environment variable check
+	// is to help debug projects running locally. It's possible for this
+	// this warning to be printed while the exporter works normally. See
```
simonz130, line 238: nit: "this this"

PR #361, opened 2020-08-21
reorder e2e cleanup to clean monitoring first

nicklin37:
This will cleanup the monitoring before the cluster is destroyed and the Istio services are lost.

### DIFF
--- a/.github/workflows/e2e_scripts/cleanup_monitoring.py
+++ b/.github/workflows/e2e_scripts/cleanup_monitoring.py
@@ -35,13 +35,12 @@ def cleanupLogBasedMetrics(project_name):
     client = logging_v2.MetricsServiceV2Client()
     metrics = True
     while metrics:
-        metrics = list(client.list_log_metrics(project_name))
+        metrics = ["{}/metrics/{}".format(project_name, metric.name) for metric in client.list_log_metrics(project_name)]
         for metric in metrics:
-            metric_name = project_name + "/metrics/" + metric.name
             try:
-                client.delete_log_metric(metric_name)
+                client.delete_log_metric(metric)
             except:
-                print('Could not delete metric: ' + metric_name)
+                print('Could not delete metric: ' + metric)
 
 def cleanupPolicies(project_name):
     """ Delete all alerting policies for both uptime checks and SLOs. """
--- a/.github/workflows/e2e_scripts/cleanup_monitoring.py
+++ b/.github/workflows/e2e_scripts/cleanup_monitoring.py
@@ -37,10 +37,11 @@ def cleanupLogBasedMetrics(project_name):
     while metrics:
         metrics = list(client.list_log_metrics(project_name))
         for metric in metrics:
+            metric_name = project_name + "/metrics/" + metric.name
             try:
-                client.delete_log_metric(metric.name)
+                client.delete_log_metric(metric_name)
             except:
-                print('Could not delete metric: ' + metric.name)
+                print('Could not delete metric: ' + metric_name)
 
 def cleanupPolicies(project_name):
     """ Delete all alerting policies for both uptime checks and SLOs. """
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -227,10 +227,10 @@ installMonitoring() {
     log "${YELLOW}⚠️ Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
     log ""
     read -p "${YELLOW}When you are done, please PRESS ENTER TO CONTINUE"
-  else
-    sleep 30 # sleep to allow for services to be detected by Cloud Monitoring
   fi
 
+  log "Checking to make sure necessary Istio services are ready for monitoring"
+  python3 monitoring/istio_service_setup.py $project_id $CLUSTER_ZONE
   log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."
   pushd monitoring/
   terraform init -lock=false
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -227,6 +227,8 @@ installMonitoring() {
     log "${YELLOW}⚠️ Please create a monitoring workspace for the project by clicking on the following link: $gcp_monitoring_path"
     log ""
     read -p "${YELLOW}When you are done, please PRESS ENTER TO CONTINUE"
+  else
+    sleep 20 # sleep to allow for services to be detected by Cloud Monitoring
   fi
 
   log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -346,6 +346,7 @@ parseArguments() {
       log "-p|--project|--project-id     GCP project to deploy Cloud Operations Sandbox to"
       log "-v|--verbose                  print commands as they run (set -x)"
       log "--skip-workspace-prompt       Don't pause for Cloud Monitoring workspace set up"
+      log "--service-wait                Wait indefinitely for services to be detected by Cloud Monitoring"
       log ""
       exit 0
       ;;
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -228,7 +228,7 @@ installMonitoring() {
     log ""
     read -p "${YELLOW}When you are done, please PRESS ENTER TO CONTINUE"
   else
-    sleep 20 # sleep to allow for services to be detected by Cloud Monitoring
+    sleep 30 # sleep to allow for services to be detected by Cloud Monitoring
   fi
 
   log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -230,6 +230,7 @@ installMonitoring() {
   fi
 
   log "Checking to make sure necessary Istio services are ready for monitoring"
+  python3 -m pip install google-cloud-monitoring
   python3 monitoring/istio_service_setup.py $project_id $CLUSTER_ZONE
   log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."
   pushd monitoring/
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -231,7 +231,7 @@ installMonitoring() {
 
   log "Checking to make sure necessary Istio services are ready for monitoring"
   python3 -m pip install google-cloud-monitoring
-  python3 monitoring/istio_service_setup.py $project_id $CLUSTER_ZONE
+  python3 monitoring/istio_service_setup.py $project_id $CLUSTER_ZONE $service-wait
   log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."
   pushd monitoring/
   terraform init -lock=false
@@ -329,6 +329,10 @@ parseArguments() {
       ;;
     --skip-workspace-prompt)
       skip_workspace_prompt=1
+      shift
+      ;;
+    --service-wait)
+      service_wait=1
       shift
       ;;
     -v|--verbose)
--- a/terraform/monitoring/istio_service_setup.py
+++ b/terraform/monitoring/istio_service_setup.py
@@ -38,7 +38,7 @@ def findService(client, service_name, project_id, zone, timeout):
 	while not found_service and num_tries <= 20:
 		try:
 			found_service = client.get_service(service)
-		except:
+		except: # possible exceptions include GoogleAPICallError and ValueError
 			if timeout:
 				num_tries += 1
 			time.sleep(3)
--- a/terraform/monitoring/istio_service_setup.py
+++ b/terraform/monitoring/istio_service_setup.py
@@ -18,7 +18,7 @@ import sys
 from google.cloud import monitoring_v3
 
 def getIstioServiceName(service_name, project_id, zone):
-	return 'ist:' + project_id + '-zone-' + zone + '-stackdriver-sandbox-default-' + service_name
+	return 'ist:' + project_id + '-zone-' + zone + '-cloud-ops-sandbox-default-' + service_name
 
 def findService(client, service_name, project_id, zone):
 	found_service = False
--- a/terraform/monitoring/istio_service_setup.py
+++ b/terraform/monitoring/istio_service_setup.py
@@ -59,7 +59,7 @@ def waitForIstioServicesDetection(project_id, zone, timeout):
 	client = monitoring_v3.ServiceMonitoringServiceClient()
 
 	# wait for each Istio service to be detected by Cloud Monitoring
-	for service in ["cartservice", "productcatalogservice", "currencyservice", "recommendationservice", "adservice"]:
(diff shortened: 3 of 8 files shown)


nicklin37:
Waiting with the Service API for the services we need to become available solved the issue. The test has passed 4x in a row now. This commit also solves some terraform warnings about interpolations. It's ready for review/merge now.

nicklin37:
Closes #374

.github/workflows/e2e_scripts/cleanup_monitoring.py
```diff
@@ -37,10 +37,11 @@ def cleanupLogBasedMetrics(project_name):
     while metrics:
         metrics = list(client.list_log_metrics(project_name))
         for metric in metrics:
+            metric_name = project_name + "/metrics/" + metric.name
```
Daniel-Sanche, line 40: This seems cleaner to me: 

```
metrics = ["{}/metrics/{}".format(project_name, metric) for metric in client.list_log_metrics(project_name)]
```

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+
+from google.cloud import monitoring_v3
+
+def getIstioServiceName(service_name, project_id, zone):
+	return 'ist:' + project_id + '-zone-' + zone + '-cloud-ops-sandbox-default-' + service_name
```
Daniel-Sanche, line 21: This kind of string concatenation is discouraged. I'd recommend using `.format()`

`'ist:{}zone-{}-cloud-ops-sandbox-default-{}'.format(project_id, zone, service_name)

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+if __name__ == '__main__':
+    project_id = ''
+    zone = ''
+    try:
+        project_id = sys.argv[1]
+        zone = sys.argv[2]
+    except:
+        exit('Missing Project Name. Usage: python3 istio_service_setup.py $project_id $zone')
```
Daniel-Sanche, line 59: This error message seems misleading. What if it's the zone that is missing?

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+	findService(client, "adservice", project_id, zone)
+
+
+if __name__ == '__main__':
+    project_id = ''
+    zone = ''
+    try:
+        project_id = sys.argv[1]
+        zone = sys.argv[2]
+    except:
```
Daniel-Sanche, line 58: Probably a nit, but it's a good idea to only catch the exception you expect. That doesn't matter too much for argument parsing code though

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+	return 'ist:' + project_id + '-zone-' + zone + '-cloud-ops-sandbox-default-' + service_name
+
+def findService(client, service_name, project_id, zone):
+	found_service = False
+	full_service_name = getIstioServiceName(service_name, project_id, zone)
+	service = client.service_path(project_id, full_service_name)
+	while not found_service:
+		try:
+			found_service = client.get_service(service)
+		except:
```
Daniel-Sanche, line 30: Should you be catching a specific exception here?

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+from google.cloud import monitoring_v3
+
+def getIstioServiceName(service_name, project_id, zone):
+	return 'ist:' + project_id + '-zone-' + zone + '-cloud-ops-sandbox-default-' + service_name
+
+def findService(client, service_name, project_id, zone):
+	found_service = False
+	full_service_name = getIstioServiceName(service_name, project_id, zone)
+	service = client.service_path(project_id, full_service_name)
+	while not found_service:
```
Daniel-Sanche, line 27: We should put some kind of optional retry loop here, and make sure a friendly error is displayed after it fails (probably outside of this script)

It was ok to loop forever in the other test scripts, because eventually GitHub Actions will timeout/we'll catch the error. If this script is to be run as part of install.sh, we want to make sure it's more friendly

terraform/install.sh
```diff
@@ -229,6 +229,9 @@ installMonitoring() {
     read -p "${YELLOW}When you are done, please PRESS ENTER TO CONTINUE"
   fi
 
+  log "Checking to make sure necessary Istio services are ready for monitoring"
+  python3 -m pip install google-cloud-monitoring
```
Daniel-Sanche, line 233: This should probably be installed as part of our base image (@eaberman)

eaberman: Good point! I'll open up an issue for this and address it.

terraform/monitoring/istio_service_setup.py
nicklin37: I guess there are 3 different exceptions we want to catch but if there is any kind of error or exception we want to retry the API call anyways - do you think I should specify the exceptions?

terraform/monitoring/istio_service_setup.py
simonz130: +1 on format

simonz130: Also, perhaps you can extract the name of the product into a const and use it throughout

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+
+import os
+import sys
+
+from google.cloud import monitoring_v3
+
+def getIstioServiceName(service_name, project_id, zone):
+	return 'ist:' + project_id + '-zone-' + zone + '-cloud-ops-sandbox-default-' + service_name
+
+def findService(client, service_name, project_id, zone):
```
simonz130, line 23: document the method

terraform/monitoring/istio_service_setup.py
simonz130: I think it's fine to be catching generic exception here - given the fact it might run in provisioning. But maybe worth adding a comment saying which exceptions can happen here.

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+	found_service = False
+	full_service_name = getIstioServiceName(service_name, project_id, zone)
+	service = client.service_path(project_id, full_service_name)
+	while not found_service:
+		try:
+			found_service = client.get_service(service)
+		except:
+			found_service = False
+	print("Found " + service_name)
```
simonz130, line 32: This message lacks context (found where, why were you trying to find it, etc)

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+
+def waitForIstioServices(project_id, zone):
```
simonz130, line 34: document this and also the name should reflect what are you waiting for.
Example: waitForIstioServicesDetection

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+	client = monitoring_v3.ServiceMonitoringServiceClient()
+	# wait for cart service
+	findService(client, "cartservice", project_id, zone)
```
simonz130, line 37: I see only 5 services here. Should we add all of them?

nicklin37: Yeah this is true - these are the only 5 services we need for the Terraform monitoring, so it's okay if the others haven't shown up since monitoring provisioning won't fail. Do you think we should try to detect all 10?

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,60 @@
...
+	findService(client, "productcatalogservice", project_id, zone)
+
+	# wait for currency service
+	findService(client, "currencyservice", project_id, zone)
+
+	# wait for recommendation service
+	findService(client, "recommendationservice", project_id, zone)
+
+	# wait for ad service
+	findService(client, "adservice", project_id, zone)
```
Daniel-Sanche, line 49: This would be a good use of a loop:

```
for service in ['adservice', 'currencyservice'...
```

terraform/monitoring/istio_service_setup.py
Daniel-Sanche: Yeah a comment or something might be helpful

.github/workflows/e2e_scripts/cleanup_monitoring.py
nicklin37: Fixed!

terraform/monitoring/istio_service_setup.py
nicklin37: Fixed!

terraform/monitoring/istio_service_setup.py
nicklin37: Woops! Adjusted the Error message!

terraform/monitoring/istio_service_setup.py
nicklin37: Switched it to get IndexError which is what we'd expect for missing arguments.

terraform/monitoring/istio_service_setup.py
nicklin37: Added a comment!

terraform/monitoring/istio_service_setup.py
nicklin37: Fixed!

terraform/monitoring/istio_service_setup.py
nicklin37: Fixed!

terraform/monitoring/istio_service_setup.py
nicklin37: Fixed!

terraform/monitoring/istio_service_setup.py
nicklin37: Added context!

terraform/monitoring/istio_service_setup.py
nicklin37: I've added a parameter called service_wait where when set means we'll wait forever for the services. I'm a little iffy about the naming of the variable so let me know if it needs to be adjusted. I have it default set to false.

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,79 @@
...
+	Arguments:
+	project_id - the Sandbox project id (cloud-ops-sandbox-###)
+	zone - the zone of the Sandbox cluster
+	timeout - whether to timeout after 1 minute or wait indefinitely for the service
+	"""
+	client = monitoring_v3.ServiceMonitoringServiceClient()
+
+	# wait for each Istio service to be detected by Cloud Monitoring
+	for service in ["cartservice", "productcatalogservice", "currencyservice", "recommendationservice", "adservice"]:
+		findService(client, service, project_id, zone, timeout)
```
Daniel-Sanche, line 63: It's a bit strange that the tests all run in sequence. If it takes 3 minues for them to spin up, the first 2 will fail and print messages, even though they eventually spin up

You might want to try this to run them in parallel:
```
import concurrent.futures

services = ["cartservice", "productcatalogservice", "currencyservice", "recommendationservice", "adservice"]
with concurrent.futures.ProcessPoolExecutor(len(services)) as executor:
    for service in services:
	executor.submit(findService, client, service, project_id, zone, timeout):
```

But this may be a bit extreme. If it doesn't work on the first try, I'm fine with leaving as is

Daniel-Sanche: (If you do this, you should change the timeout to 5 mins each)

simonz130: +1 on running findService in parallel

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,79 @@
...
+	return "ist:{}-zone-{}-cloud-ops-sandbox-default-{}".format(project_id, zone, service_name)
+
+def findService(client, service_name, project_id, zone, timeout):
+	""" Checks to see if a service exists in Cloud Monitoring 
+	Arguments:
+	client - the API client
+	service_name - the Istio service name, returned from getIstioServiceName
+	project_id - the Sandbox project id
+	zone - the zone of the Sandbox cluster
+	timeout - whether to timeout after 1 minute or wait indefinitely for the service
```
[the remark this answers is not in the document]
simonz130: Timeout sounds like a time span, and the comment assumes that it's 1 min. timeout. It's confusing.
Timeout should be a timespan object and the method would be called passing a  time interval defined as const.

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,79 @@
...
+    project_id = ''
+    zone = ''
+    try:
+        project_id = sys.argv[1]
+        zone = sys.argv[2]
+    except IndexError:
+        exit('Missing Project Name or Zone. Usage: python3 istio_service_setup.py $project_id $zone')	
+
+    # optional timeout parameter, default = True
+    if len(sys.argv) == 4:
```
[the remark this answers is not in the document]
nicklin37: Yeah right now i have it set as a flag 1 or 0 but maybe I'll switch it to use getopts to support cleaner arg parsing for next sprint.

terraform/monitoring/istio_service_setup.py
```diff
@@ -0,0 +1,79 @@
...
+	service_name - the Istio service name, returned from getIstioServiceName
+	project_id - the Sandbox project id
+	zone - the zone of the Sandbox cluster
+	timeout - whether to timeout after 1 minute or wait indefinitely for the service
+	"""
+	found_service = False
+	full_service_name = getIstioServiceName(service_name, project_id, zone)
+	service = client.service_path(project_id, full_service_name)
+	num_tries = 0
+	while not found_service and num_tries <= 20:
```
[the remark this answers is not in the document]
nicklin37: Good idea - fixed!

terraform/monitoring/istio_service_setup.py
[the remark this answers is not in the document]
nicklin37: Sorry I think I'm confused now haha - I renamed it to should_timeout which is a boolean flag. Should I change the type of the variable to timespan? or should I change the name and the type?

terraform/install.sh
```diff
@@ -328,6 +331,10 @@ parseArguments() {
       skip_workspace_prompt=1
       shift
       ;;
+    --service-wait)
+      service_wait=1
+      shift
+      ;;
```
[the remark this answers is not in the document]
nicklin37: Fixed!

terraform/monitoring/istio_service_setup.py
nicklin37: fixed! Timeout is increased to 5 minutes per service (in parallel so ~5 min total for script)

terraform/monitoring/istio_service_setup.py
nicklin37: All 10 are being detected now.

Issue #11, opened 2019-04-06
Running install.sh multiple times


Daniel-Sanche:
How do we want the script to behave when it is run multiple times? Currently, terraform will skip deploying a new project, but the load balancer portion will create a new load balancer.

Do we want the script to cancel if it sees tfstate files in the current directory? Or should it always deploy another sandbox instance?  Or something else?

simonz130:
Good point. We should make it idempotent:

For each step (such as creating a loadgen, deploying manfiests, etc) - we should make sure that the result of running the script once with that step or multiple times with that step yield the same end result. 
Compare desired state to deployed state and do the delta. Terraform already has that baked in.
We need to update the pieces of install.sh: it should not create additional load balancers but rather remove existing and create a new one instead, etc.

It can wait until after Next though.

jiang95-dev:
@simonz130 This should be delayed since load generator will be moved to Cloud Run.

PR #387, opened 2020-08-25
remove existing load generator when using old projects

### LINKED ISSUE #32: Install script should display cluster removal info in next steps

### LINKED ISSUE #378: Loadgenerator shows incorrect IP address sometimes
When provisioning a Sandbox from an existing Sandbox, sometimes the loadgenerator will show two different IP addresses and one of them (the first one) will be incorrect. 
Screenshot 2020-08-24 at 4 16 40 PM - Display 2

jiang95-dev:
1. Remove existing load generator when using old projects.
Fixes #378, #11.
2. Display cluster removal info in next step.
Fixes #32.

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -298,6 +298,11 @@ displaySuccessMessage() {
     log "     Google Cloud Console Monitoring Workspace: $gcp_monitoring_path"
     log "     Hipstershop web app address: http://$external_ip"
     log "     Load generator web interface: $loadgen_addr"
+    log ""
+    log "To remove the Sandbox once finished using it, run"
+    log ""
+    log "     ./destroy.sh"
+    log ""
     log "********************************************************************************"
 }
 
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -257,7 +257,7 @@ loadGen() {
   old_loadgen=$(gcloud compute instances list --project $project_id --filter="name:loadgenerator*" --format="value(name)")
   if [[ -n "$old_loadgen" ]]; then
     loadgen_zone=$(gcloud compute instances list --project $project_id --filter="name:loadgenerator*" --format="value(zone)")
-    $(gcloud compute instances delete $old_loadgen --zone $loadgen_zone --quiet)
+    gcloud compute instances delete $old_loadgen --zone $loadgen_zone --quiet
   fi
   # launch a new load generator
   ../loadgenerator/loadgen autostart $external_ip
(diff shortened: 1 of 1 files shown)


terraform/install.sh
```diff
@@ -253,6 +253,13 @@ getExternalIp() {
 # Install Load Generator service and start generating synthetic traffic to Sandbox
 loadGen() {
   log "Running load generator"
+  # remove existing load generator
+  old_loadgen=$(gcloud compute instances list --project $project_id --filter="name:loadgenerator*" --format="value(name)")
+  if [[ -n "$old_loadgen" ]]; then
+    loadgen_zone=$(gcloud compute instances list --project $project_id --filter="name:loadgenerator*" --format="value(zone)")
+    $(gcloud compute instances delete $old_loadgen --zone $loadgen_zone --quiet)
```
Daniel-Sanche, line 260: This seems like a bit of a hack. Instead of deleting the old one, can't we just only create a new one if `old_loadgen` doesn't exist?

```
loadgen_ip=$(gcloud compute instances list --project "$project_id" \
                                           --filter="name:loadgenerator*" \
                                           --format="value(networkInterfaces[0].accessConfigs.natIP)")
if [[ -z "$loadgen_ip" ]]; then
    # load generator not found. Launch a new one
  ../loadgenerator/loadgen autostart $external_ip
fi

while [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip:8080" -o /dev/null --max-time 1) -ne 200  && \
...
```

Daniel-Sanche: Or is the issue that it might be pointing to the wrong external IP?

Daniel-Sanche, line 260: You should be able to remove the `$()` around this

jiang95-dev: @simonz130 This was pointed out by you in the issue. I'm also a bit confused. Can you explain why? Thanks!

simonz130: When reusing the Sandbox which might already have the loadgen provisioned, terraform will re-provision loadgen so we'll end up having two and both IPs will be reported upon completion of provisioning script.
What we need is a single loadgen. 

So two options:

1) Remove the "old" loadgen and provision a new one
2) Keep the "old" one and don't provision the new one

If we go with (2), it might be running against demo cluster or it might be that demo cluster now has a new IP so "old" loadgen needs to be restarted against the new IP of the FE. As long as we ensure that the end result is that loadgen is running against the correct IP of the FE - I'm fine.

If we go with (1), it's simpler - we remove the loadgen - no matter what it does currently and start a new one against the right IP.

For simplicity I prefer (1) but I'm also OK if we do (2). We are planning to replace current loadgen with Cloud Run, so we'll need to revisit this logic anyway.

Issue #394, opened 2020-08-26
Cloud Shell image GitHub Actions uses cahced image


Daniel-Sanche:
We currently have a GA pipeline set up to re-build the cloud shell image [on each push ](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/a52b6964e7430939bd80065fd331f97e3f850ab7/.github/workflows/push-master.yml#L39) and each [release tag](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/a52b6964e7430939bd80065fd331f97e3f850ab7/.github/workflows/push-master.yml#L39). The problem is this push uses the cahced image layers instead of re-pulling, so it will get out of date with the Cloud Shell team's base image. 

We could do a `--no-cache` to force a full re-build, but that would slow down the tests considerably. Instead, I think we should change our [existing periodic rebuild action](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/master/.github/workflows/update-custom-image.yml) to kick off on each commit to master, and each new tag pushed

PR #402, opened 2020-08-27
Fix Cloud Shell GitHub Actions

Daniel-Sanche:
Addresses https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/394

I realized the other day that our GitHub Actions are writing Cloud Shell images using cached base images, which effectively expire every week. I updated the Actions to use Elena's new Cloud Build trigger instead, which is more robust. I also added some improvements to the GitHub Action that kicks off the Cloud Build. Now, we can see the Cloud Build logs within the GitHub UI, and GitHub will mirror the Cloud Build completion status so we'll know if anything fails

Here's an example run:
https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/actions/runs/227498277

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -8,12 +8,12 @@ workloads run using [GitHub self-hosted runners](https://help.github.com/en/acti
     - VM should be at least n1-standard-4 with 100GB persistent disk
     - VM should be created with appropriate service account for desired [worker tag](#Tags)
 2. SSH into new VM through Google Cloud Console
-3. Follow the instructions to add a new runner on the [Actions Settings page](https://github.com/GoogleCloudPlatform/stackdriver-sandbox/settings/actions) to authenticate the new runner
+3. Follow the instructions to add a new runner on the [Actions Settings page](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/settings/actions) to authenticate the new runner
 4. Attach the [appropriate tag(s)](#Tags) to the new runner
 5. Set GitHub Actions as a background service
     - `sudo ~/actions-runner/svc.sh install ; sudo ~/actions-runner/svc.sh start`
 6. Run the following command to install dependencies
-    - `wget -O - https://raw.githubusercontent.com/GoogleCloudPlatform/stackdriver-sandbox/master/.github/workflows/install-dependencies.sh | bash`
+    - `wget -O - https://raw.githubusercontent.com/GoogleCloudPlatform/cloud-ops-sandbox/master/.github/workflows/install-dependencies.sh | bash`
 
 ## Tags
 - `kind-cluster`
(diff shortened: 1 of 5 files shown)


.github/workflows/README.md
```diff
@@ -22,7 +22,12 @@ workloads run using [GitHub self-hosted runners](https://help.github.com/en/acti
...
 - `push-privilege`
   - image push worker
-  - requires permissions to push images to GCR storage bucket, and deploy to App Engine
+  - requires the following permissions on the CI test project:
+    - `App Engine Admin`
+    - `Cloud Build Service Account`
+    - `Cloud Build Editor`
+    - `read permissions on GCS bucket holding cloud build logs`
+    - `write permissons on GCS bucket holding GCR images`
   - requires `PROJECT_ID` to be set properly in the [repo's secrets](https://github.com/GoogleCloudPlatform/stackdriver-sandbox/settings/secrets)
```
simonz130, line 31: cloud-ops-sandbox in the repo link

cloud-shell/cloudbuild.yaml
```diff
@@ -38,3 +38,4 @@ steps:
             echo "No cloud-shell directory in tag $$TAG"
         fi
     done
+logsBucket: 'gs://sandbox-cloud-build-logs'
```
simonz130, line 41: Where is this used and how was this bucket created?

Daniel-Sanche: Cloud Build stores its logs in a bucket. We need to be able to read from the bucket to stream the logs from our GitHub Actions worker. It looks like the permission system is much simpler if we use a custom bucket, rather than a Google-managed one

I just created a standard empty bucket in the UI

.github/workflows/update-custom-image.yml
```diff
@@ -12,23 +12,42 @@
...
+    - name: Display Logs
+      continue-on-error: true
+      run: |
+        gcloud beta builds log --stream ${{ steps.cloud_build.outputs.run_id }}
+    - name: Check Completion Status
+      run: |
+        set -x
+        STATUS="QUEUED"
+        while [[ "$STATUS" == "WORKING" || "$STATUS" == "QUEUED" || "$STATUS" == "STATUS_UNKNOWN" ]]; do
+            sleep 5
```
eaberman, line 49: Is there any chance that this could become stuck? e.g. if STATUS= STATUS_UNKNOWN permanently?

Daniel-Sanche: It could, but the GitHub Action has a timeout that will stop it if it takes too long. I think relying on that is better than spinning up our own timeout functionality

eaberman: Okay, sounds good!

Issue #404, opened 2020-08-27
loadgen image does not update on push to master and new releases


gzhao408:
The image we use to deploy is gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator:latest
which is from March 2019. We need to be able to update it, and automatically!

PR #405, opened 2020-08-27
[loadgen] update docker image, auto-push to gcr, and add script for automatic load generator on Cloud Run

### LINKED ISSUE #403: loadgen: uses outdated locust libraries
our Load Generator is overdue for an update :)

gzhao408:
Resolves #403 and #404 which are, funnily, http response codes I got a million times while trying to implement this.

This PR:
- [x] Updates our load generator's base image and the Locust libraries we use.
- [x] Adds a command to the loadgen script to `autorun`, which spins up a load generator on Cloud Run and outputs the URL at which you can access it.
- [x] Auto-configures fully-managed Cloud Run, `us-central1` region, and allowing unauthenticated requests for the load generator.
- [x] Adds CI workflows to update our stackdriver-sandbox loadgenerator image on every push to master and new release.

This does not:
- [ ] Remove any of the load generation stuff from before.
- [ ] Update the install script. On start, it will still use the old loadgenerator on GCE. But you now have the option to use the `autorun` method as well. Obviously, I'm planning on fully replacing it, but haven't done it in this PR.


src/loadgenerator/loadgenerator-entrypoint.sh
```diff
@@ -24,4 +24,13 @@ if [[ -z "${FRONTEND_ADDR}" ]]; then
...
+
+# if one request to the frontend fails, then exit
+STATUSCODE=$(curl --silent --output /dev/stderr --write-out "%{http_code}" http://${FRONTEND_ADDR})
+if test $STATUSCODE -ne 200; then
+    echo "Error: Could not reach frontend - Status code: ${STATUSCODE}"
+    exit 1
+fi
+
+# run locust
+locust --host="http://${FRONTEND_ADDR}" --web-port=$PORT
```
Daniel-Sanche, line 36: I don't see a default port being set anywhere. It will be auto-injected by Cloud Run, but it's a good idea to have a default value for re-usability

I believe you can set a PORT value in the Dockerfile that will be overwritten at runtime. Or you can do a null check in this script

loadgenerator/loadgen
```diff
@@ -458,6 +496,13 @@ def _get_parser():
             description='Show the status of all load generator jobs whether running or not')
     subparser.set_defaults(func=status)
 
+    subparser = subparsers.add_parser('autorun',
+            help='Create load generator with Cloud Run',
+            description='automatically starts a load generation job for a target address with Cloud Run')
+    subparser.add_argument('target_address', type=str,
+            help='The target address (without http/s prefix) to send laod to')
+    subparser.set_defaults(func=run_auto)
```
Daniel-Sanche, line 504: Can you think of a way to test this? Maybe write a pytest like Nick's or Levi's?

gzhao408: Yeah definitely. Test the parser or test the autorun function?

Daniel-Sanche: Both could be useful. I know we're getting short on time, so whatever you think would be highest impact. It would be nice if we could get some testing in at least though

gzhao408: For sure!

loadgenerator/loadgen
```diff
@@ -400,6 +399,45 @@ def _check_prereqs():
...
+    logger.info('Creating load generation job {0}...'.format(name))
+
+    if not _is_address_accessible(args.target_address):
+        logger.error('Could not access target address {0}'.format(args.target_address))
+        print('Please double check the target address and try again.')
+        return
+
+    command = (
+        'gcloud run deploy {0}'
+        ' --image gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator:gz'
```
simonz130, line 422: the tag should be the release tag or alternatively, if we push the loadgen image to the project where sandbox is created, it should be latest tag from that project.

loadgenerator/loadgen
```diff
@@ -400,6 +399,45 @@ def _check_prereqs():
...
+    # _setup_cloud_run()
+    name = _generate_job_name()
+    logger.info('Creating load generation job {0}...'.format(name))
+
+    if not _is_address_accessible(args.target_address):
+        logger.error('Could not access target address {0}'.format(args.target_address))
+        print('Please double check the target address and try again.')
+        return
+
+    command = (
```
simonz130, line 420: document what you do here

loadgenerator/loadgen
```diff
@@ -400,6 +399,45 @@ def _check_prereqs():
...
+        ' --update-env-vars="FRONTEND_ADDR={1}"'
+        ' --allow-unauthenticated'
+    ).format(name, args.target_address)
+
+    subprocess.call(['echo'] + split(command))
+    try:
+        output = subprocess.check_output(split(command)).decode('utf-8')
+        print(output)
+    except subprocess.CalledProcessError:
+        return
```
simonz130, line 432: log human readable error message here

Daniel-Sanche:
Test link in case it helps reviewers: https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/gzhao408/stackdriver-sandbox.git&cloudshell_git_branch=loadgen-step&cloudshell_working_dir=terraform&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image:v0.2.5

loadgenerator/loadgen
```diff
@@ -400,6 +399,45 @@ def _check_prereqs():
...
+
+def _setup_cloud_run():
+    """Configure Cloud Run with default settings so that the user isn't
+    prompted every time a command runs.
+    """
+    logger.info('Setting default platform and region')
+    subprocess.run(split('gcloud config set run/platform managed'))
+    subprocess.run(split('gcloud config set run/region us-central1'))
```
simonz130, line 440: Not sure what is the advantage of having this wrapped in python script as opposed to being a bash script. Could you elaborate?

loadgenerator/loadgen
```diff
@@ -458,6 +496,13 @@ def _get_parser():
             description='Show the status of all load generator jobs whether running or not')
     subparser.set_defaults(func=status)
 
+    subparser = subparsers.add_parser('autorun',
+            help='Create load generator with Cloud Run',
+            description='automatically starts a load generation job for a target address with Cloud Run')
+    subparser.add_argument('target_address', type=str,
+            help='The target address (without http/s prefix) to send laod to')
```
Daniel-Sanche, line 503: It would be great if we could find this automatically instead of expecting the user to know it. Levi did that [here](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/dd3341773b4c3aa60926d06b2e0d18b703d48321/tests/provisioning/provision_test.py#L58). You also may need to authenticate kubectl first to talk to the cluster, which he did [here](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/dd3341773b4c3aa60926d06b2e0d18b703d48321/tests/provisioning/provision_test.py#L122)

loadgenerator/loadgen
```diff
@@ -400,6 +399,45 @@ def _check_prereqs():
...
+    if not _is_address_accessible(args.target_address):
+        logger.error('Could not access target address {0}'.format(args.target_address))
+        print('Please double check the target address and try again.')
+        return
+
+    command = (
+        'gcloud run deploy {0}'
+        ' --image gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator:gz'
+        ' --update-env-vars="FRONTEND_ADDR={1}"'
+        ' --allow-unauthenticated'
```
Daniel-Sanche, line 424: I get a prompt asking for a target platform. You should be able to set the value as an argument here (we want fully managed)

Daniel-Sanche: You may also want to set the region, but that might also make sense to keep manual

loadgenerator/loadgen
```diff
@@ -458,6 +496,13 @@ def _get_parser():
             description='Show the status of all load generator jobs whether running or not')
     subparser.set_defaults(func=status)
 
+    subparser = subparsers.add_parser('autorun',
```
Daniel-Sanche, line 499: I got confused by the difference between `autostart` and `autorun`. Is `autostart` going away? Or are they two different things?

loadgenerator/loadgen
```diff
@@ -400,6 +399,45 @@ def _check_prereqs():
...
+        ' --allow-unauthenticated'
+    ).format(name, args.target_address)
+
+    subprocess.call(['echo'] + split(command))
+    try:
+        output = subprocess.check_output(split(command)).decode('utf-8')
+        print(output)
+    except subprocess.CalledProcessError:
+        return
+
```
Daniel-Sanche, line 433: It took a couple seconds between the command completing and the site being available for me. It might be useful to wait for HTTP 200s before finishing the command

loadgenerator/loadgen
gzhao408: I guess I'm just more familiar with python. Don't think there's any advantage/disadvantage

loadgenerator/loadgen
gzhao408: :O they should be automatically set in the _setup_cloud_run() function?

gzhao408: Oh oops it got commented out somehow - I'll put it back in with the next push

loadgenerator/loadgen
gzhao408: autostart is the original one using GCE, it's what's called in the install script - wasn't sure if I wanted to get rid of it yet

loadgenerator/loadgen
gzhao408: awesome, I'll try to do that instead

loadgenerator/loadgen
gzhao408: Yeah I was going to replace it with latest. Hopefully this helps with testing (I didn't push a new latest tag, used a gz tag). I'll also replace the GCE image with the one tagged by commit so it doesn't change.

Daniel-Sanche:
I believe Cloud Run will scale to 0 if the web Ui isn't getting traffic. It looks like we might need to [set up a load generator on our load generator](https://github.com/ahmetb/cloud-run-faq#how-to-keep-a-cloud-run-service-warm) if we want to keep it live in the background (unless there's a min-replicas argument I missed somewhere?).

I wonder if we can make our locust instance ping itself :thinking:

gzhao408:
> I believe Cloud Run will scale to 0 if the web Ui isn't getting traffic. It looks like we might need to [set up a load generator on our load generator](https://github.com/ahmetb/cloud-run-faq#how-to-keep-a-cloud-run-service-warm) if we want to keep it live in the background (unless there's a min-replicas argument I missed somewhere?).

😭   So the load generator might just stop running?

Thanks for scouting info 🙏  For context if it helps, the idea is to use Cloud Run in two scenarios:
- The automatic loadgen with the normal Web UI (which is what `autorun` does) - the plan is to give this link to the user and they'll hopefully send requests to it from their browser. If not, I guess it'll scale to 0 and need a cold start later on when the user wants it again?
- A temporary one that sends a specific traffic pattern with no UI; I had planned to set a time limit for these, let them scale to 0, and then basically "garbage collect" in the destroy script. I would assume when Cloud Run first starts, it'll stay up for a little bit of time even if it doesn't receive traffic?

gzhao408:
GKE is an option if Cloud Run is just unsuitable, we have the container now anyway

loadgenerator/loadgen
gzhao408: 👍  will do

Daniel-Sanche:
Ok, I talked to Ahmet (Cloud Run expert) and he confirmed it will scale to 0 on the fully managed version. We'd need to move to self-hosting with Anthos to get more control, which isn't worth it. He suggested wither using GKE or deploying the container on GCE (which is what we're doing now, what was wrong with that approach again?) At least since we have a container, moving should be easy if we decide to

If you really just want these to be temporary, we could probably get away with Cloud Run. The FAQ says it should be able to  keep working for 15 minutes after a request. But it wasn't designed for this kind of use case, so we may need to fight against the product for it to work right (and set a bad example for customers). And I don't have a ton of experience here, so it's hard to foresee what other issues might come up

gzhao408:
I think the idea was that GCE isn't super flexible, and we'd want to be able to scale to thousands of users in the load generator... but the bottleneck right now is honestly the Hipster Shop that can't handle all the requests. We could keep GCE 🤷

gzhao408:
Closing, will open a new one with GKE implementation

PR #406, opened 2020-08-27
End-to-end release test

Daniel-Sanche:
We currently have an end-to-end test in place to test the latest code in the repo. This PR adds a second test that tests our latest release. It scrapes the version tags from the public website (testing that the website is live in the process), and then uses those values to launch the rest of our e2e tests.

This one isn't tied to any code changes, so I set it to run daily and manually when requested


.github/workflows/e2e-release.yml
```diff
@@ -0,0 +1,95 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "End to End - Release"
+on:
+  schedule:
+    # run every day at 8pm
+    - cron:  '* 20 * * *'
```
simonz130, line 19: 8 PM of which time zone?

Daniel-Sanche: Good point, it looks like GitHub Actions uses UTC. I don't think the timing matters too much though, just want to make sure it runs frequently

PR #420, opened 2020-08-31
Add RatingService to the demo app

jiang95-dev:
- [x] Deploy the RatingService to the App Engine.

- [ ] Communication with ProductCatalogService.

- [ ] Frontend UI.

- [x] Auto provisioning.

- [x] Testing.


src/productcatalogservice/server.go
```diff
@@ -53,7 +53,9 @@ import (
 	"go.opentelemetry.io/otel/instrumentation/grpctrace"
 	apitrace "go.opentelemetry.io/otel/api/trace"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
-	"go.opentelemetry.io/otel/sdk/resource"
+    "go.opentelemetry.io/otel/sdk/resource"
+
```
simonz130, line 57: nit: remove empty space

src/productcatalogservice/server.go
```diff
@@ -288,7 +290,11 @@ func (p *productCatalog) GetProduct(ctx context.Context, req *pb.GetProductReque
 	}
 	if found == nil {
 		return nil, status.Errorf(codes.NotFound, "no product with ID %s", req.Id)
-	}
+    }
+    resp, err := http.Get("http://cloud-ops-sandbox-406964100.wl.r.appspot.com/getRating/66VCHSJNUP")
```
simonz130, line 294: the value of the server needs to come from the config - it will be different each time the app is deployed on a different sandbox. You also need to set this value in the config as part of provisioning (i.e. first deploy appEngine app, get its address, inject into config of hipstershop for productcatalogservice, then deploy hipstershop with this config

src/ratingservice/app.yaml
```diff
@@ -0,0 +1,10 @@
+# [START gae_python38_cloudsql_config]
+runtime: python38
+
+env_variables:
+  CLOUD_SQL_USERNAME: postgres
+  CLOUD_SQL_PASSWORD: 520jiang
+  CLOUD_SQL_DATABASE_NAME: postgres
+  CLOUD_SQL_CONNECTION_NAME: cloud-ops-sandbox-406964100:us-west2:rating
```
simonz130, line 8: you can't checkin sql password or database name that includes the project id. You need to auto-populate these

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+
+db_user = os.environ.get('CLOUD_SQL_USERNAME')
+db_password = os.environ.get('CLOUD_SQL_PASSWORD')
+db_name = os.environ.get('CLOUD_SQL_DATABASE_NAME')
+db_connection_name = os.environ.get('CLOUD_SQL_CONNECTION_NAME')
+
+host = '/cloudsql/{}'.format(db_connection_name)
+
+def init():
+    products = ['OLJCESPC7Z','66VCHSJNUP','1YMWWN1N4O','L9ECAV7KIM','2ZYFJ3GM2N','0PUK6V6EV0','LS4PSXUNUM','9SIQT8TOJO', '6E92ZMYYFZ']
```
simonz130, line 27: comment on where these come from

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+    conn = psycopg2.connect(dbname=db_name, user=db_user, password=db_password, host=host)
+    with conn.cursor() as cursor:
+        cursor.execute("SELECT EXISTS(SELECT * FROM information_schema.tables WHERE table_name = 'ratings');")
+        result = cursor.fetchone()
+        if not result[0]:
+            cursor.execute("CREATE TABLE ratings (product_id varchar(20) PRIMARY KEY, sum_of_score int DEFAULT 0, count_of_score int DEFAULT 0);")
+            for product in products:
+                cursor.execute("INSERT INTO ratings (product_id) VALUES ('{}');".format(product))
+    conn.commit()
+    conn.close()
```
simonz130, line 37: This is better to do as part of provisioning script as opposed to in code.

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+
+app = Flask(__name__)
+init()
+
+@app.route('/getRating/<id>')
+def getRating(id):
+    conn = psycopg2.connect(dbname=db_name, user=db_user, password=db_password, host=host)
+    try:
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT product_id, sum_of_score, count_of_score FROM ratings WHERE product_id='{}';".format(id))
```
simonz130, line 47: If you store aggregations, you'll have a problem when multiple users will try to update the database with rating at the same time. One of them will have to lock the entire database and the rest will wait.

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+    return str(rating)
+
+@app.route('/rate/<id>/<score>')
+def rate(id, score):
+    conn = psycopg2.connect(dbname=db_name, user=db_user, password=db_password, host=host)
+    try:
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT product_id, sum_of_score, count_of_score FROM ratings WHERE product_id='{}';".format(id))
+            result = cursor.fetchone()
+            cursor.execute("UPDATE ratings SET sum_of_score={0}, count_of_score={1} WHERE product_id='{2}';".format(int(score) + result[1], 1 + result[2], id))
```
simonz130, line 62: What happens if this code is being executed in parallel by multiple callers?

src/productcatalogservice/server.go
```diff
@@ -288,7 +290,11 @@ func (p *productCatalog) GetProduct(ctx context.Context, req *pb.GetProductReque
...
+    defer resp.Body.Close()
+    body, err := ioutil.ReadAll(resp.Body)
+    log.Info(body)
```
Daniel-Sanche, line 297: this indenting doesn't match the line below

src/ratingservice/main.py
Daniel-Sanche: I think we should at very least be reading these out of a file (maybe copy products.json here?)

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+# limitations under the License.
+
+import os
+from flask import Flask
+import psycopg2
+
+db_user = os.environ.get('CLOUD_SQL_USERNAME')
+db_password = os.environ.get('CLOUD_SQL_PASSWORD')
+db_name = os.environ.get('CLOUD_SQL_DATABASE_NAME')
+db_connection_name = os.environ.get('CLOUD_SQL_CONNECTION_NAME')
```
Daniel-Sanche, line 22: Do you fail gracefully if any environment variables are unset?

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+    finally:
+        conn.close()
+    return str(rating)
+
+@app.route('/rate/<id>/<score>')
+def rate(id, score):
+    conn = psycopg2.connect(dbname=db_name, user=db_user, password=db_password, host=host)
+    try:
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT product_id, sum_of_score, count_of_score FROM ratings WHERE product_id='{}';".format(id))
```
Daniel-Sanche, line 60: I would really prefer something like sqlalchemy instead of seeing all these SQL strings in the code. But I'm not sure if it's worth pushing for this

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+            result = cursor.fetchone()
+            cursor.execute("UPDATE ratings SET sum_of_score={0}, count_of_score={1} WHERE product_id='{2}';".format(int(score) + result[1], 1 + result[2], id))
+        conn.commit()
+    finally:
+        conn.close()    
+    return 'Success'
+
+@app.route('/')
+def index():
+    return "Hello!"
```
Daniel-Sanche, line 70: Is this needed?

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+            result = cursor.fetchone()
+        rating = result[1] / result[2] if result[2] != 0 else 0
+        conn.commit()
+    finally:
+        conn.close()
+    return str(rating)
+
+@app.route('/rate/<id>/<score>')
+def rate(id, score):
+    conn = psycopg2.connect(dbname=db_name, user=db_user, password=db_password, host=host)
```
Daniel-Sanche, line 57: You shouldn't have to form a new database handshake for each request, right?

src/productcatalogservice/server.go
Daniel-Sanche: > inject into config of hipstershop for productcatalogservice

I don't think we should be re-writing the kubernetes manifests manually. We should probably add it dynamically as a [ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#create-configmaps-from-literal-values). We'd also have to add some detection code here to make sure it was set properly

Daniel-Sanche (dismissed):
I have a lot of concerns about adding this new, complex service in the last week. I think it would still be easier to just transfer an existing service to AppEngine

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+    try:
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT product_id, sum_of_score, count_of_score FROM ratings WHERE product_id='{}';".format(id))
+            result = cursor.fetchone()
```
Daniel-Sanche, line 61: What if no product is found? I don't see any error handling in place here

src/ratingservice/main.py
```diff
@@ -0,0 +1,71 @@
...
+        rating = result[1] / result[2] if result[2] != 0 else 0
+        conn.commit()
+    finally:
+        conn.close()
+    return str(rating)
```
Daniel-Sanche, line 53: If the input id isn't valid, it will return `str(None)` and crash the service

src/ratingservice/main.py
Daniel-Sanche: The server should return appropriate HTTP error codes if the input is incorrect

src/productcatalogservice/server.go
jiang95-dev: I have removed changes from server.go

src/productcatalogservice/server.go
jiang95-dev: I have removed changes from server.go

src/productcatalogservice/server.go
jiang95-dev: I have removed changes from server.go

src/ratingservice/main.py
jiang95-dev: This is removed.

src/ratingservice/main.py
jiang95-dev: I'm using a connection pool now.

src/ratingservice/main.py
jiang95-dev: I return json response now. In failure it returns 'fail'.

src/ratingservice/main.py
jiang95-dev: It won't fail gracefully because the server can't proceed without db connection strings.

src/ratingservice/main.py
jiang95-dev: If product not found, it will return 404 now.

src/ratingservice/app.yaml
jiang95-dev: It is provisioned in Terraform now.

src/ratingservice/main.py
jiang95-dev: I just create an empty table now. No need to populate items.

src/ratingservice/main.py
jiang95-dev: It's in install.py now.

src/ratingservice/main.py
jiang95-dev: I changed the relational schema.

src/ratingservice/main.py
jiang95-dev: I use the default isolation level which is read_uncommitted since it doesn't require high accuracy.

src/ratingservice/main.py
Daniel-Sanche: What does the error look like though? Will it be sensible to userss? It's usually a good practice to check the variable values and exit(1) with a clear message if they don't look right

jiang95-dev: To users, it will be 500 server internal error since the server hasn't started it yet. The error is auto reported to the Error Reporting.

src/ratingservice/05_app_engine.tf
```diff
@@ -0,0 +1,38 @@
...
+  runtime    = "python38"
+
+  entrypoint {
+    shell = "gunicorn -b :8080 main:app"
+  }
+
+  deployment {
+    zip {
+      # Need to zip source code and upload to GCS
+      source_url = "https://storage.googleapis.com/${var.bucket_name}/ratingservice_code.zip"
```
Daniel-Sanche, line 13: Can you explain this a bit more? Is it automated?

src/ratingservice/main.py
```diff
@@ -0,0 +1,96 @@
...
+
+@app.route('/getRating/<id>')
+def getRating(id):
+    conn = connpool.getconn()
+    resp = None
+    try:
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT AVG(score), COUNT(*) FROM ratings WHERE product_id='{}';".format(id))
+            result = cursor.fetchone()
+            if result[1] > 0:
```
Daniel-Sanche, line 45: what's result[1]? Please comment and/or give it a variable name

Could you do `rating, count = cursor.fetchone()`?

src/ratingservice/main.py
```diff
@@ -0,0 +1,96 @@
...
+                resp.status_code = 404   
+        conn.commit()    
+    except:
+        data = {
+            'status'  : 'fail',
+            'message' : 'Internal server error'
+        }
+        resp = jsonify(data)
+        resp.status_code = 500
+        return resp
```
Daniel-Sanche, line 68: If you return here, won't it skip over the finally?

It's probably a better idea to use the same return later

src/ratingservice/main.py
```diff
@@ -0,0 +1,96 @@
...
+    finally:
+        connpool.putconn(conn)
+    return resp
```
Daniel-Sanche, line 71: This is a very large function without any comments. I'd prefer if you could find a way to make it more readable. 

For example: you probably don't need to build and send all the `fail` json messages. I think the HTTP status code gives essentially the same information

I'd also suggest building the `resp` object in one place at the bottom. You can assign different values to a `status_code` and `data` variable in each branch, but just build the response in one place

src/ratingservice/main.py
```diff
@@ -0,0 +1,96 @@
...
```
Daniel-Sanche, line 96: same issues as above here:
1) you shouldn't be returning early if there's a finally clause
2) there seems to be a lot of boiler plate code, making this harder to follow

src/ratingservice/main_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+import os
+import unittest
+import requests
+
+class TestEndpoints(unittest.TestCase):
+
+    def testRate(self):
+        """ Test if rating a product returns success """
+        products = ['OLJCESPC7Z','66VCHSJNUP','1YMWWN1N4O','L9ECAV7KIM','2ZYFJ3GM2N','0PUK6V6EV0','LS4PSXUNUM','9SIQT8TOJO', '6E92ZMYYFZ']
```
Daniel-Sanche, line 25: Are these products still hardcoded in your db somewhere? I don't see it in your new code

src/ratingservice/main_test.py
```diff
@@ -0,0 +1,54 @@
...
+            res = requests.get(url)
+            self.assertEqual(res.json()['status'], 'success')
+
+    
+    def testGetRatingCorrect(self):
+        """ Test if getting the rating of a product returns a correct number """
+        url_get = "https://{0}.wl.r.appspot.com/getRating/{1}".format(getProjectId(), "OLJCESPC7Z")
+        url_post = "https://{0}.wl.r.appspot.com/rate/{1}/{2}".format(getProjectId(), "OLJCESPC7Z", 5)
+        res1 = requests.get(url_get).json()
+        requests.get(url_post)
```
Daniel-Sanche, line 45: shouldn't this be requests.post?

src/ratingservice/main_test.py
```diff
@@ -0,0 +1,54 @@
...
+        res2 = requests.get(url_get).json()
+        self.assertTrue(abs(float(res1['rating']) * float(res1['count']) + 5 
+                        - float(res2['rating']) * float(res2['count'])) < 1e-5)
```
Daniel-Sanche, line 48: This could use some commenting at the very least. What's the expected result of res1? Shouldn't you also be testing that the count has incremented>

src/ratingservice/main_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+class TestEndpoints(unittest.TestCase):
+
+    def testRate(self):
+        """ Test if rating a product returns success """
+        products = ['OLJCESPC7Z','66VCHSJNUP','1YMWWN1N4O','L9ECAV7KIM','2ZYFJ3GM2N','0PUK6V6EV0','LS4PSXUNUM','9SIQT8TOJO', '6E92ZMYYFZ']
+        for product in products:
+            url = "https://{0}.wl.r.appspot.com/rate/{1}/{2}".format(getProjectId(), product, 5)
+            res = requests.get(url)
+            self.assertEqual(res.json()['status'], 'success')
```
Daniel-Sanche, line 29: You should also be making assertions on the data somewhere. And testing your failure cases

src/ratingservice/main_test.py
```diff
@@ -0,0 +1,54 @@
...
+    
+    def testGetRating(self):
+        """ Test if getting the rating of a product returns success """
+        products = ['OLJCESPC7Z','66VCHSJNUP','1YMWWN1N4O','L9ECAV7KIM','2ZYFJ3GM2N','0PUK6V6EV0','LS4PSXUNUM','9SIQT8TOJO', '6E92ZMYYFZ']
+        for product in products:
+            url = "https://{0}.wl.r.appspot.com/getRating/{1}".format(getProjectId(), product)
+            res = requests.get(url)
+            self.assertEqual(res.json()['status'], 'success')
```
Daniel-Sanche, line 37: Same as above - make sure the json results are as you expect them to be

src/ratingservice/05_app_engine.tf
jiang95-dev: Yeah, we can keep the zip file in source code or zip them in install.sh. And then we need to upload the zip file to GCS.

src/ratingservice/main.py
jiang95-dev: Finally will get executed no matter what. So it will execute after return.

src/ratingservice/05_app_engine.tf
Daniel-Sanche: Is that automation coming later? I don't see it here

src/ratingservice/main_test.py
jiang95-dev: It's in products.json.

src/ratingservice/main_test.py
jiang95-dev: Yeah. I changed it.

src/ratingservice/main_test.py
jiang95-dev: I added a failure case.

src/ratingservice/main_test.py
jiang95-dev: They are what I expected.

src/ratingservice/main_test.py
jiang95-dev: I added the comments.

src/ratingservice/main.py
Daniel-Sanche: Interesting. I'd still argue that returning in an except block is an antipattern because that behaviour is non-obvious, but I don't think this is too high priority

src/ratingservice/05_app_engine.tf
jiang95-dev: This is not in scope now since the service will only be integrated afterwards. I won't modify the install.sh for now.

src/ratingservice/main.py
jiang95-dev: I have given them name now.

src/ratingservice/main.py
jiang95-dev: I removed the return now.

src/ratingservice/main.py
jiang95-dev: I have make it clearer.

src/ratingservice/main.py
jiang95-dev: I remove the return now.

src/ratingservice/main_test.py
jiang95-dev: I use http code now.

src/ratingservice/.gcloudignore
```diff
@@ -0,0 +1,19 @@
+# This file specifies files that are *not* uploaded to Google Cloud Platform
```
simonz130, line 1: you might need to include the license header there

src/ratingservice/05_app_engine.tf
```diff
@@ -0,0 +1,43 @@
...
+  runtime    = "python38"
+
+  entrypoint {
+    shell = "uwsgi --http-socket :8080 --wsgi-file main.py --callable app --master --processes 1 --threads 2"
+  }
+
+  deployment {
+    zip {
+      # Need to zip source code and upload to GCS
+      source_url = "https://storage.googleapis.com/${var.bucket_name}/ratingservice_code.zip"
```
simonz130, line 27: can the name include the version tag?

src/ratingservice/05_app_engine.tf
```diff
@@ -0,0 +1,43 @@
...
+
+  env_variables = {
+    CLOUD_SQL_DATABASE_NAME = google_sql_database.database.name
+    CLOUD_SQL_USERNAME = google_sql_user.default.name
+    CLOUD_SQL_PASSWORD = google_sql_user.default.password
+    CLOUD_SQL_CONNECTION_NAME = google_sql_database_instance.master.connection_name
+  }
+
+  basic_scaling {
+    max_instances = 5
```
simonz130, line 39: add a note about this setting

src/ratingservice/main.py
```diff
@@ -0,0 +1,115 @@
...
+            res.append(product['id'])
+    return res
+
+# create and populate table
+def populate_database():
+    try:
+        products = read_products()
+        conn = connpool.getconn()
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT EXISTS(SELECT * FROM information_schema.tables WHERE table_name = 'ratings');")
```
simonz130, line 53: this sql statement is weird - you probably can replace it with:

```sql
SELECT * FROM information_schema.tables WHERE table_name = 'ratings' limit 1 
```

src/ratingservice/main.py
```diff
@@ -0,0 +1,115 @@
...
+            result = cursor.fetchone()
+            if not result[0]:
+                cursor.execute("CREATE TABLE ratings (id SERIAL PRIMARY KEY, product_id varchar(20) NOT NULL, score int DEFAULT 0);")
```
simonz130, line 56: are you using ids for ratings you insert from the json file or are you relying on the ids starting from 1?

src/ratingservice/main.py
```diff
@@ -0,0 +1,115 @@
...
+                # product exists
+                resp = jsonify({
+                    'status' : 'success',
+                    'rating' : str(rating),
+                    'count'  : str(count)
+                })
+                resp.status_code = 200
+            else:
+                # product not exists
+                resp = jsonify({'status' : 'fail'})
```
simonz130, line 87: Produce more meaningful error message that explains what is the failure

src/ratingservice/main_test.py
```diff
@@ -0,0 +1,76 @@
...
+            self.assertEqual(res.status_code, 200)
+    
+    def testGetRatingCorrect(self):
+        """ Test if getting the rating of a product returns a correct number """
+        url_get = "https://ratingservice-dot-{0}.wl.r.appspot.com/getRating/{1}".format(getProjectId(), "OLJCESPC7Z")
+        url_post = "https://ratingservice-dot-{0}.wl.r.appspot.com/rate".format(getProjectId())
+        res1 = requests.get(url_get).json()
+        requests.post(url_post, data={
+            'score' : 5,
+            'id' : 'OLJCESPC7Z'
```
simonz130, line 55: can we read this from the same file that is used to populate the database? Otherwise it feels weird - not clear where this is coming from

src/ratingservice/products.json
```diff
@@ -0,0 +1,110 @@
...
+            "id": "OLJCESPC7Z",
+            "name": "Vintage Typewriter",
+            "description": "This typewriter looks good in your living room.",
+            "picture": "/static/img/products/typewriter.jpg",
+            "priceUsd": {
+                "currencyCode": "USD",
+                "units": 67,
+                "nanos": 990000000
+            },
+            "categories": ["vintage"]
```
simonz130, line 13: Why do we need all of these fields in the product object that is only used by rating service?
Perhaps, id and name is enough for working with ratings? Do I miss anything?

src/ratingservice/main_test.py
```diff
@@ -0,0 +1,76 @@
...
+                        - float(res2['rating']) * float(res2['count'])) < 1e-5)
+    
+
+def getProjectId():
+    return os.environ['GOOGLE_CLOUD_PROJECT']
+
+# read product ids
+def read_products():
+    res = []
+    with open('products.json') as f:
```
simonz130, line 69: tests file should be placed with the rest of the tests we have, not in the same folder as the service code. As a result, you might need to read products.json assuming it's not on the same directory

src/ratingservice/main.py
```diff
@@ -0,0 +1,115 @@
...
+        conn.commit()
+    finally:
+        connpool.putconn(conn)
+
+app = Flask(__name__)
+connpool = initConnection()
+populate_database()
+
+# get rating of a product
+@app.route('/getRating/<id>')
```
simonz130, line 68: This will technically work but because we create a REST endpoint, it just isn't right from design perspective.
The route should be your resource (e.g. **rating**) and the method should be GET.
A good post on this (look at use nouns in the URI)
https://medium.com/hashmapinc/rest-good-practices-for-api-design-881439796dc9

src/ratingservice/main.py
```diff
@@ -0,0 +1,115 @@
...
+        conn.commit()    
+    except:
+        resp = jsonify({'status'  : 'fail'})
+        resp.status_code = 500
+    finally:
+        connpool.putconn(conn)
+    return resp
+
+# rate a product
+@app.route('/rate', methods=['POST'])
```
simonz130, line 98: This isn't right either - you should use HTTP PUT here (update semantics) and you should use the same /rating resource but also with id: /rating/{id} and incoming http request would include rating object

simonz130 (changes requested):

src/ratingservice/.gcloudignore
jiang95-dev: I added it.

src/ratingservice/05_app_engine.tf
jiang95-dev: I added it.

src/ratingservice/05_app_engine.tf
jiang95-dev: I added it now.

src/ratingservice/05_app_engine.tf
jiang95-dev: I added it.

src/ratingservice/main.py
jiang95-dev: It's auto-incremental id string from 1.

src/ratingservice/main_test.py
jiang95-dev: I changed it.

src/ratingservice/products.json
jiang95-dev: It's from the productcatalogservice. I've removed it.

src/ratingservice/main_test.py
jiang95-dev: I changed the location.

src/ratingservice/main.py
jiang95-dev: I changed it.

src/ratingservice/main.py
jiang95-dev: I changed it.

jiang95-dev: I believe according to the article, GET should be /rating/<id> and POST should be /rating.

src/ratingservice/main.py
jiang95-dev: I think most people use the sql statement in the above according to my web search?

src/ratingservice/main.py
jiang95-dev: I added it.

src/ratingservice/main.py
simonz130: What is the purpose of this id field? You are setting primary key on it but then when querying the table, you are going to use product_id field (which is also the field on which you need to create an index for faster querying).

You can remove the id field and set product_id to have unique constraint and to be a primary key

src/ratingservice/main.py
```diff
@@ -0,0 +1,115 @@
...
+    product_id = request.form['id']
+    score = request.form['score']
+    try:
+        with conn.cursor() as cursor:
+            cursor.execute("INSERT INTO ratings (product_id, score) VALUES ('{0}', {1});".format(product_id, score))
+            resp = jsonify({'status' : 'success'})
+            resp.status_code = 200
+        conn.commit()
+    except:
+        resp = jsonify({'error' : 'Unexpected error in database'})
```
simonz130, line 111: This error is not descriptive. It's better to say something about the operation that failed. At the very least - "failed to add a rating for XXX" where XXX is the product id

terraform/serverless/05_app_engine.tf
```diff
@@ -0,0 +1,43 @@
...
+  service    = "ratingservice"
+  runtime    = "python38"
+
+  entrypoint {
+    shell = "uwsgi --http-socket :8080 --wsgi-file main.py --callable app --master --processes 1 --threads 2"
+  }
+
+  deployment {
+    zip {
+      source_url = "https://storage.googleapis.com/${var.bucket_name}/ratingservice_code:latest.zip"
```
simonz130, line 26: instead of "latest" it should be a variable that holds the value of the release. It's value should be "latest" by default and be updated upon releasing to corresponding version of the release.

terraform/serverless/install_serverless.sh
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#!/bin/bash
+
+upload_source_code(){
+    source_name="ratingservice_code:latest.zip"
```
simonz130, line 17: same comment as in 05_app_engine.tf about "latest"

src/ratingservice/main.py
jiang95-dev: Reply to - what is the purpose of the id field: This is because we insert a row at every user rating. If we set product_id to be the primary key, it won't allow duplicate, so there can be only one rating for a product.

Reply to - error not descriptive: I changed it.

terraform/serverless/05_app_engine.tf
jiang95-dev: I added it.

terraform/serverless/install_serverless.sh
jiang95-dev: I added it.

terraform/serverless/05_app_engine.tf
```diff
@@ -0,0 +1,49 @@
...
+    max_instances = 5
+  }
+
+  delete_service_on_destroy = true
+}
+
+variable "code_tag" {
+  type        = string
+  description = "The release tag of the source code. Case-sensitive."
+  default     = "latest"
```
simonz130, line 48: Can this be provided via command line when invoking terraform script? This way it's going to be easier to update/replace it upon releasing a new version. If we keep it here, we need to do replacements of tf file content for the sake of releasing.

terraform/serverless/install_serverless.sh
```diff
@@ -0,0 +1,26 @@
...
+    if [[ -n $1 ]]; then
+        source_name="ratingservice_code:$1.zip"
+    else
+        source_name="ratingservice_code:latest.zip"
+    fi
+    zip -j ${source_name} ../../src/ratingservice/main.py ../../src/ratingservice/requirements.txt ../../src/productcatalogservice/products.json
+    gsutil cp ${source_name} "gs://$GOOGLE_CLOUD_PROJECT-bucket/${source_name}"
+}
+
+upload_source_code $1;
```
simonz130, line 26: The name of your script is misleading. It implies that you are going to deploy a service but in fact it's only pushing the source code into storage bucket. You should add a call to terraform that installs your service into this script (for now) and then in the future we might want to integrate this install script into broader terraform provisioning of the sandbox

terraform/serverless/05_app_engine.tf
jiang95-dev: Yes. This variable resource can capture the command line arguments.

simonz130 (approved):

Daniel-Sanche (commented):
Looks like my comments were adressed. I haven't done a full review, but I'll remove my change request for now

PR #421, opened 2020-08-31
Add telemetry to Terraform scripts

### LINKED ISSUE #380: Install cloud monitoring in Custom Cloud Shell Image
Currently there's a line in `install.sh` which installs Cloud Monitoring: ` python3 -m pip install google-cloud-monitoring`.

It would make more sense in terms of workflow to install it in the Custom Cloud Shell image, which already handles terraform and python3 installations.

### LINKED ISSUE #412: [telemetry] Add python subscriber script to report from project to Pub/Sub in GCP
In order to learn how frequently the installation script succeeds, and how often users run destroy script, we should add a subscriber script to report information to Pub / Sub in GCP.

eaberman:
Fixes #412 and fixes #380 

## TESTING INFORMATION:

The telemetry system can be tested by using this link and running `install.sh` and `destroy.sh` scripts: https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/GoogleCloudPlatform/stackdriver-sandbox.git&cloudshell_git_branch=telemetry-system&cloudshell_working_dir=terraform&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image:telemetry_test2

Then, to verify that the messages have been sent, view the messages in the Pub/Sub topic here: https://pantheon.corp.google.com/cloudpubsub/topic/detail/telemetry?project=stackdriver-sandbox-230822

## Description

The script `telemetry.py` is a straightforward script that writes messages to a telemetry topic in stackdriver-sandbox's Pub / Sub. Every message include arguments passed to the python script, and a formatted-datetime.

The `install.sh` and `destroy.sh` files call the script with arguments `event` and `project-id`, sending information in the following events.

In `install.sh`:
- the user is creating a project and is a googler
- the user is creating a project and is not a googler
- no active billing account is found for the user
- HipsterShop application is up
- HipsterShop application is unreachable
- loadgenerator app is up
- loadgenerator app is unreachable

In `destroy.sh`:
- `destroy` script is called but project is not deleted
- `destroy` script is called and project is deleted


terraform/install.sh
```diff
@@ -157,6 +159,7 @@ createProject() {
       log "If you don't have access to this folder, please make sure to request at:"
       log "go/experimental-folder-access"
       log ""
+      python telemetry.py event=googler-project project=$project_id version=$VERSION
```
simonz130, line 162: Either the name is not descriptive or you will have a hard time knowing from the telemetry that this event instance is associated with creating a new project

terraform/install.sh
```diff
@@ -37,6 +38,7 @@ promptForBillingAccount() {
     log ""
     log "To list active billing accounts, run:"
     log "gcloud beta billing accounts list --filter open=true"
+    python telemetry.py event=no-active-billing version=$VERSION
```
simonz130, line 41: All the executions of your telemetry script from within the same install.sh (or terraform) need to be correlated. I.e. we need to know that creation of the sandbox, and then loadgen and all other events are part of the same larger event of a user trying to create a new Sandbox. You should generate a UUID and then use it in all the executions of the telemetry script within the same session of running install.sh

terraform/telemetry.py
```diff
@@ -0,0 +1,39 @@
+from google.cloud import pubsub_v1
```
simonz130, line 1: add copyright header

terraform/telemetry.py
```diff
@@ -0,0 +1,39 @@
...
+from datetime import datetime
+import sys
+
+project_id = "stackdriver-sandbox-230822"
+topic_id = "telemetry"
+
+publisher = pubsub_v1.PublisherClient()
+# The `topic_path` method creates a fully qualified identifier
+# in the form `projects/{project_id}/topics/{topic_id}`
+topic_path = publisher.topic_path(project_id, topic_id)
```
simonz130, line 11: how authentication will work? Will everyone be able to send telemetry from any project into our stackdriver-sandbox-230822 project?

terraform/telemetry.py
```diff
@@ -0,0 +1,39 @@
...
+
+# define the following helper functions
+def get_datetime_str():
+    # send current date and time for end of installation script
+    now = datetime.now()
```
simonz130, line 16: All the dates should be in UTC format to mitigate time differences associated with multiple users located across the world and try to create sandboxes

terraform/telemetry.py
```diff
@@ -0,0 +1,39 @@
...
+    data = now.strftime("%m/%d/%Y %H:%M:%S")
+    return data
+
+def send_message(args, debug=False):
+    # send in json format
+    data = "{\n\"date\":\"" + get_datetime_str() + "\","
```
simonz130, line 22: you should define your message format using proto
https://developers.google.com/protocol-buffers/docs/pythontutorial
or you can use json

terraform/telemetry.py
```diff
@@ -0,0 +1,39 @@
...
+    
+    # First arg in args is name of script, 'telemetry.py', so we skip it
+    for i in range(len(args) - 1):
+        key_value = args[i + 1].split("=")
+        data += "\n\"" + key_value[0] + "\":\"" + key_value[1]  + "\","
```
simonz130, line 27: This won't work for privacy reasons. You need to obfuscate project_id and in this script you don't treat PII differently from other parameters.

simonz130: Another issue: there is no pre-set schema. This means that users can invoke this script manually, sending a bunch of custom key-value pairs that might be read by the server side of it and produce a lot of fake data into our telemetry pipeline. We need to tighten this script to only accept payloads of a particular schemas that we support.

terraform/destroy.sh
```diff
@@ -28,6 +28,8 @@ acct=$(gcloud info --format="value(config.account)")
 SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
 
+VERSION="v0.2.5"
```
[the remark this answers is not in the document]
eaberman: Not currently. I will add logic to do that!

terraform/install.sh
```diff
@@ -23,6 +23,7 @@ if [[ -n "$DEBUG" ]]; then set -x; fi
 # ensure the working dir is the script's folder
 SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
+VERSION="v0.2.5"
```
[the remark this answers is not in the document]
eaberman: Okay. I will plan to write a test that tests the VERSION value is consistent with the actual version.

terraform/install.sh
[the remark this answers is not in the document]
eaberman: Yes, that's a good point! I will change to `python3 telemetry.py`.

terraform/install.sh
eaberman: I was thinking of storing a separate file with constants, where I could store brief variable names, longer descriptions, and numbers corresponding to events. Do you think this would be a better way to do it, or make it unnecessarily complex?

terraform/install.sh
eaberman: That's a good point. I will do that.

terraform/telemetry.py
eaberman: Currently, I am running this locally with my service account (in which I download and store GOOGLE_APPLICATION_CREDENTIALS). I'm not sure how that will translate when everyone is working on it. I'll look into it.

terraform/telemetry.py
eaberman: Okay. I'll look into obfuscating project id. I'll implement current schema to be:
- date-time
- UUID
- obfuscated project id
- event

terraform/telemetry.py
```diff
@@ -0,0 +1,83 @@
...
+    # When you publish a message, the client returns a future.
+    future = publisher.publish(topic_path, data=data)
+    if (debug): print(future.result())
+
+def main():
+    # send messages
+    # schema: session UUID, project UUID, event, date-time, version
+    session=sys.argv[1]
+    project=get_id_hash(sys.argv[2])
+    event=sys.argv[3]
```
simonz130, line 77: You should add named parameters and input validation. Perhaps, worth trying to use a CLI library like click - similar to  what Tiffany is doing here: https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/422

terraform/destroy.sh
```diff
@@ -28,6 +28,10 @@ acct=$(gcloud info --format="value(config.account)")
 SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
 
+# For the purposes of telemetry
+VERSION="v0.2.5"
+SESSION=$(python3 -c "import telemetry; print(telemetry.get_uuid())")
```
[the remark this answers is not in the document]
eaberman: Ahhh, that's a good idea. I'll work on that! On that note, is it possible to store telemetry service account credentials on Cloud Shell in a way that is not accessible by the user? e.g. telemetry.py can only be called from the script within cloud shell?

terraform/destroy.sh
```diff
@@ -76,9 +80,11 @@ gcloud projects delete $PROJECT_ID
 found=$(gcloud projects list --filter="${PROJECT_ID}" --format="value(projectId)")
 if [[ -n "${found}" ]]; then
     log "project $PROJECT_ID not deleted"
+    python3 telemetry.py $SESSION $PROJECT_ID sandbox-not-destroyed $VERSION
```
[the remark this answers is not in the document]
eaberman: Good point! I'll utilize named params.

terraform/install.sh
```diff
@@ -37,6 +42,7 @@ promptForBillingAccount() {
     log ""
     log "To list active billing accounts, run:"
     log "gcloud beta billing accounts list --filter open=true"
+    python3 telemetry.py $SESSION "No project" no-active-billing $VERSION
```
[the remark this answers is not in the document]
eaberman: In this instance, no project has been created yet, since the project is only created after an active billing account is created. Instead of using "No project", I could alternatively use a constant.

terraform/telemetry.py
eaberman: Okay, good to know! I will look into Tiffany's PR.

terraform/destroy.sh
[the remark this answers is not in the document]
Daniel-Sanche: Can't we add the env var to the container itself? That way we can avoid replacing it in multiple places

terraform/telemetry.py
```diff
@@ -0,0 +1,80 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# File usage:
+# python3 telemetry.py session project event version
```
Daniel-Sanche, line 17: This header makes me think those are all positional arguments, but they're not. Maybe show example usage?

`python3 telemetry.py --session 123  --project my-project --event my-event --version v0.0.0`

terraform/telemetry.py
```diff
@@ -0,0 +1,80 @@
...
+@click.command()
+@click.option('--session', help='Current session (unique across projects).')
+@click.option('--project_id', prompt='Project id', help='Project name in Google Cloud Platform.')
+@click.option('--event', prompt='Event code', help='The  event that occurred.')
+@click.option('--version', default="v0.2.5", prompt='Version of Sandbox', help='Release version of Sandbox.')
+def send_message(session, project_id, event, version):
+    datetime=get_datetime_str()
+    project=get_id_hash(project_id)
+    error, message = validate_args()
+    if (error): print(message)
```
Daniel-Sanche, line 60: It might be better to [silently log to a file](https://docs.python.org/2.3/lib/node304.html) instead of print. But that's pretty low priority, so it might not be worth doing at this point

terraform/telemetry.py
Daniel-Sanche: Instead of embedding a service account as we discussed today, could we just make the topic open to `allAuthenticatedUsers`? That should be essentially the same security, and then we don't have to worry about security reviews, etc

terraform/destroy.sh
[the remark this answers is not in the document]
simonz130: > Can't we add the env var to the container itself? That way we can avoid replacing it in multiple places

I like that! One of the reasons we created custom cloud shell image is to be able to customize it to what we need.

terraform/telemetry.py
simonz130: You might also want to send obfuscated userid, so we can track the # of unique users trying to use it and accomplish different tasks.

terraform/destroy.sh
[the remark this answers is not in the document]
simonz130: I think environment variable is just easier to implement and it helps to track a session of particular cloudshell VM usage for a given user at a given time.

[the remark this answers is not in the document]
Daniel-Sanche: Do we know where to set the session environment variable? Is there a bashrc type file we can modify that'll execute on each run?

terraform/telemetry.py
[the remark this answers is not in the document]
eaberman: I think there's already a shebang used in installation script (`#!/bin/bash`), so I may just have to go with the excess `python3` for now.

terraform/telemetry.py
```diff
@@ -0,0 +1,80 @@
...
+
+def get_id_hash(project_id):
+    m = hashlib.sha256()
+    m.update(project_id.encode('utf-8'))
+    hashed = m.hexdigest()
+    return hashed
+
+@click.command()
+@click.option('--session', help='Current session (unique across projects).')
+@click.option('--project_id', prompt='Project id', help='Project name in Google Cloud Platform.')
```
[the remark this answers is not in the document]
eaberman: Added in the description!

terraform/install.sh
```diff
@@ -37,6 +43,7 @@ promptForBillingAccount() {
     log ""
     log "To list active billing accounts, run:"
     log "gcloud beta billing accounts list --filter open=true"
+    python3 telemetry.py --session=$SESSION --project_id="No project" --event=no-active-billing --version=$VERSION
```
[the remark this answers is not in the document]
eaberman: Changed to `unactivated`

terraform/telemetry.py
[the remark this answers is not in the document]
eaberman: Unfortunately `project-id` is not valid syntax for the python library click.

I use `project` to act as the `project` field in the telemetry message (it is the obfuscated project id).

terraform/telemetry.py
```diff
@@ -0,0 +1,80 @@
...
+    return data
+
+def get_id_hash(project_id):
+    m = hashlib.sha256()
+    m.update(project_id.encode('utf-8'))
+    hashed = m.hexdigest()
+    return hashed
+
+@click.command()
+@click.option('--session', help='Current session (unique across projects).')
```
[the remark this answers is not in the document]
eaberman: I set it independently within the `./install.sh` script and `./destroy.sh` script. It should be unique per when the user runs the script. e.g. if a user runs the script twice, the telemetry system can record that the actions occurred on two separate "sessions."

eaberman:
This PR is ready for review and potentially to be merged. If I have time on Monday, I plan to add logging for argument validation in python script, and make sure the `VERSION` variable is updated in the CI and release script.

cloud-shell/Dockerfile
```diff
@@ -20,5 +20,13 @@ RUN sudo apt-get install unzip && \
   unzip -o terraform.zip && \
   sudo install terraform /usr/local/bin
 
+# install python libraries
+RUN python3 -m pip install google-cloud-pubsub
+RUN python3 -m pip install click
+RUN python3 -m pip install google-cloud-monitoring
+
+# set env var
+RUN echo "VERSION=v0.2.5" >> /etc/environment
```
simonz130, line 29: you probably want to set the version to "latest" or something else that means "master branch". Then in the release script it will be updated to whatever tag we use to make a release.

terraform/destroy.sh
```diff
@@ -28,6 +28,14 @@ acct=$(gcloud info --format="value(config.account)")
 SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
 
+# For the purposes of telemetry
+# session is defined as the current time the user runs destroy.sh file
```
simonz130, line 32: don't we record time too?

terraform/telemetry.py
```diff
@@ -0,0 +1,66 @@
...
+
+import sys
+import json
+import hashlib
+import click
+import time
+
+from google.cloud import pubsub_v1
+
+def obfuscate_project_id(project_id):
```
simonz130, line 26: nit: add a comment above to explain why do we need this method

simonz130 (approved):

terraform/telemetry.py
eaberman: Added!

terraform/destroy.sh
eaberman: We do; I meant current instance. I will clarify that in my documentation.

cloud-shell/Dockerfile
eaberman: Okay, I see. I will look into doing this.

terraform/install.sh
```diff
@@ -24,8 +24,15 @@ if [[ -n "$DEBUG" ]]; then set -x; fi
 SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
 
+# create variables for telemetry purposes
+SESSION=$(python3 -c "import uuid; print(uuid.uuid4())")
+
 log() { echo "$1" >&2; }
 
+sendTelemetry() {
```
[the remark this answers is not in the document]
eaberman: I added some documentation!

cloud-shell/Dockerfile
Daniel-Sanche: This is for metrics purposes, right? Do we want "latest" to show up in our metrics? Or just the last released version?

Either way, we'll have to update the make-release script to make sure this value is updated

make-release.sh
```diff
@@ -39,6 +39,9 @@ sed -i -e "s/cloudshell-image:v\([0-9\.]\+\)/cloudshell-image:${NEW_VERSION}/g"
 sed -i -e "s/cloudshell_git_branch=v\([0-9\.]\+\)/cloudshell_git_branch=${NEW_VERSION}/g" ${REPO_ROOT}/docs/index.html;
 sed -i -e "s/cloudshell-image:v\([0-9\.]\+\)/cloudshell-image:${NEW_VERSION}/g" ${REPO_ROOT}/docs/index.html;
 
+# update custom Cloud Shell image variable
+sed -i -e "s/VERSION=v\([0-9\.]\+\)/VERSION=${NEW_VERSION}/g" ${REPO_ROOT}/cloud-shell/Dockerfile;
```
Daniel-Sanche, line 43: You'll also have to do a `git add` for this file later to make sure it gets in the release commit below

terraform/destroy.sh
```diff
@@ -28,6 +28,16 @@ acct=$(gcloud info --format="value(config.account)")
...
+SESSION=$(python3 -c "import uuid; print(uuid.uuid4())")
+
+# this function sends de-identified information to the Google Cloud Platform database
+# on what events occur in users' Sandbox projects for development purposes
+sendTelemetry() {
+  python3 telemetry.py --session=$SESSION --project_id=$1 --event=$2 --version=$VERSION
+}
```
Daniel-Sanche, line 39: It seems a little strange to re-define telemetry.py as sendTelemetry. Why not just rename the file to sendTelemetry if you prefer this?

terraform/telemetry.py
```diff
@@ -0,0 +1,99 @@
...
+        not isinstance(project_id, str) or not isinstance(session, str)):
+        return False, "An argument passed to telemetry was an invalid type."
+    
+    # check format
+    if (re.fullmatch(args_exp["version"], version) is None or
+        re.fullmatch(args_exp["project_id"], project_id) is None or
+        re.fullmatch(args_exp["session"], session) is None):
+        return False, "An argument passed to telemetry was an invalid format."
+    
+    return True, ""
```
Daniel-Sanche, line 70: This pattern isn't really typical for Python. Exceptions feel like the natural solution for what you're trying to do. You could also just return a nullable error message to remove the redundant variable, but that's not quite as clean

terraform/telemetry.py
```diff
@@ -0,0 +1,99 @@
...
+        print("Failed to send telemetry.")
+        print(err_msg)
+        return
+    
+    # send data as a bytestring
+    msg = get_telemetry_msg(session, project_id, event, version)
+    bytes = msg.encode("utf-8")
+    
+    # connect to pubsub and send message
+    project_id = "stackdriver-sandbox-230822"
```
Daniel-Sanche, line 90: nit: I wonder if it would also be a good idea to save this as an environment variable in the Cloud Shell instead of hardcoding it (will it be needed in other places?)

terraform/destroy.sh
```diff
@@ -28,6 +28,16 @@ acct=$(gcloud info --format="value(config.account)")
 SCRIPT_DIR=$(realpath $(dirname "$0"))
 cd $SCRIPT_DIR
 
+# For the purposes of telemetry
+# session is defined as the current time the user runs destroy.sh file
+SESSION=$(python3 -c "import uuid; print(uuid.uuid4())")
```
Daniel-Sanche, line 33: Are you sure you always want the session to be unique for each script run? You could try to save the session id as an environment variable if you want it to span multiple scripts

terraform/destroy.sh
eaberman: I don't necessarily mean to re-define telemetry.py here: I'm actually trying to reduce redundancies with sending the same `SESSION` and `VERSION` variables in every instance telemetry is sent.

terraform/telemetry.py
eaberman: Okay, I'll look into throwing errors here.

terraform/telemetry.py
eaberman: I don't think it is currently needed in other places. Do you think it would be useful to have it there for the future?

terraform/destroy.sh
eaberman: Yes, actually that sounds like a much better approach to what I want to do! I will try to do that.

Daniel-Sanche (changes requested):
The only change that is completely necessary is the make-release change so that the value is updated on each release. Other than that you can treat my comments as suggestions

Daniel-Sanche (approved):
LGTM

PR #422, opened 2020-08-31
added CLI for sre recipes

tishen25:


sre-recipes/sre_recipes.py
```diff
@@ -0,0 +1,83 @@
...
+import click
+
+dir_abs_path = os.path.dirname(os.path.abspath(__file__))
+manifest_abs_path = dir_abs_path + '/../kubernetes-manifests/'
+
+# Returns the absolute path of the working service's kubernetes manifest
+# and the broken service's kubernetes manifest
+def get_paths(service):
+    working_service_abs_path = manifest_abs_path + service + ".yaml"
+    broken_service_abs_path = manifest_abs_path + service + "srerecipes.yaml"
```
simonz130: can we store sre recipes each it its own folder for better isolation and ease of management?
So, for example:
sre-recipes folder at the root level, then subfolder for each recipe and inside all relevant content?

sre-recipes/sre_recipes.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+# Rolls back the working version of the given service and deploys the
+# broken version of the given service
+def break_func(service):
+    working_service_abs_path, broken_service_abs_path = get_paths(service)
+
+    print("Rolling back working " + service + " service...")
+    delete_command = 'kubectl delete -f ' + working_service_abs_path
+    output, error = run_command(delete_command)
```
simonz130, line 66: What if it failed to delete?

sre-recipes/sre_recipes.py
```diff
@@ -0,0 +1,83 @@
...
+
+    print("Deploying working " + service + " service...")
+    apply_command = 'kubectl apply -f ' + working_service_abs_path
+    output, error = run_command(apply_command)
+
+
+@click.command()
+@click.argument('action', type=click.Choice(['break', 'restore', 'verify']))
+@click.argument('service', type=click.Choice(['frontend']))
```
simonz130: The command line should have minimal required number of arguments.
The user will apply SRE recipe setup by saying one of:
**break, restore, verify**
The fact that you update the frontend (or other microservice) should be internal implementation detail of the Sandbox and corresponding sre recipe and user should not be asked to provide microservice name.

sre-recipes/sre_recipes.py
```diff
@@ -0,0 +1,83 @@
...
+def main(action, service):
+    if action == 'break':
+        break_func(service)
+    elif action == 'restore':
+        restore(service)
+    elif action == 'verify':
+        print('verify!')
+
+if __name__ == "__main__":
+    main()
```
simonz130, line 110: could you rename this tool from **sre_recipes** to **sandboxctl** or **sbctl**?

tishen25: sure!

sre-recipes/sre_recipes.py
tishen25: Yep! Changing the design to have each recipe be contained in a folder.

sre-recipes/sre_recipes.py
tishen25: Changed design so that the user will choose a recipe, not a service

sre-recipes/sre_recipes.py
tishen25: Good question! From what I've seen, the delete fails if the service that we want to remove isn't there. In this case, I think it's fine to do nothing because the point is to remove the service anyway. I'm not quite sure about the case where the delete fails because kubernetes actually failed to delete. I'll some investigating on this after I put in changes for the CLI restructure.

sre-recipes/sre_recipes.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+# Returns the absolute path of the working service's kubernetes manifest
+def get_working_path(service):
+    return get_relative_path('/../kubernetes-manifests/', service, '.yaml')
+
+
+# Returns the absolute path of the broken service's kubernetes manifest
+def get_broken_path(service):
+    return get_relative_path('/kubernetes-manifests/', service, '.yaml')
```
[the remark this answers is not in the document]
tishen25: I think the newest version is like that? Let me know if it isn't, and I'll take a look!

sre-recipes/sandboxctl.py
```diff
@@ -0,0 +1,75 @@
...
+    Gets all valid recipes in sre-recipes/recipes, and returns
+    the recipe names
+    """
+    recipes = {}
+    recipes_root_dir = os.path.dirname(os.path.abspath(__file__)) + '/recipes'
+    recipes_root_modules_path = Path(recipes_root_dir).resolve()
+    for (_, recipe_module_name, _) in iter_modules([recipes_root_modules_path]):
+        recipe_module_path = Path(recipes_root_dir + '/' + recipe_module_name).resolve()
+        for (_, recipe_name, _) in iter_modules([recipe_module_path]):
+            recipe_module = import_module('.'.join(['recipes', recipe_module_name, recipe_name]))
```
Daniel-Sanche, line 45: I find this file parsing code above this hard to read. You should try to avoid changing between paths and strings so often, and make use of [Path's built-in `/` operator](https://stackoverflow.com/questions/48190959/how-do-i-append-a-string-to-a-path-in-python/48191073). You should also consider using something like `os.walk` to iterate files more cleanly

Many-nested loops like this aren't very pythonic in general though. You could replace 40-45 with a list comprehension:
```
from os.path import dirname, abspath

root_dir = Path(dirname(abspath(__file__))) / 'recipes'
modules = [f"recipes.{subdir}.{name}" 
                for subdir in os.listdir(root_dir)
                for _, name, _ in iter_modules([root_dir / subdir])]
```
That should give you a list of strings you can feed into `import_module`

sre-recipes/sandboxctl.py
```diff
@@ -0,0 +1,75 @@
...
+            for attribute_name in dir(recipe_module):
+                attribute = getattr(recipe_module, attribute_name)
+                if isclass(attribute) and attribute is not Recipe \
```
Daniel-Sanche, line 48: This section could use some commenting

sre-recipes/recipes/recipe_1/recipe_1.py
```diff
@@ -0,0 +1,112 @@
...
+    introduces latency into the frontend service.
+    """
+
+    def __init__(self):
+        self.__service = 'frontend'
+
+    @staticmethod
+    def get_relative_path(relative_path, filename, extension):
+        """
+        Returns the absolute path to the filename given the relative
```
briancorwin, line 37: It's kind of confusing to call it "get_relative_path" when it's returning the absolute path.

sre-recipes/recipes/recipe_1/recipe_1.py
```diff
@@ -0,0 +1,112 @@
...
+        path to the file, filename, and file extension
+        """
+        dir_abs_path = os.path.dirname(os.path.abspath(__file__))
+        return dir_abs_path + relative_path + filename + extension
+
+    @staticmethod
+    def get_working_path(service):
```
briancorwin, line 44: If you want a method to be "private" you can start it with an underscore, like _get_working_path

sre-recipes/recipes/recipe_1/recipe_1.py
```diff
@@ -0,0 +1,112 @@
...
+"""
+This module contains the implementation of recipe 1
+"""
+
+import os
+import subprocess
+from recipe import Recipe
+
+
+class Recipe1(Recipe):
```
briancorwin, line 25: Why is this Recipe1 and extending the Recipe base class?  Is this to support the first approach to SRE recipes?  With the broken and fixed builds being swapped and "Recipe2" would be another approach?

If so I think we need a more descriptive class name than Recipe1.  Also I might consider making this an ABC or just a class that is invoked with a different service param each time to vary the service used.

sre-recipes/sandboxctl.py
```diff
@@ -0,0 +1,75 @@
...
+                    globals()[attribute_name] = attribute
+                    try:
+                        recipe_obj = attribute()
+                        recipes[attribute_name] = recipe_obj
+                    except TypeError:
+                        pass
+    return recipes
+
+
+RECIPES = get_recipes()
```
briancorwin, line 59: I'm not aware of what "click" does?  Could you provide a comment or a README giving an idea of what the CLI will look like?

sre-recipes/sandboxctl.py
briancorwin: +1

In basically any language if you get past 2 or 3 levels of nesting that's a code smell that you might want to use a different approach (like list comprehensions here) or at least break things up into separate functions so the logic is clearer.

sre-recipes/sre_recipes.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+# Rolls back the broken version of the given service and deploys the
+# working version of the given service
+def restore(recipe):
+    service = recipe_service_mapping[recipe]
+    working_service_abs_path = get_working_path(service)
+    broken_service_abs_path = get_broken_path(service)
+
+    print("Rolling back broken service...")
```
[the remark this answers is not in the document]
tishen25: good idea!

sre-recipes/recipes/recipe_1/recipe_1.py
```diff
@@ -0,0 +1,112 @@
...
+        self.__service = 'frontend'
+
+    @staticmethod
+    def get_relative_path(relative_path, filename, extension):
+        """
+        Returns the absolute path to the filename given the relative
+        path to the file, filename, and file extension
+        """
+        dir_abs_path = os.path.dirname(os.path.abspath(__file__))
+        return dir_abs_path + relative_path + filename + extension
```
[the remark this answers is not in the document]
tishen25: `dir_abs_path` is unique to the recipe, but the rest of the args aren't. Maybe I could make it generic by adding `dir_abs_path` as an arg, then putting it in the abstract class?

sre-recipes/recipes/recipe_1/recipe_1.py
```diff
@@ -0,0 +1,112 @@
...
+        broken_service_abs_path = Recipe1.get_broken_path(self.__service)
+        recipe_config_map_abs_path = Recipe1.get_config_map_path(config_map)
+
+        print("Rolling back working service...")
+        delete_command = 'kubectl delete -f ' + working_service_abs_path
+        Recipe1.run_command(delete_command)
+
+        print("Setting up broken service...")
+        create_config_map_command = 'kubectl create configmap ' + config_map_name \
+            + ' --from-file=' + recipe_config_map_abs_path
```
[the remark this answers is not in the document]
tishen25: Thanks for letting me know! I don't think we ended up needing config maps, so this code will be removed for the next iteration, but it's good to know the convention regardless!

sre-recipes/sandboxctl.py
```diff
@@ -0,0 +1,75 @@
...
+# limitations under the License.
+
+# -*- coding: utf-8 -*-
+"""
+This CLI is the interface for sre recipes. It deploys and rolls back
+broken services as well as verifies that the user found the correct
+cause of the broken service.
+
+For information on how to run the CLI, run the following:
+`python3 sandboxctl.py --help`
```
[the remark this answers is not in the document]
tishen25: Wow, that's neat! I didn't know you could do that

sre-recipes/recipes/recipe_1/recipe_1.py
Daniel-Sanche, line 44: These names are hard to parse. Shouldn't this be something like `get_manifest_path`?

sre-recipes/recipes/recipe_1/recipe_1.py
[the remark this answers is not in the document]
Daniel-Sanche: Maybe, but don't worry about it too much if most of the logic is recipe-specific

sre-recipes/sre_recipes.py
tishen25: I think this will no longer be an issue because we aren't going to use different images to switch between working and broken versions anymore.

sre-recipes/sandboxctl.py
tishen25: Good point - I ended up factoring this into a separate method + used list comprehensions as much as possible.

sre-recipes/recipes/recipe_1/recipe_1.py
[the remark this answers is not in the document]
tishen25: I think this will no longer be an issue because we aren't going to use different images to switch between working and broken versions anymore, which means we won't need to be searching for different images!

sre-recipes/sandboxctl.py
tishen25: Thanks! I ended up converting as much as possible into list comprehensions.

sre-recipes/recipes/recipe_1/recipe_1.py
tishen25: I think this will no longer be an issue because we aren't going to use different images to switch between working and broken versions anymore, which means we won't need to be searching for different images + using these methods to find paths!

sre-recipes/recipes/recipe_1/recipe_1.py
tishen25: Cool, good to know for future reference! As mentioned in previous comments, I don't think I'll need the pathfinding methods anymore.

tishen25: Good to know for future reference! As mentioned in previous comments, I don't think I'll need the pathfinding methods anymore.

sre-recipes/recipes/recipe_1/recipe_1.py
tishen25: Yeah, Recipes is an ABC that defines methods/behavior that each recipe must have. Any recipe would need to extend ABC in order to be considered a recipe. The goal is that each recipe is contained within a folder so that we have as little to maintain as possible. It also makes it easier for people adding recipes to have flexibility over the implementation of breaking and restoring services. Simon's goal is to have the implementation of breaking and restoring be the responsibility of the author of the recipe instead of the CLI tool. This means that we probably don't want to make Recipe1 an ABC that is invoked with a different service param because we don't want that as our responsibility.

As for the naming, I can rename the class to something like ConvertCurrencies instead.

sre-recipes/sandboxctl.py
tishen25: Click is just the library I'm using to make the CLI. I can provide a README to explain what the CLI does/looks like.

sre-recipes/recipes/recipe_1/currencies_recipe.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        print("Deploying working service...")
+        self._deploy_state(False)
+        logging.info('Deployed working service')
+
+    @staticmethod
+    def _get_user_input():
+        """Prompts users for an answer"""
+        answer = input('Your answer: ')
+        return answer
```
Daniel-Sanche, line 80: nit: I'm not sure this needs to be a function, since it's a single line

sre-recipes/recipes/recipe_1/currencies_recipe.py
```diff
@@ -0,0 +1,138 @@
...
+        print('\t[b] cart service')
+        print('\t[c] checkout service')
+        print('\t[d] currency service')
+        print('\t[e] email service')
+        print('\t[f] frontend service')
+        print('\t[g] payment service')
+        print('\t[h] product catalog service')
+        print('\t[i] recommendation service')
+        print('\t[j] redis')
+        print('\t[k] shipping service')
```
Daniel-Sanche, line 98: nit: This would probably look better as printing a [multiline string](https://stackoverflow.com/questions/10660435/pythonic-way-to-create-a-long-multi-line-string)

sre-recipes/recipes/recipe_1/currencies_recipe.py
```diff
@@ -0,0 +1,138 @@
...
+        answer = self._service_multiple_choice()
+        while answer not in ('f', 'F'):
+            print('Incorrect. Please try again.')
+            answer = self._get_user_input()
+        print('Correct! The frontend service is broken.')
+
+    def _verify_broken_cause(self):
+        """Verifies the user found the root cause of the breakage"""
+        answer = self._cause_multiple_choice()
+        while answer not in ('c', 'C'):
```
Daniel-Sanche, line 126: nit: `answer.lower() != 'c'` is a bit more direct

sre-recipes/recipes/recipe_1/currencies_recipe.py
```diff
@@ -0,0 +1,138 @@
...
+            print('Incorrect. Please try again.')
+            answer = self._get_user_input()
+        print('Correct! High latency caused the breakage.')
+
+    def verify(self):
+        """Verifies the user found the root cause of the broken service"""
+        print('This is a multiple choice quiz to verify that you\'ve')
```
Daniel-Sanche, line 133: nit: using double quotes here would avoid the `\`

sre-recipes/README.md
```diff
@@ -0,0 +1,27 @@
+# SRE Recipes
+
+SRE Recipes is a tool to help users familiarize themselves with finding the root cause of a breakage using monitoring.
+
+## Usage
+
+To view which recipes exist, run the command below in this directory:
+
+```
+$ python3 sandboxctl --help
```
Daniel-Sanche, line 10: You should be able to run sandboxctl directly instead of with python3 since you added the shebang, right?

sre-recipes/sandboxctl
```diff
@@ -0,0 +1,101 @@
...
+
+def get_valid_recipes(attributes):
+    """
+    Gets all valid recipes from the given list of tuples of attribute names
+    and attributes, and returns a map containing recipe names and recipe objects
+    """
+
+    recipe_objs = {}
+    recipe_num = 1
+    for attribute_name, attribute in attributes:
```
Daniel-Sanche, line 45: instead of managing `recipe_num` manually, you could do:

`for recipe_num, (attribute_name, attribute) in enumerate(attributes):`

That gives you all three as part of the loop

sre-recipes/sandboxctl
```diff
@@ -0,0 +1,101 @@
...
+        if isclass(attribute) and attribute is not Recipe and issubclass(attribute, Recipe):
+            globals()[attribute_name] = attribute
```
Daniel-Sanche, line 47: This could use a comment. I'm not exactly sure if I know what's going on here

Daniel-Sanche (approved):
I left some style comments that you might want to consider, but overall looks great!

sre-recipes/recipes/recipe_1/currencies_recipe.py
tishen25: fixed!

sre-recipes/recipes/recipe_1/currencies_recipe.py
tishen25: fixed!

sre-recipes/recipes/recipe_1/currencies_recipe.py
tishen25: fixed!

sre-recipes/recipes/recipe_1/currencies_recipe.py
tishen25: fixed!

sre-recipes/README.md
tishen25: Maybe I'm doing something wrong, but I get this error when running `sandboxctl --help` (for example):
```
bash: sandboxctl: command not found
```

sre-recipes/sandboxctl
tishen25: I didn't know that, thanks!

sre-recipes/sandboxctl
tishen25: I realized I actually don't need that line of code. I think if you want to reference the class name, you need to set the global variable, but since I'm storing an instance of each recipe in a map, I don't need to set it globally.

sre-recipes/README.md
Daniel-Sanche: Hmm does it work when you run it with `./sandboxctl --help`? I guess the problem may be that it's not on the path, so you have to call it more directly. We could pretty easily change the cloud shell image to make that directory part of the path, but that may not be worth it. Calling it with python is fine too!

tishen25: Ah, it works with `./sandboxctl --help`! I will update the README

Issue #428, opened 2020-09-01
Manage Load Generator provisioning using terraform


gzhao408:
As part of our larger initiative to use terraform for all our provisioning

Issue #427, opened 2020-09-01
Use GKE for load generator infrastructure


gzhao408:
As per the design decisions made, we would like to switch from GCE to GKE for more flexibility and scalability.

PR #429, opened 2020-09-01
Loadgen: containerize, use terraform to provision and deploy on GKE, CI for image pushing, e2e tests

### LINKED ISSUE #404: loadgen image does not update on push to master and new releases

gzhao408:
Resolves #404 #427 #428 

-[x] Containerize the load generator
-[x] Provision GKE cluster and deploy load generator using Terraform
-[x] CI workflow: tag and push loadgen docker image on push to master and new release
-[x] Modify/add e2e testing for the load generator
-[x] Remove the old load generator
-[x] Alias kubectl contexts to make it easier fo the user to see what's going on


gzhao408:
[cloud shell link for testing this branch](https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/gzhao408/stackdriver-sandbox.git&cloudshell_git_branch=loadgen-gke&cloudshell_working_dir=terraform&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image:v0.2.5 )

kubernetes-config/locust-master-controller.yaml
```diff
@@ -0,0 +1,52 @@
...
+    matchLabels:
+      app: locust-master
+  template:
+    metadata:
+      labels:
+        app: locust-master
+    spec:
+      containers:
+        - name: locust-master
+          image: gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator:wip
```
simonz130, line 34: why this tag?

kubernetes-config/locust-master-controller.yaml
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+apiVersion: "extensions/v1beta1"
+kind: "Deployment"
+metadata:
+  name: locust-master
```
simonz130, line 19: can we rename locust-master to locust-main or locust-lead or something along these lines? (more inclusive naming)

terraform/install.sh
```diff
@@ -198,6 +198,8 @@ applyTerraform() {
 authenticateCluster() {
   CLUSTER_ZONE=$(gcloud container clusters list --filter="name:cloud-ops-sandbox" --project $project_id --format="value(zone)")
   gcloud container clusters get-credentials cloud-ops-sandbox --zone "$CLUSTER_ZONE"
+  # Make alias for this kubectl context
+  kubectx main=.
```
simonz130, line 202: why do we need this? We don't bundle kubectx into our custom image and I think it's unnecessary dependency

terraform/install.sh
```diff
@@ -256,28 +258,30 @@ getExternalIp() {
...
-  old_loadgen=$(gcloud compute instances list --project $project_id --filter="name:loadgenerator*" --format="value(name)")
-  if [[ -n "$old_loadgen" ]]; then
-    loadgen_zone=$(gcloud compute instances list --project $project_id --filter="name:loadgenerator*" --format="value(zone)")
-    gcloud compute instances delete $old_loadgen --zone $loadgen_zone --quiet
-  fi
   # launch a new load generator
-  ../loadgenerator/loadgen autostart $external_ip
+  pushd loadgen/
+  terraform init -lock=false
+  terraform apply --auto-approve -var="project_id=${project_id}" -var="external_ip=${external_ip}"
```
simonz130, line 264: loadgen provisioning should be part of terraform provisioning that includes the rest of the cluster. It should not be invoked separately as this way we'll have a dedicated state (.tfstate) and it will hurt our ability to resume cluster creation in case it didn't work for the first time.

terraform/install.sh
```diff
@@ -256,28 +258,30 @@ getExternalIp() {
...
+  popd
+
   # find the IP of the load generator web interface
   TRIES=0
-  while [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip:8080" -o /dev/null --max-time 1) -ne 200  && \
+  while [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip:8089" -o /dev/null --max-time 1) -ne 200  && \
```
simonz130, line 269: This also can be done via terraform

terraform/install.sh
```diff
@@ -256,28 +258,30 @@ getExternalIp() {
...
   done
-  if [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip:8080" -o /dev/null  --max-time 1) -ne 200 ]]; then
+  if [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip:8089" -o /dev/null  --max-time 1) -ne 200 ]]; then
     log "error: load generator unreachable"
   fi
+
+  # Make kubectx alias for this kubectl context
+  kubectx loadgenerator=.
+  # Return kubectl context to the main cluster and show this to the user
+  log $(kubectx main)
```
simonz130, line 284: These should be removed

terraform/loadgen/00_loadgen.tf
```diff
@@ -0,0 +1,150 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Let's create the GKE cluster! This one's pretty complicated so buckle up.
+
+# This is another example of the random provider. Here we're using it to pick a
+# zone in us-west1 at random - a different region from where we provisioned
+# the Hipster Shop.
+resource "random_shuffle" "zone" {
+  input = ["us-west1-a", "us-west1-b", "us-west1-c"]
```
simonz130, line 21: add other locations that are not us-west and are not locations for creation of the demo app cluster

terraform/loadgen/00_loadgen.tf
```diff
@@ -0,0 +1,150 @@
...
+
+    management {
+      auto_repair  = true
+      auto_upgrade = true
+    }
+  }
+
+  # Stores the zone of created loadgenerator gke cluster
+  provisioner "local-exec" {
+    command = "gcloud config set compute/zone ${element(random_shuffle.zone.result, 0)}"
```
simonz130, line 98: once you integrate this with main terraform script, you might need to change this so the zone is appended for each gcloud command that needs it as opposed to setting it up globally like this

terraform/loadgen/00_loadgen.tf
```diff
@@ -0,0 +1,150 @@
...
+resource "null_resource" "current_project" {
+  provisioner "local-exec" {
+    command = "gcloud config set project ${var.project_id}"
+  }
+}
+
+# Setting kubectl context to currently deployed loadgenerator GKE cluster
+resource "null_resource" "set_gke_context" {
+  provisioner "local-exec" {
+    command = "gcloud container clusters get-credentials cloud-ops-sandbox-loadgenerator --zone ${element(random_shuffle.zone.result, 0)} --project ${var.project_id}"
```
simonz130, line 113: use variable substitution that evaluates to cloud-ops-sandbox-loadgenerator (instead of using a literal constant)

tests/provisioning/loadgen_test.py
```diff
@@ -0,0 +1,69 @@
...
+        cluster_info = client.get_cluster(name=TestLoadGenerator.name)
+        node_count = cluster_info.current_node_count
+        self.assertTrue(node_count == 3)
+
+    def testReachOfLoadgen(self):
+        """ Test if querying load generator returns 200 """
+        command = ("kubectl get service locust-master -o jsonpath='{.status.loadBalancer.ingress[0].ip}'")
+        result = subprocess.run(split(command), encoding='utf-8', capture_output=True)
+        loadgen_ip = result.stdout.replace('\n', '')
+        url = 'http://{0}:8089'.format(loadgen_ip)
```
simonz130, line 58: extract port into const

tests/provisioning/loadgen_test.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertTrue(urllib.request.urlopen(url).getcode() == 200)
+
+    def testDifferentZone(self):
+        """ Test if load generator is in a different zone from the GKE cluster """
```
simonz130, line 62: ...from the demo app cluster

kubernetes-config/locust-master-controller.yaml
gzhao408: just mentioned in the meeting - we'll have it just always use the latest release tag (working with Dan to make this happen)

terraform/install.sh
gzhao408: We need kubectl context to match the cluster (the hipstershop and loadgen are on different clusters). `kubectx` is a pretty easy way to do it (and aliasing it for the user makes a better user exp in my opinion) and @Daniel-Sanche told me the dependency is included in the cloud shell

terraform/install.sh
gzhao408: i disagree, I think it's best to return the kubectl context to the main one because it's what a user would be more likely to expect. they shouldn't need to really look at what's going on with the load generator.

terraform/loadgen/00_loadgen.tf
gzhao408: afaik the demo app cluster only uses `us-central1`, and loadgen cluster only uses `us-west1`... is it necessary to have a lot of possible regions? as long as they're in different geographic areas?

terraform/loadgen/00_loadgen.tf
gzhao408: it's not variable though, we name it in the terraform script? could put it in a constant i guess, but we do the exact same for cloud-ops-sandbox above

terraform/install.sh
gzhao408: will work on this - the clusters can be provisioned at the same time, but it'll require a bit of refactoring

kubernetes-config/locust-master-controller.yaml
gzhao408: sure, will do. potentially more problematic is the Web UI displays "number of slaves" so i'm not super sure how to deal with that

gzhao408: done

tests/provisioning/loadgen_test.py
gzhao408: done

terraform/install.sh
Daniel-Sanche: Yeah, kubectx is baked into the default Cloud Shell image. I think we should make use of it so we can switch between clusters without needing to find the zone each time

simonz130: Surprised it's included in base cloud shell image, but ok.

terraform/loadgen/00_loadgen.tf
simonz130: yeah, different geo-locations is important. Having it be based not only in US is a secondary but might benefit folks too

.github/workflows/push-master.yml
```diff
@@ -36,3 +36,18 @@ jobs:
...
       env:
         PROJECT_ID: ${{ secrets.PROJECT_ID }}
+    - name: Push Load Generator Image to GCR
+      timeout-minutes: 20
+      run: |
+        set -x
+        gcloud auth configure-docker --quiet
+        # build and tag with git hash
+        IMAGE_SHA=gcr.io/$PROJECT_ID/sandbox/loadgenerator:$GITHUB_SHA
+        docker build -t $IMAGE_SHA ./src/loadgenerator
```
simonz130, line 46: Should we have a different path here to reflect new loadgen changes compared to previous image?

terraform/install.sh
```diff
@@ -256,28 +258,30 @@ getExternalIp() {
...
-    loadgen_ip=$(gcloud compute instances list --project "$project_id" \
-                                               --filter="name:loadgenerator*" \
-                                               --format="value(networkInterfaces[0].accessConfigs.natIP)")
+    sleep 10
+    loadgen_ip=$(kubectl get service locust-main -o jsonpath='{.status.loadBalancer.ingress[0].ip}');
+     [ -z "$loadgen_ip" ] && sleep 10;
     TRIES=$((TRIES + 1))
   done
-  if [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip:8080" -o /dev/null  --max-time 1) -ne 200 ]]; then
+  if [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip:$LOCUST_PORT" -o /dev/null  --max-time 1) -ne 200 ]]; then
```
Daniel-Sanche, line 277: nit: I don't like seeing the query duplicated here. It would be great if you could find a way to use the result from above. But I know your time is limited, so don't worry too much about this

terraform/install.sh
```diff
@@ -256,28 +258,30 @@ getExternalIp() {
...
+  while [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip:$LOCUST_PORT" -o /dev/null --max-time 1) -ne 200  && \
       "${TRIES}" -lt 20  ]]; do
     log "waiting for load generator instance..."
-    sleep 1
-    loadgen_ip=$(gcloud compute instances list --project "$project_id" \
-                                               --filter="name:loadgenerator*" \
-                                               --format="value(networkInterfaces[0].accessConfigs.natIP)")
+    sleep 10
+    loadgen_ip=$(kubectl get service locust-main -o jsonpath='{.status.loadBalancer.ingress[0].ip}');
+     [ -z "$loadgen_ip" ] && sleep 10;
```
Daniel-Sanche, line 274: is sleeping both before and after the `kubectl` line intentional?

terraform/install.sh
simonz130: Given that your change is already pretty big, I suggest splitting it into a separate CR after this one is merged.

terraform/install.sh
Daniel-Sanche: Yeah, I also think we should return the user to the main cluster after provisioning the load generator

I'm not so sure about putting it in a log statement though, it would probably be better to do it as two lines to separate the logging from the functionality

terraform/loadgen/00_loadgen.tf
```diff
@@ -0,0 +1,150 @@
...
+  #
+  # For this particular case we're not going to be modifying the node pool once
+  # it's deployed, so it makes sense to accept the tradeoff for the convenience
+  # of having it inline.
+  #
+  # Many of the paramaters below are self-explanatory so I'll only call out
+  # interesting things.
+  node_pool {
+    node_config {
+      machine_type = "n1-standard-2"
```
Daniel-Sanche, line 71: I'm a bit unsure about needing 3 n1-standard-2 nodes. That could get expensive if customers accidently leave them live, and we really shouldn't need too many resources for a simple load generator

terraform/install.sh
simonz130: Good point Gloria, agree - let's keep it then.

terraform/install.sh
Daniel-Sanche: How much work/time would be involved in making the change?

Daniel-Sanche (changes requested):

terraform/install.sh
gzhao408: i can just put the cluster-provisioning in the same folder as the hipster shop stuff - that is pretty straightforward and achieves our main goal of having those two tasks run concurrently (saving a lot of time). Moving the rest of the load generator stuff over might get a little complicated.

terraform/install.sh
gzhao408: yeah - i think i'd run `kubectx main` and then log `kubectx` to print the current context?

terraform/loadgen/00_loadgen.tf
gzhao408: yeah I could do that! wasn't sure if it'd be restricted or something, but i can add a few other regions

tests/provisioning/loadgen_test.py
```diff
@@ -0,0 +1,69 @@
...
+class TestLoadGenerator(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        """ Get the location of GKE cluster for later queries """
+        cls.name = 'projects/{0}/locations/{1}/clusters/cloud-ops-sandbox-loadgenerator'.format(getProjectId(), getClusterZone())
+        # authenticate container cluster
+        command=('gcloud config set project {0}'.format(getProjectId()))
+        subprocess.run(split(command))
+        # set kubectl context to cloud-ops-sandbox-loadgenerator
+        command=('gcloud container clusters get-credentials cloud-ops-sandbox-loadgenerator --zone {0}'.format(getClusterZone()))
```
simonz130, line 36: This might be problem when your run e2e tests of various components involving kubernetes on the same CI agent.
For example, say you run loadgen e2e test wired against gke1 and hipstershop e2e test wired to gke2.
If you switch the context, it's switches it for all the sessions for the same user on the same VM/container, so by switching to gke1, you might affect tests that run in parallel and expect to run against gke2.
To prevent that, in your tests, you should specify _--context_ with every kubectl command.

tests/provisioning/provision_test.py
```diff
@@ -30,6 +30,12 @@ class TestGKECluster(unittest.TestCase):
     def setUpClass(cls):
         """ Get the location of GKE cluster for later queries """
         cls.name = 'projects/{0}/locations/{1}/clusters/cloud-ops-sandbox'.format(getProjectId(), getClusterZone())
+        # authenticate container cluster
+        command=('gcloud config set project {0}'.format(getProjectId()))
+        subprocess.run(split(command))
+        # set kubectl context
+        command=('gcloud container clusters get-credentials cloud-ops-sandbox --zone {0}'.format(getClusterZone()))
```
simonz130, line 37: see a similar comment in loadgen tests

simonz130 (approved):

.github/workflows/push-master.yml
```diff
@@ -36,3 +36,18 @@ jobs:
...
+        docker push $IMAGE_SHA
+        # build and tag with latest
+        IMAGE_LATEST=gcr.io/$PROJECT_ID/sandbox/loadgenerator:latest
```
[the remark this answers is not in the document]
gzhao408: just pushed a change, renamed to [gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator/basic:latest](https://pantheon.corp.google.com/gcr/images/stackdriver-sandbox-230822/GLOBAL/sandbox/loadgenerator/basic?gcrImageListsize=30)

kubernetes-config/locust-worker-controller.yaml
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+
+apiVersion: "extensions/v1beta1"
+kind: "Deployment"
+metadata:
+  name: locust-worker
+  labels:
+    name: locust-worker
+spec:
+  replicas: 5
```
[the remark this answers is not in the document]
gzhao408: oops 😅  mostly just followed by example. what would you recommend for number of workers, type of machine, etc? I have no concept of what we'd need tbh

.github/workflows/push-master.yml
gzhao408: oop! let me fix this up, slipped my mind

terraform/install.sh
nicklin37: I'd recommend using a module https://www.terraform.io/docs/configuration/modules.html to be able to keep all the loadgen stuff together and you can pass all the variables you need for the module in. You can use a script to pull the external IP and use the external data Terraform resource to retrieve it for your variables. I have an incomplete version here https://github.com/nicklin37/stackdriver-sandbox/blob/monitoring_integration/terraform/04_monitoring.tf

terraform/install.sh
nicklin37: +1 I think if you're waiting for the service already in the Terraform it might be a bit extra to wait twice here.

terraform/loadgen/00_loadgen.tf
gzhao408: just added some, our loadgenerator is Mr. Worldwide

gzhao408:
just pushed a bunch of small fixes but I tested it and I think I accidentally messed up the install script. a few moments please...

kubernetes-config/locust-worker-controller.yaml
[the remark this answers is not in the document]
Daniel-Sanche: I usually do some trial and error. I can't imagine we need more than 1 worker given our cap is 150QPS. I wonder if it's possible/feasible to use a single container instead of a master/workers (I thought that's what we did on upstream hipstershop)

For machine types, it might be good to see if a single n1-standard-2 works. if not, maybe 2 n1-standard-1's? Or try out a single higher-CPU node. We want to use the minimum resources so we're not wasting customer's money, but you don't need to spend too much time experimenting

terraform/install.sh
gzhao408: this was causing the bug haha, it's part of the while loop.
this is waiting for the frontend to become available, not the service to be running

kubernetes-config/locust-main-service.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+kind: Service
+apiVersion: v1
+metadata:
+  name: locust-main
+  labels:
+    app: locust-main
+spec:
+  ports:
+    - port: 8089
```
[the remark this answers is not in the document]
gzhao408: I think it's locust's default port for whatever reason 🤷  I can switch it out in a followup

kubernetes-config/locust-worker-controller.yaml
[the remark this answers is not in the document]
gzhao408: Ok I basically just reduced everything to 1 and things still work. From my humble noobernetes + locust knowledge, apparently the worker is there for scalability (we can increase the number of replicas / locust workers easily?)

terraform/install.sh
```diff
@@ -14,7 +14,7 @@
 #!/bin/bash
 
 # This script provisions Hipster Shop Cluster for Cloud Operations Sandbox using Terraform
-
+LOCUST_PORT="8089"
```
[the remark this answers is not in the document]
gzhao408: moved it back down, yeah the global up top doesn't make much sense

terraform/loadgen/00_loadgen.tf
gzhao408: lowered to 1

gzhao408:
Ok everything works now. Last changes:
- The image used in config files and auto-pushed in CI is loadgenerator/basic:latest. This doesn't interfere with any current loadgen stuff.
- Fewer resources, 1 worker replica, 1 n1-standard-2.
- Load Generator Worldwide™
- Cleaned up the install script - use non-global constants, don't log and execute at the same time, check variable assignment instead of repeating curl command
- Rename locust-master to locust-main
- Track the provisioning test_runner in git (oops)
- docs/style/nits

@Daniel-Sanche Ready 4 Merj? 😄

kubernetes-config/locust-main-service.yaml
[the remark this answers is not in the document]
Daniel-Sanche: Using the service, you can map any external port number to any internal one. So locust can keep listening at 8089, but you can expose it to users at the default :80 if you want

But this is pretty low priority, just a suggestion

kubernetes-config/locust-worker-controller.yaml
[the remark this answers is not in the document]
Daniel-Sanche: Ok sounds good! I don't think we need to worry too much about load generator scalability when the service itself only handles 150QPS, but yeah it's good to have it set up for easy growth later

terraform/install.sh
Daniel-Sanche: That's a bit concerning. Instead of generic sleeps, maybe you could try a `kubectl wait` for the frontend to be ready? (But this is low priority at this point)

Daniel-Sanche (approved):
LGTM!

terraform/install.sh
simonz130: +1 to Nic's suggestion. We can do it in a new PR though after this one is merged.

kubernetes-config/locust-worker-controller.yaml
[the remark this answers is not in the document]
simonz130: The service itself handles 150qps - this needs to be fixed (not in this PR). The scalability of the loadgen will help us to ramp-up the number and variety of load patterns we can send into the cluster.

make-release.sh
```diff
@@ -38,6 +38,9 @@ sed -i -e "s/cloudshell-image:v\([0-9\.]\+\)/cloudshell-image:${NEW_VERSION}/g"
 sed -i -e "s/cloudshell_git_branch=v\([0-9\.]\+\)/cloudshell_git_branch=${NEW_VERSION}/g" ${REPO_ROOT}/docs/index.html;
 sed -i -e "s/cloudshell-image:v\([0-9\.]\+\)/cloudshell-image:${NEW_VERSION}/g" ${REPO_ROOT}/docs/index.html;
 
+# update tags for load generator image
+find "${REPO_ROOT}/kubernetes-config" -name '*.yaml' -exec sed -i -e "s/basic:[0-9]+\.[0-9]+\.[0-9]+/basic:${NEW_VERSION}/g" {};
```
Daniel-Sanche, line 42: This seems to be doing a find and repace on a version code, when it should be replacing latest
You'll also need to add a line to revert back to latest later, and to run git add on the file

You can run `./make-release --dry-run` and  `./make-release --no-push` to make sure everything works as expected

Daniel-Sanche (changes requested):

make-release.sh
```diff
@@ -38,6 +38,9 @@ sed -i -e "s/cloudshell-image:v\([0-9\.]\+\)/cloudshell-image:${NEW_VERSION}/g"
 sed -i -e "s/cloudshell_git_branch=v\([0-9\.]\+\)/cloudshell_git_branch=${NEW_VERSION}/g" ${REPO_ROOT}/docs/index.html;
 sed -i -e "s/cloudshell-image:v\([0-9\.]\+\)/cloudshell-image:${NEW_VERSION}/g" ${REPO_ROOT}/docs/index.html;
 
+# update tags for load generator image
+find "${REPO_ROOT}/kubernetes-config" -name '*.yaml' -exec sed -i -e "s/:latest/:${NEW_VERSION}/g" {};
```
Daniel-Sanche, line 42: This looks like it's missing the \ at the end. Otherwise it gives an error

Also, it would probably fit better with the other line under `# temporarily pin manifests to :$NEW_VERSION`

make-release.sh
```diff
@@ -56,7 +59,10 @@ else
     # change back manifests to :latest
     find "${REPO_ROOT}/kubernetes-manifests" -name '*.yaml' -exec sed -i -e "s/:${NEW_VERSION}/:latest/g" {} \;
     git add "${REPO_ROOT}/kubernetes-manifests/*.yaml"
-    git commit -m "revert to latest images"
+    git commit -m "revert demo app images to latest"
+    find "${REPO_ROOT}/kubernetes-config" -name '*.yaml' -exec sed -i -e "s/:${NEW_VERSION}/:latest/g" {} \;
+    git add "${REPO_ROOT}/kubernetes-config/*.yaml"
+    git commit -m "revert loadgeenerator images to latest"
```
Daniel-Sanche, line 65: This shouldn't have to be a new commit. You should be able to make the changes as part of the other one

Also, you need to make the `git add "${REPO_ROOT}/kubernetes-config/*.yaml"` part of the other commit too, to make sure it's in the release commit (put it under `# create release commit`)

Daniel-Sanche, line 65: Also, this is a larger problem I'm just noticing now: Why is the load generator in a new `/kubernetes-config` file? It's confusing to have both that and `/kubernetes-manifests` in the repo root. i'd suggest putting your manifests in `/kubernetes-manifests/load-generator`

Daniel-Sanche (changes requested):

simonz130 (approved):

gzhao408:
refactored the manifests, fixed up some stuff in make-release script

i added this branch to the e2e config so it'll run here, also testing using the cloud shell link.
once it passes i'll revert the top commit

gzhao408:
Reverted https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/429/commits/0a27a520ed162af085491e030def81fc0b19ec1c, it passed.

Daniel-Sanche (approved):
LGTM

gzhao408:
OK IT WORKS

Issue #372, opened 2020-08-24
Errors in python microservices of the demo app after provisioning completes


simonz130:
The issue is here and it's related to OpenTelemetry exporter:
/usr/local/lib/python3.7/site-packages/google/auth/_default.py:354

```python
google.auth.exceptions.DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application. For more information, please see https://cloud.google.com/docs/authentication/getting-started
at default (/usr/local/lib/python3.7/site-packages/google/auth/_default.py:354)
at _create_composite_credentials (/usr/local/lib/python3.7/site-packages/google/api_core/grpc_helpers.py:217)
at create_channel (/usr/local/lib/python3.7/site-packages/google/api_core/grpc_helpers.py:280)
at create_channel (/usr/local/lib/python3.7/site-packages/google/cloud/trace_v2/gapic/transports/trace_service_grpc_transport.py:99)
at __init__ (/usr/local/lib/python3.7/site-packages/google/cloud/trace_v2/gapic/transports/trace_service_grpc_transport.py:69)
at __init__ (/usr/local/lib/python3.7/site-packages/google/cloud/trace_v2/gapic/trace_service_client.py:187)
at __init__ (/usr/local/lib/python3.7/site-packages/opentelemetry/exporter/cloud_trace/__init__.py:90)
at <module> (email_server.py:218)
```

gzhao408:
Should this be an issue opened in the opentelemetry repo maybe?

gzhao408:
I'm having trouble reproducing this issue, where did you get this error log from?

What we can do is error first if we can't find the `GOOGLE_APPLICATION_CREDENTIALS ` or `GCLOUD_PROJECT` or `PROJECT_ID` or `GOOGLE_CLOUD_PROJECT` environment variables, but it's more likely to create misleading error messages that aren't actually related to authentication (like the case with the golang services).

eaberman:
I may be getting this error as well. I looked in Error Reporting in the Sandbox I provisioned in non-Google account yesterday. This is what it looks like.

![Screenshot 2020-08-25 at 4 04 05 PM](https://user-images.githubusercontent.com/38052620/91227797-ba455e00-e6ec-11ea-88ae-d5517d5357c5.png)
[...]

PR #435, opened 2020-09-04
Pass GOOGLE_APPLICATION_CREDENTIALS to Python microservices as secret env variable

nicoleczhu:
Solves #372 DefaultCredentialsError where python services cant autodetect GOOGLE_APPLICATION_CREDENTIALS, needed by tracing exporter. 

Implements [suggestion 3](https://cloud.google.com/kubernetes-engine/docs/tutorials/authenticating-to-cloud-platform):
1. During cluster provisioning, we retrieve GKE’s default services key
2. Pass key.json into K8S cluster as a secret (recommended)
3. Python microservices (email & recommendation) now recognize GOOGLE_APPLICATION_CREDENTIALS as an environment variable 

CMIIW, the part that tripped me up a bit is that `cloudtracespanexporter` expects Google credentials, not just Project_id. My understanding of the original bug is that the Python exporter doesn't detect default credentials, [when we expect it to](https://cloud.google.com/docs/authentication/production).

### DIFF
--- a/terraform/03_gke_cluster.tf
+++ b/terraform/03_gke_cluster.tf
@@ -146,13 +146,33 @@ resource "null_resource" "install_istio" {
   depends_on = [null_resource.set_gke_context]
 }
 
+# Get GKE cluster's default service credentials through project_number
+# Solves Issue #372. When upstream bug is resolved, remove this.
+resource "null_resource" "get_credentials" {
+  provisioner "local-exec" {
+    command = "gcloud iam service-accounts keys create key.json --iam-account $(gcloud projects list --filter='${data.google_project.project.project_id}' --format='value(PROJECT_NUMBER)')-compute@developer.gserviceaccount.com"
+  }
+
+  depends_on = [null_resource.install_istio]
+}
+
+# Forwards credentials as a secret called "my-key"
+# Accessible to containers via GOOGLE_APP_CREDENTIALS env variable.
+resource "null_resource" "forward_credentials" {
+  provisioner "local-exec" {
+    command = "kubectl create secret generic my-key --from-file=key.json=key.json"
+  }
+
+  depends_on = [null_resource.get_credentials]
+}
+
 # Deploy microservices into GKE cluster 
 resource "null_resource" "deploy_services" {
   provisioner "local-exec" {
     command = "kubectl apply -f ../kubernetes-manifests"
   }
 
-  depends_on = [null_resource.install_istio]
+  depends_on = [null_resource.forward_credentials]
 }
 
 # We wait for all of our microservices to become available on kubernetes
--- a/kubernetes-manifests/emailservice.yaml
+++ b/kubernetes-manifests/emailservice.yaml
@@ -26,9 +26,19 @@ spec:
         app: emailservice
     spec:
       terminationGracePeriodSeconds: 5
+      volumes:
+      - name: google-cloud-key
+        secret:
+          secretName: my-key
       containers:
       - name: server
         image: gcr.io/stackdriver-sandbox-230822/sandbox/emailservice:latest
+        volumeMounts:
+        - name: google-cloud-key
+          mountPath: /var/secrets/google
+        env:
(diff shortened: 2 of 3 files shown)


nicoleczhu:
The failing CI tests are a bit of a chicken and egg issue: the secret is created during GKE provisioning & doesn't pre-exist locally. So CI tests fail on a missing envar value. BUT, I need to declare envars in the manifest files, in order to fix this bug in the first place. 

**Initial thoughts on solutions:** 
- Post injecting envars into pods solve the CI issue. But then we still get error logs in the few seconds right after deploying pods, right before the envars are injected.
- Use helm to manage deployment/env conditions, so local deployments don't see the envar declarations
- Wait until upstream bug is fixed, since this PR just patches behavior that shouldn't be occurring in the first place
- Hope GKE logging and monitoring agents will use Workload Identity rather than the node's service account soon, as [mentioned in Limitations here](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity). 

Or is there a simple solution I'm not seeing here?

terraform/03_gke_cluster.tf
```diff
@@ -146,13 +146,33 @@ resource "null_resource" "install_istio" {
   depends_on = [null_resource.set_gke_context]
 }
 
+# Get GKE cluster's default service credentials through project_number
+# Solves Issue #372. When upstream bug is resolved, remove this.
+resource "null_resource" "get_credentials" {
+  provisioner "local-exec" {
+    command = "gcloud iam service-accounts keys create key.json --iam-account $(gcloud projects list --filter='${data.google_project.project.project_id}' --format='value(PROJECT_NUMBER)')-compute@developer.gserviceaccount.com"
```
simonz130, line 153: Is there terraform provider for iam? If so, it's better to use it as opposed to a gcloud. However it can also be solved separately, after this PR is merged.

terraform/03_gke_cluster.tf
```diff
@@ -146,13 +146,33 @@ resource "null_resource" "install_istio" {
...
+  }
+
+  depends_on = [null_resource.install_istio]
+}
+
+# Forwards credentials as a secret called "my-key"
+# Accessible to containers via GOOGLE_APP_CREDENTIALS env variable.
+resource "null_resource" "forward_credentials" {
+  provisioner "local-exec" {
+    command = "kubectl create secret generic my-key --from-file=key.json=key.json"
```
simonz130, line 163: similar note on using terraform provider for k8s if possible

simonz130 (commented):
> * Hope GKE logging and monitoring agents will use Workload Identity rather than the node's service account soon, as [mentioned in Limitations here](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity).
> 

The problem AFAIU is not logging or monitoring agents using incorrect identity for auth but rather Trace client library (used by python trace exporter "behind the scenes") that expects GOOGLE_APPLICATION_CREDENTIALS to be set to do auth.
[Here](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#using_from_your_code) it says that using workload identity from GKE cluster and binding it to service accounts should be "transparent" to client libraries in the sense that auth will just work. I would try to explore this path first and see if there is a bug. If so, we need to fix it in auth of client libraries.

terraform/03_gke_cluster.tf
nicoleczhu: Noted on the above. Yes, it looks like I can set a "kubernetes_secret" to do this instead. Similarly with Google IAM.

nicoleczhu:
Closing this PR since: 

1. We should be using Workload Identity to manage auth for k8s. I'll look into this to solve #372 next.

2. The failing CI tests in this solution can't be patched at our level. This PR shouldn't be merged. The local deployment is failing because stackdriver library errors on invalid keys & fails the entire deployment. Where as we expect this kind of error (at the lib auth level) would fail more gracefully and not fail the entire app deployment.

PR #448, opened 2020-09-09
Set up Workload Identity for GKE

nicoleczhu:
- GKE now uses Workload Identity to authenticate to Google Cloud services, as a best practice
- Solves #372, ~50% of time (see race condition) 

**Race condition**
There exists a ~50% race condition where we still get a DefaultCredentialsError in the first few seconds. This is due to documented [WorkloadIdentity limitations](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#limitations): 
`The GKE metadata server takes a few seconds to start to run on a newly created Pod. Therefore, attempts to authenticate or authorize using Workload Identity made within the first few seconds of a Pod's life may fail. Retrying the call will resolve the problem.`

**Suggestions to solve RC:** 
- Introduce latency between when Python pods start up, vs when code is executed
- File issue with Python auth lib team to resolve it. Yoshi team noted it's bad ux. 
- Do nothing, since the Error only lasts in the first few seconds.
- [Simon] use initContainer, wait till metadata server is ready


terraform/03_gke_cluster.tf
```diff
@@ -117,33 +128,62 @@ resource "null_resource" "current_project" {
...
+      gcloud container clusters get-credentials cloud-ops-sandbox --zone ${element(random_shuffle.zone.result, 0)} --project ${data.google_project.project.project_id}
+      kubectl annotate serviceaccount --namespace default default iam.gke.io/gcp-service-account=${google_service_account.set_gsa.email}
+    EOT
+  }
+
+  depends_on = [google_service_account_iam_binding.set_gsa_binding]
+}
+
+# Set editor role for GSA to enable Cloud Trace functionality in cluster
+resource "null_resource" "set_editor" {
```
simonz130, line 172: too much power.
Can we use roles/cloudtrace.agent?
If not, then maybe roles/cloudtrace.user?

kubernetes-manifests/emailservice.yaml
```diff
@@ -26,6 +26,10 @@ spec:
         app: emailservice
     spec:
       terminationGracePeriodSeconds: 5
+      initContainers:
+      - name: wait-for-gke-metadata-server
+        image: gcr.io/stackdriver-sandbox-230822/sandbox/emailservice:latest
+        command: ['sh', '-c', 'sleep 5']
```
simonz130, line 32: Can we try to reach metadata endpoint instead of waiting 5 sec?

terraform/03_gke_cluster.tf
nicoleczhu: Sure. I'll extend the permissioning to cover a monitoring role as well. Otherwise I get: 
`com.google.api.gax.rpc.PermissionDeniedException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: Permission monitoring.timeSeries.create denied (or the resource may not exist).`

nicoleczhu: It might need to be `cloudtrace.admin` . the other two roles still yielded access errors

nicoleczhu: Actually, we need to keep it as `editor` role. My inline comment was wrong/misleading. Trace isn’t the only set of permissions this GSA needs, esp during deployment. Given the WorkloadId setup, this GSA now replaces Compute Engine default service account, which had editor role. So we can keep parity.

Btw, I did try and demote GSA roles to just trace.admin & monitoring.editor. During deploy, all other Google APIs became disabled (iam/error/all permissions were missing).

kubernetes-manifests/emailservice.yaml
nicoleczhu: Trying it with `for i in {1..10}; do sleep 1; if nslookup metadata.google.internal; then exit 0; fi; done` seems to work. It doesn't break local runs/CI but it introduces a some latency.

In principle, there could still be x% chance that metadata server takes longer than 10s... we can always increase the timeout duration.

terraform/03_gke_cluster.tf
simonz130: OK, sounds good, let's keep it editor role. Thanks for checking on this.
Worth updating the comment in line 171 to reflect that it's not only trace that requires editor.

simonz130 (approved):
Thanks Nicole!

nicoleczhu:
Not merging yet since test runner doesn't have iam/createSA permissions. 

Dan's pt was that we might not want to grant these permissions for security reasons. Suggested I can look into: 
- [Optimal] Restrict who can run GitHub Actions. Lock it to only run e2e tests on PRs submitted/approved by us
- Change test runner to re-use an existing service account instead of creating a new one. 
- Add a test flag to terraform

kubernetes-manifests/emailservice.yaml
```diff
@@ -26,6 +26,10 @@ spec:
         app: emailservice
     spec:
       terminationGracePeriodSeconds: 5
+      initContainers:
```
simonz130, line 29: add a comment to explain why you do this (also applies to other yaml files where you have this)

terraform/03_gke_cluster.tf
```diff
@@ -117,33 +128,62 @@ resource "null_resource" "current_project" {
...
+# Annotate KSA
+resource "null_resource" "annotate_ksa" {
+  triggers = {
+    cluster_ep = google_container_cluster.gke.endpoint  #kubernetes cluster endpoint
+  }
+
+  provisioner "local-exec" {
+    command = <<EOT
+      gcloud container clusters get-credentials cloud-ops-sandbox --zone ${element(random_shuffle.zone.result, 0)} --project ${data.google_project.project.project_id}
+      kubectl annotate serviceaccount --namespace default default iam.gke.io/gcp-service-account=${google_service_account.set_gsa.email}
```
simonz130, line 164: are there terraform providers available to do this?

nicoleczhu: I checked. There are no terraform resources for just `kubectl annotate`. 

I can use `kubernetes_service_account` if I create a new namespace/KSA, but I'm just using the default k8s namespace for simplicity.

kubernetes-manifests/emailservice.yaml
nicoleczhu: Added! Thanks for feedback

simonz130 (approved):

Issue #452, opened 2020-09-10
Provisioning test incorrectly reports status


Daniel-Sanche:
Test status is saying it passed when it should be failing

![image](https://user-images.githubusercontent.com/3914119/92670609-3af98200-f2c9-11ea-815a-135de9a9a691.png)

Issue #453, opened 2020-09-10
Connecting to existing project creates duplicate monitoring resources


daniel-sanche:
If the ` Test Using Existing Project` e2e test is moved above monitoring tests, it will fail, due to multiple resources being created. If we can integrate the tfstate files, this will be fixed

PR #460, opened 2020-09-10
Fix e2e provisioning tests

Daniel-Sanche:
fixes https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/452
- fixed test-runner.py to exit 0 when tests fail
- fixed provisioning tests to pass (use implicit context instead of named one)
- moved "Test Using Existing Project" to bottom of the end-to-end release test, so it doesn't fail due to https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/453

### DIFF
--- a/tests/provisioning/loadgen_test.py
+++ b/tests/provisioning/loadgen_test.py
@@ -35,6 +35,10 @@ class TestLoadGenerator(unittest.TestCase):
         # set kubectl context to loadgenerator
         command=('gcloud container clusters get-credentials loadgenerator --zone {0}'.format(getClusterZone()))
         subprocess.run(split(command))
+        # obtain the context name
+        command=('kubectl config current-context')
+        result = subprocess.run(split(command), encoding='utf-8', capture_output=True)
+        cls.context = result.stdout
 
     def testNodeMachineType(self):
         """Test if the machine type for the nodes is as specified"""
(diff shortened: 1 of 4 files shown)


tests/provisioning/loadgen_test.py
```diff
@@ -52,7 +52,7 @@ def testNumberOfNode(self):
 
     def testReachOfLoadgen(self):
         """Test if querying load generator returns 200"""
-        command = ("kubectl get service locust-main --context=loadgenerator -o jsonpath='{.status.loadBalancer.ingress[0].ip}'")
+        command = ("kubectl get service locust-main -o jsonpath='{.status.loadBalancer.ingress[0].ip}'")
```
simonz130, line 55: why did you remove the context? What will happen in case when two test runs happen on the same worker in parallel such that both use kubernetes clusters but one expects loadgen k8s and another expects hipstershop k8s?

simonz130 (changes requested):

tests/provisioning/loadgen_test.py
Daniel-Sanche: The context was causing the tests to break - I assume the context name wasn't set as expected in the container?

These tests are made to run inside a container, so I don't think we have to worry about context overlap as long as we follow the pattern of restricting each container to a single main process. The context is set at the top of each test as part of the setup function

simonz130: The tests should not be aware of the environment in which they are running. So even though we run tests in a container, we should still be careful when we use shared resources. 
So if we know the context name, we should update line 55 to include the right name.
If we don't know or it might take time to troubleshoot it and we want to unblock the tests, we can go ahead with the current fix but also adding a TODO comment explaining what's problematic here so we can fix it in the future.

simonz130 (commented):
.

simonz130 (approved):

PR #461, opened 2020-09-10
Add Sandboxctl to path

Daniel-Sanche:
- added sandboxctl to the path so we can now call it from anywhere in the custom cloud shell container.
- also added the terraform directory so we can call `install.sh` and `destroy.sh` from other directories (should we merge these into sandboxctl?)
- I added paths for both stackdriver-sandbox and cloud-ops-sandbox so Open in Cloud Shell buttons pointing at legacy forks still get this feature

[Open in Cloud Shell](https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/GoogleCloudPlatform/cloud-ops-sandbox.git&cloudshell_git_branch=master&cloudshell_working_dir=terraform&shellonly=true&cloudshell_image=gcr.io/sanche-testing-project/cloud-shell:test3)

### DIFF
--- a/cloud-shell/cloudshell_open_cp.sh
+++ b/cloud-shell/cloudshell_open_cp.sh
@@ -84,6 +84,8 @@ function cloudshell_open {
     done <<< "$output"
   fi
   # add sandboxctl to path
+  # the repo was previouly called stackdriver-sandbox. Now it's cloud-ops-sandbox
+  # we will add both directories to the path to support forks with either name
   export PATH="~/cloudshell_open/cloud-ops-sandbox/sre-recipes:$PATH"
   export PATH="~/cloudshell_open/stackdriver-sandbox/sre-recipes:$PATH"
   # add terraform directory to path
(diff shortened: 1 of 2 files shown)


cloud-shell/cloudshell_open_cp.sh
```diff
@@ -83,5 +83,11 @@ function cloudshell_open {
       try_envset "$line"
     done <<< "$output"
   fi
+  # add sandboxctl to path
+  export PATH="~/cloudshell_open/cloud-ops-sandbox/sre-recipes:$PATH"
+  export PATH="~/cloudshell_open/stackdriver-sandbox/sre-recipes:$PATH"
```
nicklin37, line 88: Do we need to leave a comment letting people know this is for previous releases?

nicklin37 (approved):
LGTM - left a small comment, not sure if needs to be addressed

simonz130 (approved):
Comment in code why you included both previous and current project name but other than that it's great.

cloud-shell/cloudshell_open_cp.sh
Daniel-Sanche: good point, added

PR #466, opened 2020-09-11
Log based metric fix

### LINKED ISSUE #370: Log-based Metric should use total count instead of avg per minute
The log-based metric of products ordered grouped by product is AWESOME. I'm super excited to use it. However, it's hard for me to reason about products ordered per minute (e.g. see the screenshot, where it's 0.017 for all of them because I used locust to send a steady stream of users ordering uniformly random products). I think it would make more sense to keep a counter of products ordered so far.

image

nicklin37:
Changes the alignment to show per day - results look like this: 
Screenshot 2020-09-11 at 2 22 09 PM - Display 2

Closes #370

### DIFF
--- a/terraform/monitoring/06_log_based_metric.tf
+++ b/terraform/monitoring/06_log_based_metric.tf
@@ -53,7 +53,7 @@ resource "google_monitoring_dashboard" "log_based_metric_dashboard" {
     "columns": "2",
     "widgets": [
       {
-        "title": "Number of Products Ordered grouped by Product Name",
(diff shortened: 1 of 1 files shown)


Daniel-Sanche:
Is there a test we could add for this?

nicklin37:
> Is there a test we could add for this?

The return value would be an integer that maps to an enum so I'm not too sure if it's worth testing for the alignment type?

simonz130 (approved):

PR #468, opened 2020-09-14
Patch #448: Better solves MDS/GoogleAuth/Istio race condition

### LINKED ISSUE #371: Profiler agent fails when provisioning new Sandbox
This is what I see in error reporting when provisioning new Sandbox (it doesn't say which node microservice that is)
image
(node:1) UnhandledPromiseRejectionWarning: Error: Project ID must be specified in the configuration
    at initConfigMetadata (/usr/src/app/node_modules/@google-cloud/profiler/build/src/index.js:120:15)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)

nicoleczhu:
This PR patches #448 , and fully closes #371 and #372 . Also:
- Upgrades GKE to v1.15 -> v1.16 
- Upgrades Istio 1.5-1.7, which contained the `holdApplicationUntilProxyStarts: true` flag that fixes race condition

**Details**
The previous PR didn't handle Istio's role in the race condition, esp since InitContainer runs before Istio injection. So we occasionally still got DefaultCredentialsError. This PR fixes the race conditions in a less latent/hacky way.

The Full Race Condition:
1. `Initcontainer` within `AppContainer` starts and finishes
2. `Istio` starts.
3. [In parallel as 2] AppContainer starts. _AppContainer errors and crashes bc of missing certs. This is because Istio's envoy proxy isn't always ready yet._
4. Envoy finally gets cert configs & proxy is ready. _This delay is in part due to slow GKE-MDS startup_
5. AppContainer restarts after crashing in step 3
6. AppContainer succeeds auth this time, as proxy certs are available

Steps 2-4 take 10-30 seconds. Hence this issue was difficult to consistently reproduce. 

**More info on the Istio fix** 
https://banzaicloud.com/blog/k8s-sidecars/ 
https://github.com/kubernetes/kubernetes/issues/65502 
Fix: https://github.com/istio/istio/pull/26149

**Before:**  
Initializing RecommendationService during Istio configs
Run before configuring istio

**After:** 
Initializing RecommendationService after Istio
Run after configuring istio

### DIFF
--- a/terraform/03_gke_cluster.tf
+++ b/terraform/03_gke_cluster.tf
@@ -42,6 +42,7 @@ resource "random_shuffle" "zone" {
 resource "google_container_cluster" "gke" {
   provider = google-beta
   project = data.google_project.project.project_id
+  min_master_version = "1.16.13-gke.1"
 
   # Here's how you specify the name
   name = "cloud-ops-sandbox"
--- a/terraform/03_gke_cluster.tf
+++ b/terraform/03_gke_cluster.tf
@@ -42,7 +42,7 @@ resource "random_shuffle" "zone" {
 resource "google_container_cluster" "gke" {
   provider = google-beta
   project = data.google_project.project.project_id
-  min_master_version = "1.16.13-gke.1"
+  min_master_version = "1.17.9-gke.1504"
 
   # Here's how you specify the name
   name = "cloud-ops-sandbox"
@@ -122,7 +122,7 @@ resource "google_container_cluster" "gke" {
 }
 
 
-# Set current project 
+# Set current project
 resource "null_resource" "current_project" {
   provisioner "local-exec" {
     command = "gcloud config set project ${data.google_project.project.project_id}"
@@ -187,7 +187,7 @@ resource "null_resource" "install_istio" {
   depends_on = [null_resource.set_editor]
 }
 
-# Deploy microservices into GKE cluster 
+# Deploy microservices into GKE cluster
 resource "null_resource" "deploy_services" {
   provisioner "local-exec" {
     command = "kubectl apply -f ../kubernetes-manifests"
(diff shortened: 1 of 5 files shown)


terraform/03_gke_cluster.tf
```diff
@@ -42,6 +42,7 @@ resource "random_shuffle" "zone" {
 resource "google_container_cluster" "gke" {
   provider = google-beta
   project = data.google_project.project.project_id
+  min_master_version = "1.16.13-gke.1"
```
nicklin37, line 45: I just got an email about a Linux vulnerability in certain GKE versions - do we need to worry about this? This version is listed as an OK patch: 1.16.13-gke.401

nicoleczhu: This version is stable. I think we're up to v1.18 now, so it's not a risky upgrade.

simonz130: We need to require one of these versions (that solve the vulnerability that was recently detected):

1.17.9-gke.1504
1.18.6-gke.3504

nicoleczhu: Oh wait you guys are right, it has to be one of the following patches. Will upgrade and test things shortly 
1.16.13-gke.401
1.17.9-gke.1504
1.18.6-gke.3504

Daniel-Sanche:
I set up an end-to-end test to make sure this all works well here: https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/actions/runs/256502639

Daniel-Sanche (approved):
LGTM!

nicoleczhu:
Thanks @Daniel-Sanche . 

Btw I changed the GKE version per @nicklin37 / @simonz130 's comment above. e2e tests are passing again: https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/actions/runs/256609289

Note: We have to use the 1.16.13-gke.401 patch. For some reason 1.17.9-gke.1504 is giving me a: 
`googleapi: Error 400: Node version .. is unsupported., badRequest`which is either due to some transient upstream issue or some zones not supporting this version yet.

PR #469, opened 2020-09-14
CLI refactor and hint method for sre recipes

nicklin37:
#WHAT: Refactors the sandboxctl CLI - moves the authentication and command running into the base class so it's usable by other recipes (since this will most likely be the case) and also uniforms naming across static and class methods. Also fixes some naming.

#WHY: Makes the CLI more portable for additional recipes

### DIFF
--- a/sre-recipes/README.md
+++ b/sre-recipes/README.md
@@ -21,7 +21,10 @@ To verify the root cause of the breakage in specific recipe, run:
 ```
 $ ./sandboxctl verify <recipe_name>
 ```
-
+To receive a hint about the root cause of the breakage, run:
+```
+$ ./sandboxctl hint <recipe_name>
+```
 ## Contributing
 
 To contribute a new recipe, create a new folder in the [recipes directory](./recipes). The directory must include a class that extends the abstract base class found in [recipe.py](recipe.py).
--- a/sre-recipes/recipe.py
+++ b/sre-recipes/recipe.py
@@ -20,6 +20,8 @@ behavior of each recipe.
 """
 
 import abc
+import subprocess
+import logging
 
 class Recipe(abc.ABC):
     """
@@ -40,3 +42,32 @@ class Recipe(abc.ABC):
         Verifies that the user of the recipe found the root cause
         of the breakage
         """
+
+    @abc.abstractmethod
+    def hint(self):
+        """
+        Provides a hint about the root cause of the issue
+        """
+
+    @staticmethod
+    def _run_command(command):
+        """Runs the given command and returns any output and error"""
+        process = subprocess.Popen(command.split(), stdout=subprocess.PIPE)
+        output, error = process.communicate()
+        return output, error
+
+    @staticmethod
+    def _auth_cluster():
+        """ Authenticates for kubectl commands. """
+        logging.info('Authenticating cluster')
+        get_project_command = "gcloud config list --format value(core.project)"
+        project_id, error = Recipe._run_command(get_project_command)
+        project_id = project_id.decode("utf-8").replace('"', '')
+        zone_command = "gcloud container clusters list --filter name:cloud-ops-sandbox --project {} --format value(zone)".format(project_id)
+        zone, error = Recipe._run_command(zone_command)
+        zone = zone.decode("utf-8").replace('"', '')
+        auth_command = "gcloud container clusters get-credentials cloud-ops-sandbox --project {} --zone {}".format(project_id, zone)
+        Recipe._run_command(auth_command)
+        logging.info('Cluster has been authenticated')
+
+
--- a/sre-recipes/recipe.py
+++ b/sre-recipes/recipe.py
@@ -63,7 +63,7 @@ class Recipe(abc.ABC):
         get_project_command = 'gcloud config list --format value(core.project)'
         project_id, error = Recipe._run_command(get_project_command)
         project_id = project_id.decode("utf-8").replace('"', '')
-        if project_id == '':
+        if not project_id:
             print('No project found.')
             logging.error('Could not authenticate cluster. No project ID found.')
             exit(1)
@@ -73,7 +73,7 @@ class Recipe(abc.ABC):
         zone_command = 'gcloud container clusters list --filter name:{} --project {} --format value(zone)'.format(name, project_id)
         zone, error = Recipe._run_command(zone_command)
         zone = zone.decode("utf-8").replace('"', '')
-        if zone == '':
+        if not zone:
             print('Failed to set up recipe. No cluster for {} was found.'.format(name))
             logging.error('Could not authenticate cluster. No cluster found for {} found.'.format(name))
             exit(1)
--- a/sre-recipes/recipe.py
+++ b/sre-recipes/recipe.py
@@ -85,21 +85,18 @@ class Recipe(abc.ABC):
             No output
         """
         # Verify the correct exists as a choice
-        found_ans = False
-        for choice in choices:
-            if correct_answer == choice:
-                found_ans = True
-
-        if not found_ans:
-            logging.info('Correct answer not found in available choices for prompt: {}'.format(prompt))
+        if not correct_answer in choices:
+            logging.error('Correct answer not found in available choices for prompt: {}'.format(prompt))
             return
 
+        # Show the multiple choice
         print(prompt)
-        for index in range(1, len(choices)+1):
-            print("\t {}) {}".format(index, choices[index-1]))
+        for index in range(0, len(choices)):
+            print("\t {}) {}".format(index + 1, choices[index]))
 
         answer = input('Enter the number of your answer: ')
 
+        # Verify the answer
         while True:
             try:
                 answer = int(answer)
--- a/sre-recipes/recipe.py
+++ b/sre-recipes/recipe.py
@@ -57,15 +57,19 @@ class Recipe(abc.ABC):
         return output, error
 
     @staticmethod
-    def _auth_cluster():
+    def _auth_cluster(cluster="APP"):
         """ Authenticates for kubectl commands. """
         logging.info('Authenticating cluster')
         get_project_command = "gcloud config list --format value(core.project)"
         project_id, error = Recipe._run_command(get_project_command)
         project_id = project_id.decode("utf-8").replace('"', '')
-        zone_command = "gcloud container clusters list --filter name:cloud-ops-sandbox --project {} --format value(zone)".format(project_id)
+        name = "cloud-ops-sandbox"
+        if cluster == "LOADGEN":
+            name = "loadgenerator"
+        zone_command = "gcloud container clusters list --filter name:{} --project {} --format value(zone)".format(name, project_id)
         zone, error = Recipe._run_command(zone_command)
         zone = zone.decode("utf-8").replace('"', '')
-        auth_command = "gcloud container clusters get-credentials cloud-ops-sandbox --project {} --zone {}".format(project_id, zone)
+        print(zone)
+        auth_command = "gcloud container clusters get-credentials {} --project {} --zone {}".format(name, project_id, zone)
         Recipe._run_command(auth_command)
         logging.info('Cluster has been authenticated')
--- a/sre-recipes/recipe.py
+++ b/sre-recipes/recipe.py
@@ -73,3 +73,44 @@ class Recipe(abc.ABC):
         auth_command = "gcloud container clusters get-credentials {} --project {} --zone {}".format(name, project_id, zone)
         Recipe._run_command(auth_command)
         logging.info('Cluster has been authenticated')
+
+    @staticmethod
+    def _generate_multiple_choice(prompt, choices, correct_answer):
+        """ Creates a multiple choice quiz using numeric answers and prints to terminal. Automatically polls for user response.
+        Input:
+            prompt - (string) the question asked to the user
+            choice - (list of strings) a list of responses. They will automatically be ennumerated 
+            correct_answer - (string) the correct answer - must string match with one of the entries in the choice array
+        Output:
+            No output
+        """
+        # Verify the correct exists as a choice
+        found_ans = False
+        for choice in choices:
+            if correct_answer == choice:
+                found_ans = True
+
+        if not found_ans:
+            logging.info('Correct answer not found in available choices for prompt: {}'.format(prompt))
+            return
+
+        print(prompt)
+        for index in range(1, len(choices)+1):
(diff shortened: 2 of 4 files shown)


sre-recipes/README.md
```diff
@@ -21,7 +21,10 @@ To verify the root cause of the breakage in specific recipe, run:
 ```
 $ ./sandboxctl verify <recipe_name>
 ```
-
+To receive a hint about the root cause of the breakage, run:
+```
+$ ./sandboxctl hint <recipe_name>
```
simonz130, line 26: nice!

sre-recipes/recipe.py
```diff
@@ -40,3 +42,30 @@ def verify(self):
...
+        return output, error
+
+    @staticmethod
+    def _auth_cluster():
+        """ Authenticates for kubectl commands. """
+        logging.info('Authenticating cluster')
+        get_project_command = "gcloud config list --format value(core.project)"
+        project_id, error = Recipe._run_command(get_project_command)
+        project_id = project_id.decode("utf-8").replace('"', '')
+        zone_command = "gcloud container clusters list --filter name:cloud-ops-sandbox --project {} --format value(zone)".format(project_id)
```
simonz130, line 66: perhaps worth making the name of the gke cluster to authenticate as an input parameter to the method, in case we might want to use it to switch context to loadgen cluster in order to send load into demo app as part of the recipe

simonz130 (approved):

sre-recipes/README.md
nicklin37: Thanks!!

sre-recipes/recipe.py
nicklin37: good idea! i added a parameter to switch to loadgen but I configured it to have a default value for the app cluster.

sre-recipes/recipes/currencies_recipe/currencies_recipe.py
```diff
@@ -124,17 +117,17 @@ def _cause_multiple_choice(self):
...
         """Verifies the user found which service broke"""
-        answer = self._service_multiple_choice()
+        answer = self.service_multiple_choice()
         while answer.lower() != 'f':
             print('Incorrect. Please try again.')
             answer = input('Your answer: ')
         print('Correct! The frontend service is broken.')
 
-    def _verify_broken_cause(self):
+    def verify_broken_cause(self):
```
eliotw, line 128: As I mentioned in the meeting, would it make sense to refactor these into a common method that can be used across different recipes. Perhaps a method that takes in a prompt, a list of choices, and the correct answer.

This would help duplicate some of the logic and formatting logic so if you need to change it, it will only be in a single location.

nicklin37: Great idea! I've refactored the generating multiple choice method into the base class to make things simpler!

sre-recipes/recipe.py
```diff
@@ -40,3 +42,75 @@ def verify(self):
...
+        Output:
+            No output
+        """
+        # Verify the correct exists as a choice
+        found_ans = False
+        for choice in choices:
+            if correct_answer == choice:
+                found_ans = True
+
+        if not found_ans:
```
eliotw, line 93: Could this simply be `if not correct_answer in choices:`

sre-recipes/recipe.py
```diff
@@ -40,3 +42,75 @@ def verify(self):
...
+            logging.info('Correct answer not found in available choices for prompt: {}'.format(prompt))
```
eliotw, line 94: Does a logging level of info make sense here? Since it would indicate an error in the recipe code maybe it should be a higher level?

sre-recipes/recipe.py
```diff
@@ -40,3 +42,75 @@ def verify(self):
...
+            return
+
+        print(prompt)
+        for index in range(1, len(choices)+1):
+            print("\t {}) {}".format(index, choices[index-1]))
```
eliotw, line 99: When using a for loop with a index variable, I think it is generally easier for people to understand it when the variable is actually used as the list's index (i.e. choices[index] instead of choices[index-1]). This also avoids having to start your loop with 1.

I would recommend:

```
for index in range(len(choices)):
  print("\t {}) {}".format(index + 1, choices[index]))

**Review by eliotw — APPROVED** [review]

File: sre-recipes/recipe.py (hunk 4 above)
Comment (nicklin37) [reply sre-recipes/recipe.py]: yes! fixed!

File: sre-recipes/recipe.py (hunk 5 above)
Comment (nicklin37) [reply sre-recipes/recipe.py]: Good idea - moved it to error level

File: sre-recipes/recipe.py (hunk 6 above)
Comment (nicklin37) [reply sre-recipes/recipe.py]: Fixed!

File: sre-recipes/recipe.py [hunk 7]
```diff
@@ -40,3 +42,72 @@ def verify(self):
...
+
+    @staticmethod
+    def _run_command(command):
+        """Runs the given command and returns any output and error"""
+        process = subprocess.Popen(command.split(), stdout=subprocess.PIPE)
+        output, error = process.communicate()
+        return output, error
+
+    @staticmethod
+    def _auth_cluster(cluster="APP"):
```
Daniel-Sanche, line 60: nit: it might make more sense to use an enum for the cluster instead of a string

sre-recipes/recipe.py
```diff
@@ -40,3 +42,72 @@ def verify(self):
...
+            No output
+        """
+        # Verify the correct exists as a choice
+        if not correct_answer in choices:
+            logging.error('Correct answer not found in available choices for prompt: {}'.format(prompt))
+            return
+
+        # Show the multiple choice
+        print(prompt)
+        for index in range(0, len(choices)):
```
Daniel-Sanche, line 94: This is a good place for [enumerate](https://book.pythontips.com/en/latest/enumerate.html): iterate over the items and their indices at once:
`for index, choice in enumerate(choices, 1):
    print("\t {}) {}".format(index, choice))

sre-recipes/recipe.py
```diff
@@ -40,3 +42,72 @@ def verify(self):
...
+        Recipe._run_command(auth_command)
+        logging.info('Cluster has been authenticated')
+
+    @staticmethod
+    def _generate_multiple_choice(prompt, choices, correct_answer):
+        """ Creates a multiple choice quiz using numeric answers and prints to terminal. Automatically polls for user response.
+        Input:
+            prompt - (string) the question asked to the user
+            choice - (list of strings) a list of responses. They will automatically be ennumerated 
+            correct_answer - (string) the correct answer - must string match with one of the entries in the choice array
```
Daniel-Sanche, line 83: would it make more sense for correct_answer to be an index? It seems a bit ugly to expect the same string to be input twice here, and requiring string comparison

sre-recipes/recipe.py
```diff
@@ -40,3 +42,72 @@ def verify(self):
...
+                    print('Not a valid choice.')
+                elif choices[answer-1] == correct_answer:
+                    print('Correct!')
+                    return
+                else:
+                    print('Incorrect. Try again.')
+            except ValueError:
+                print('Please enter the number of your answer.')
+
+            answer = input('Enter the number of your answer: ')
```
Daniel-Sanche, line 113: nit: It would be nice if you could find a way to restructure this loop to avoid repeating this line

sre-recipes/recipe.py
Daniel-Sanche: although I guess that could be confusing since you're presenting a 1-based index to users

sre-recipes/recipes/currencies_recipe/currencies_recipe.py
```diff
@@ -48,103 +41,74 @@ def _deploy_state(state):
...
-                "\t[g] payment service \n"
-                "\t[h] product catalog service \n"
-                "\t[i] recommendation service \n"
-                "\t[j] redis \n"
-                "\t[k] shipping service")
-        answer = input('Your answer: ')
-        return answer
+        prompt = 'Which service has a breakage?'
+        choices = ['email service', 'frontend service', 'ad service', 'recommendation service', 'currency service']
+        answer = 'frontend service'
```
Daniel-Sanche, line 94: Re-typing 'frontend service' in the choices and answer seems a little risky to me. Strings should really only be defined once in code. What if it changes in one place but not the other?

I already suggested using an index for the answer before. Another option might be to make choices a list of tuples containing the name and the success value: `[('email service', False), ('frontend service', True) ...]`

This way would keep things more tightly coupled. But this is just a suggestion, I don't think it's a huge issue as is either

sre-recipes/recipe.py
```diff
@@ -40,3 +42,72 @@ def verify(self):
...
+        get_project_command = "gcloud config list --format value(core.project)"
+        project_id, error = Recipe._run_command(get_project_command)
+        project_id = project_id.decode("utf-8").replace('"', '')
+        name = "cloud-ops-sandbox"
+        if cluster == "LOADGEN":
+            name = "loadgenerator"
+        zone_command = "gcloud container clusters list --filter name:{} --project {} --format value(zone)".format(name, project_id)
+        zone, error = Recipe._run_command(zone_command)
+        zone = zone.decode("utf-8").replace('"', '')
+        print(zone)
```
Daniel-Sanche, line 72: Is this meant to be printed to the user? Or should be be logged with `logging` like below?

Daniel-Sanche (commented):
I added some comments, but nothing super important that needs to be addressed

sre-recipes/recipes/currencies_recipe/currencies_recipe.py
```diff
@@ -48,103 +41,74 @@ def _deploy_state(state):
...
+        print('Done')
         logging.info('Deployed working service')
 
-    def _service_multiple_choice(self):
+    def hint(self):
+        """
+        Provides a hint about finding the root cause of this recipe
+        """
+        print('Giving hint for recipe')
+        external_ip_command = "kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'"
```
Daniel-Sanche, line 78: I tried running `sandboxctl hint recipe0` without authenticating to a cluster, and it seemed to hang forever. Can we error handle this somehow? Or maybe authenticate to the cluster by default?

sre-recipes/recipe.py
```diff
@@ -40,3 +42,72 @@ def verify(self):
...
+    def _run_command(command):
+        """Runs the given command and returns any output and error"""
+        process = subprocess.Popen(command.split(), stdout=subprocess.PIPE)
+        output, error = process.communicate()
+        return output, error
+
+    @staticmethod
+    def _auth_cluster(cluster="APP"):
+        """ Authenticates for kubectl commands. """
+        logging.info('Authenticating cluster')
```
Daniel-Sanche, line 62: Should we add any error handling here? What if the cluster isn't set up?

nicklin37: I set up error handling to 1) make sure there's a project, and 2) make sure there's a cluster to authenticate against. Do you think this is enough?

sre-recipes/recipes/currencies_recipe/currencies_recipe.py
nicklin37: Yeah I added an auth command to this

sre-recipes/recipe.py
nicklin37: Hehe it is gone now! :D - took it out and replaced with a better message

sre-recipes/recipe.py
nicklin37: Fixed!

sre-recipes/recipe.py
nicklin37: Fixed!

sre-recipes/recipes/currencies_recipe/currencies_recipe.py
nicklin37: Hmm I do agree it makes it dangerous to have the string declared twice - I think using an index might be confusing to the developer because of the 0 vs 1 start which I guess we could just start the index at 0 regardless? I think the tuple also makes it more work for the developer since they have to specifically designate whether each option is correct or incorrect. I'm not too sure what the best option here is- I think we want to keep it relatively simple for people to create these multiple choice questions but also avoid room for errors.

sre-recipes/recipe.py
nicklin37: Further discussion in the thread below!

sre-recipes/recipes/currencies_recipe/currencies_recipe.py
Daniel-Sanche: Yeah, I think it's probably fine as is. If we were expecting this class to grow a lot in complexity we might want to re-think it, but I'm not too worried for our use case. Especially since you added a check at the top of the function

sre-recipes/recipe.py
Daniel-Sanche: Yeah, looks good!

sre-recipes/recipe.py
```diff
@@ -40,3 +42,76 @@ def verify(self):
...
+        get_project_command = 'gcloud config list --format value(core.project)'
+        project_id, error = Recipe._run_command(get_project_command)
+        project_id = project_id.decode("utf-8").replace('"', '')
+        if project_id == '':
```
Daniel-Sanche, line 66: nit: It's usually recommended to do `if not project_id` to check for empty strings (it evaluates to False)

sre-recipes/recipe.py
```diff
@@ -40,3 +42,76 @@ def verify(self):
...
+            logging.error('Could not authenticate cluster. No project ID found.')
+            exit(1)
+        name = "cloud-ops-sandbox"
+        if cluster == 'LOADGEN':
+            name = 'loadgenerator'
+        zone_command = 'gcloud container clusters list --filter name:{} --project {} --format value(zone)'.format(name, project_id)
+        zone, error = Recipe._run_command(zone_command)
+        zone = zone.decode("utf-8").replace('"', '')
+        if zone == '':
+            print('Failed to set up recipe. No cluster for {} was found.'.format(name))
```
Daniel-Sanche, line 77: nit: it would probably be better to use try blocks and exceptions so you can handle these errors in one place instead of disrupting the flow of the code. But with only 2 I don't think it matters too much

Daniel-Sanche (approved):
LGTM, with a couple more optional comments

sre-recipes/recipe.py
nicklin37: Leaving as is for now - might be useful to move to enum later.

PR #470, opened 2020-09-16
SRE Recipe introducing errors into the email service

nicklin37:
#WHAT: Another SRE recipe that introduces grpc errors into the email service by attempting to encode the email when the feature flag for encoding is set to true. 

#WHY: We want another example of SRE recipes for users to play with!

#HOW: used encoding as a sample "feature" to hide what the recipe really is - which is about sending errors in both logging that should also be shown in dashboards and SLOs.

### DIFF
--- a/src/emailservice/email_server.py
+++ b/src/emailservice/email_server.py
@@ -20,6 +20,7 @@ import os
 import sys
 import time
 import grpc
+import base64
 from jinja2 import Environment, FileSystemLoader, select_autoescape, TemplateError
 from google.api_core.exceptions import GoogleAPICallError
 from google.auth.exceptions import DefaultCredentialsError
@@ -120,7 +121,26 @@ class EmailService(BaseEmailService):
 class DummyEmailService(BaseEmailService):
   def SendOrderConfirmation(self, request, context):
     logger.info('A request to send order confirmation email to {} has been received.'.format(request.email))
+    if self.EncodeEmail(request.email):
+      logger.error('Err: Could not encode email')
+      context.set_code(grpc.StatusCode.INTERNAL)
     return demo_pb2.Empty()
+
+  def EncodeEmail(self, email):
+    """ 
+    Encodes the email address to base64 encoding
+    Input: 
+      email - (string) the email address
+    Output:
+      boolean - whether the encoding was successful or not
+    """
+    if os.getenv('ENCODE_EMAIL').lower() == "true":
+      byte_rep = email.encode('ascii')
+      b64_bytes = base64.b64encode(byte_rep)
+      encoded = b64_bytes.decode('ascii')
+      return True
+    else:
+      return False
 
 class HealthCheck():
   def Check(self, request, context):
--- a/src/emailservice/email_server.py
+++ b/src/emailservice/email_server.py
@@ -121,9 +121,13 @@ class EmailService(BaseEmailService):
 class DummyEmailService(BaseEmailService):
   def SendOrderConfirmation(self, request, context):
     logger.info('A request to send order confirmation email to {} has been received.'.format(request.email))
-    if self.EncodeEmail(request.email):
-      logger.error('Err: Could not encode email')
-      context.set_code(grpc.StatusCode.INTERNAL)
+    if os.getenv('ENCODE_EMAIL').lower() == 'true':
+      try:
+        encoded_email = self.EncodeEmail(request.email)
+        request.email = encoded_email
+      except:
+        logger.error('Err: Could not encode email')
+        context.set_code(grpc.StatusCode.INTERNAL)
     return demo_pb2.Empty()
 
   def EncodeEmail(self, email):
@@ -132,15 +136,11 @@ class DummyEmailService(BaseEmailService):
     Input: 
       email - (string) the email address
     Output:
-      boolean - whether the encoding was successful or not
+      string - the encoded email as base64
     """
-    if os.getenv('ENCODE_EMAIL').lower() == "true":
-      byte_rep = email.encode('ascii')
-      b64_bytes = base64.b64encode(byte_rep)
-      encoded = b64_bytes.decode('ascii')
-      return True
-    else:
-      return False
+    byte_rep = email.encode('ascii')
+    b64_bytes = base64.b64encode(email)
+    return b64_bytes.decode('ascii')
 
 class HealthCheck():
   def Check(self, request, context):
--- a/sre-recipes/recipes/currencies_recipe/currencies_recipe.py
+++ b/sre-recipes/recipes/currencies_recipe/currencies_recipe.py
@@ -27,6 +27,10 @@ class CurrenciesRecipe(Recipe):
     This class implements recipe 0, which purposefully
     introduces latency into the frontend service.
     """
+    name = "currency_recipe"
+    
+    def get_name(self):
+        return self.name
 
     @staticmethod
     def _deploy_state(state):
--- a/sre-recipes/recipes/currencies_recipe/currencies_recipe.py
+++ b/sre-recipes/recipes/currencies_recipe/currencies_recipe.py
@@ -27,8 +27,8 @@ class CurrenciesRecipe(Recipe):
     This class implements recipe 0, which purposefully
     introduces latency into the frontend service.
     """
-    name = "currency_recipe"
-    
+    name = "recipe0"
+
     def get_name(self):
         return self.name
 
@@ -54,6 +54,8 @@ class CurrenciesRecipe(Recipe):
         delete_pod_command = f"kubectl delete pod {service}"
         logging.info('Deleting pod: %s', delete_pod_command)
         Recipe._run_command(delete_pod_command)
+        availability_command = "kubectl wait --for=condition=available --timeout=600s deployment/frontend"
+        Recipe._run_command(availability_command)
 
     def break_service(self):
         """
--- a/sre-recipes/recipes/encoding_recipe/encoding_recipe.py
+++ b/sre-recipes/recipes/encoding_recipe/encoding_recipe.py
@@ -43,7 +43,7 @@ class EncodingRecipe(Recipe):
         Recipe._run_command(set_env_command)
         service, error = Recipe._run_command(get_pod_command)
         service = service.decode("utf-8").replace('"', '')
-        if service == '':
+        if not service:
             print('No service found. Could not deploy state.')
             logging.error('No service found. Could not deploy state.')
             exit(1)
@@ -80,7 +80,7 @@ class EncodingRecipe(Recipe):
         get_project_command = "gcloud config list --format value(core.project)"
         project_id, error = Recipe._run_command(get_project_command)
         project_id = project_id.decode("utf-8").replace('"', '')
-        if project_id == '':
+        if not project_id:
             print('No project ID found.')
             logging.error('No project ID found.')
             exit(1)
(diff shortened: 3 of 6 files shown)


nicklin37:
Unmarking as draft as it is ready for review

sre-recipes/recipes/encoding_recipe/encoding_recipe.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        state_str = str(state).lower()
+        set_env_command = f"kubectl set env deployment/emailservice ENCODE_EMAIL={state_str}"
+        get_pod_command = """kubectl get pod -l app=emailservice -o \
+            jsonpath=\"{.items[0].metadata.name}\""""
+        logging.info('Setting env variable: %s', set_env_command)
+        logging.info('Getting pod: %s', get_pod_command)
+
+        Recipe._run_command(set_env_command)
+        service, error = Recipe._run_command(get_pod_command)
```
Daniel-Sanche, line 44: It's not a super high priority, but it would be nice to see error handling. There are a lot of places where things could go wrong here, but the code always assumes success. What if a user runs `sandboxctl` after deleting their project? Will the errors they see make sense?

sre-recipes/recipes/encoding_recipe/encoding_recipe.py
```diff
@@ -0,0 +1,108 @@
...
+        Recipe._generate_multiple_choice(prompt, choices, answer)
+
+    def verify(self):
+        """Verifies the user found the root cause of the broken service"""
+        print('This is a multiple choice quiz to verify that you have')
+        print('found the root cause of the break')
+        self.verify_broken_service()
+        self.verify_broken_cause()
+        print('Good job! You have correctly identified which service broke')
+        print('and what caused it to break!')
```
Daniel-Sanche, line 108: nit: this would probably be better as a [multiline print](https://pythonprogramming.net/multi-line-printing-python-3/)

sre-recipes/recipes/encoding_recipe/encoding_recipe.py
nicklin37: Good point - I added error handling into the CLI refactor PR

sre-recipes/recipes/encoding_recipe/encoding_recipe.py
nicklin37: Fixed!

sre-recipes/recipes/encoding_recipe/encoding_recipe.py
nicklin37: Hah I pranked myself. Added error handling for the deploy state method and for the hint method

src/emailservice/email_server.py
```diff
@@ -120,8 +121,27 @@ def SendOrderConfirmation(self, request, context):
 class DummyEmailService(BaseEmailService):
   def SendOrderConfirmation(self, request, context):
     logger.info('A request to send order confirmation email to {} has been received.'.format(request.email))
+    if self.EncodeEmail(request.email):
```
Daniel-Sanche, line 124: I know that this is just a dummy function and doesn't matter too much, but it does seem a bit strange that this is checking if True, even though EncodeEmail returns true on success.

Maybe it would be more realistic to make EncodeEmail throw an exception (like `b64encode` on the string directly instead of bytes), and then catch that exception here?

nicklin37: Do you mean that EncodeEmail would still try to encode the email regardless of the flag but merely do it incorrectly in this case? I think the function does look strange it's checking True but that's mostly to add a layer of "complexity" on it

Daniel-Sanche: I was imagining something like this:

```
if os.getenv('ENCODE_EMAIL').lower() == "true":
  try:
    encoded = self.EncodeEmail(request.email)
    request.email = encoded
  except:
    logger.error('Err: Could not encode email')
    context.set_code(grpc.StatusCode.INTERNAL)
...
  def EncodeEmail(self, email):
      byte_rep = email.encode('ascii')
      b64_bytes = base64.b64encode(email)
      return b64_bytes.decode('ascii')
```

This should give you the same result, while looking more like a real feature with a bug. But maybe I'm misunderstanding what you're doing here

nicklin37: Ah I see - no this makes sense and it does mask the feature a lot better. I made the adjustment!

Daniel-Sanche:
When I run sandboxctl, it's telling me my options are `recipe0` and `recipe12`

nicklin37:
> When I run sandboxctl, it's telling me my options are `recipe0` and `recipe12`. Running `./sandboxctl break recipe12` throws an error

I've fixed the error in the commands - it should work now. As far as the recipe number I'm having a little trouble tracking down how the attributes are getting labeled...

nicklin37:
> > When I run sandboxctl, it's telling me my options are `recipe0` and `recipe12`. Running `./sandboxctl break recipe12` throws an error
> 
> I've fixed the error in the commands - it should work now. As far as the recipe number I'm having a little trouble tracking down how the attributes are getting labeled...

Nevermind I figured it out.. but I'm not too sure if my solution is the most elegant.

sre-recipes/sandboxctl
```diff
@@ -41,11 +41,13 @@ def get_valid_recipes(attributes):
     """
 
     recipe_objs = {}
-    for recipe_num, (attribute_name, attribute) in enumerate(attributes):
+    recipe_num = 0
+    for (attribute_name, attribute) in attributes:
```
Daniel-Sanche, line 45: This works. Another (potentially cleaner) way would be to filter the list before iteration rather than as part of it:

```
filtered = [a for a in attributes if isclass(a) and a is not Recipe and issubclass(a, Recipe)]
for recipe_num, (attribute_name, attribute) in enumerate(filtered):
```
But I don't think it matters too much either way

Daniel-Sanche: And I guess if the TypeErrors are expected, your way might be safer since it only counts recipes that it knows work

sre-recipes/sandboxctl
```diff
@@ -41,11 +41,13 @@ def get_valid_recipes(attributes):
...
         if isclass(attribute) and attribute is not Recipe and issubclass(attribute, Recipe):
             try:
                 recipe_obj = attribute()
                 recipe_objs[f"recipe{recipe_num}"] = recipe_obj
+                recipe_num += 1
```
Daniel-Sanche, line 50: I do kind of question using index numbers as recipe identifiers though. Are we sure that this sorting will always be stable? Isn't it possible that previous identifiers change if you add new recipes?

sre-recipes/sandboxctl
nicklin37: ah yeah that's true - I think we can keep it as is then!

sre-recipes/sandboxctl
nicklin37: I was thinking the same earlier - and I think at the moment how it is configured is alphabetical in a sense so if someone wrote a "a_recipe" that recipe would then become recipe0 and then the rest of the recipes would shift accordingly...

Daniel-Sanche: It looks like it should be pretty easy to add a `get_name` method or attribute to all recipes, and grab it with `recipe_obj.get_name()`

Daniel-Sanche (changes requested):
I added a suggestion to add names as a part of the recipe object. Also, I'm seeing an error when I try to run `sandboxctl break recipe1`:

```
  File "/home/sanche/cloudshell_open/stackdriver-sandbox/sre-recipes/recipes/encoding_recipe/encoding_recipe.py", line 60, in break_service
    Recipe._auth_cluster()
AttributeError: type object 'Recipe' has no attribute '_auth_cluster'
```

nicklin37:
> I added a suggestion to add names as a part of the recipe object. Also, I'm seeing an error when I try to run `sandboxctl break recipe1`:
> 
> ```
>   File "/home/sanche/cloudshell_open/stackdriver-sandbox/sre-recipes/recipes/encoding_recipe/encoding_recipe.py", line 60, in break_service
>     Recipe._auth_cluster()
> AttributeError: type object 'Recipe' has no attribute '_auth_cluster'
> ```

Sorry this was because this PR started before the refactor got merged and I never rebased - it should be fixed now if you want to test. I'm working on fixing the naming rn.

sre-recipes/sandboxctl
nicklin37: Gotcha - I gave them all a name now, this is what the help command looks like:
```
Usage: sandboxctl [OPTIONS] [break|restore|verify|hint] [currency_recipe|encoding_recipe]
```

Daniel-Sanche (commented):
I was able to get sandboxctl to run the new recipe without errors, but I'm not sure I'm seeing the logs we expect for the inserted issue. My emailservice logs seem flooded with a warning about the Profiler API

sre-recipes/recipes/currencies_recipe/currencies_recipe.py
```diff
@@ -27,6 +27,10 @@ class CurrenciesRecipe(Recipe):
     This class implements recipe 0, which purposefully
     introduces latency into the frontend service.
     """
+    name = "currency_recipe"
```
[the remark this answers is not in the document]
nicklin37: Yeah I renamed them back to recipe0 and recipe1. In the future maybe there is a better way of naming cause this way the developer has to know what name is up next

sre-recipes/recipes/encoding_recipe/encoding_recipe.py
```diff
@@ -0,0 +1,128 @@
...
+        Recipe._run_command(set_env_command)
+        service, error = Recipe._run_command(get_pod_command)
+        service = service.decode("utf-8").replace('"', '')
+        if not service:
+            print('No service found. Could not deploy state.')
+            logging.error('No service found. Could not deploy state.')
+            exit(1)
+        delete_pod_command = f"kubectl delete pod {service}"
+        logging.info('Deleting pod: %s', delete_pod_command)
+        Recipe._run_command(delete_pod_command)
```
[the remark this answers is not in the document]
nicklin37: Gotcha - I added a wait command. I think it's okay for users to see that only one service is reset since hopefully they're using the recipes to learn Cloud Ops. Otherwise they could just look into the recipe code directory anyways.

nicklin37:
> I was able to get sandboxctl to run the new recipe without errors, but I'm not sure I'm seeing the logs we expect for the inserted issue. My emailservice logs seem flooded with a warning about the Profiler API

Screenshot 2020-09-18 at 3 17 39 PM - Display 2
I'm able to see this in the logs when I run it...

sre-recipes/recipes/encoding_recipe/encoding_recipe.py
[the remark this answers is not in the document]
Daniel-Sanche: I'd recommend making the wait on the emailservice, since that's the one that's reset. 

And yeah, that makes sense, just thought I would point out that `kubectl get pods` gives a bit of a spoiler if they wanted to hunt it down themselves haha

Daniel-Sanche (approved):
> I'm able to see this in the logs when I run it...

I just realized I was testing with the open in cloud shell link, which uses pre-built images instead of building myself... That explains my issue

LGTM! This looks great, good job

sre-recipes/recipes/encoding_recipe/encoding_recipe.py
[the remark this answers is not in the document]
nicklin37: Fixed!

PR #475, opened 2020-09-19
E2e Fix: Use default GCE service account for Workload Identity

nicoleczhu:
Improves on: #448

Changes: 
- Use default GCE service account for Workload Identity
- No longer creates a new Google service account for the GKE service account binding
- Fixes e2e test issue where we can't reapply to deleted service accounts

Note: While docs recommend we create new SAs for K8S clusters as a best practice, implementing it for Sandbox introduces unnecessary complexity to users. Agreeing with @Daniel-Sanche 's point that: "The only reason I'd hesitate is because we don't want to show bad patterns to our users, but I don't think workload identity is really in scope for what our customers would be looking at anyway."

### DIFF
--- a/.github/workflows/e2e-latest.yml
+++ b/.github/workflows/e2e-latest.yml
@@ -21,6 +21,7 @@ on:
       - release/*
       - e2e*
       - loadgen_integration # will remove before merging
+      - undo-gsa # will remove before merging
   workflow_dispatch:
     # trigger through UI or API
 jobs:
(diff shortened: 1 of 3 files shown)


nicoleczhu:
Passing e2e run: https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/runs/1136520030?check_suite_focus=true

Daniel-Sanche (approved):
LGTM!

I'd suggest adding this branch [here](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/2c57086e0d853ab0ec73caad6ba50dd9de7409da/.github/workflows/e2e-latest.yml#L23) so we get CI runs listed here on the PR though. We want to make sure the e2e tests pass before merging this one

simonz130 (approved):

PR #481, opened 2020-09-25
feat: node runtime12 and updated nodejs debugger

### LINKED ISSUE #480: Update nodejs container images to node v12

simonz130:
- Updated nodejs runtime to 12 and nodejs debugger version to 5.1.2 in currencyservice
- Updated gitignore to ignore istio binaries
- Updated makefile to delete test cluster before starting building a new one. Also updated cluster name to include username, so we can share the same test project on GCP
- MISC: Updated provisioning to wait for all microservices to deploy - previously we missed one

Fixes #480


Makefile
```diff
@@ -15,9 +15,14 @@
 .-PHONY: cluster deploy deploy-continuous logs check-env check-service
 
 ZONE=us-west1-a
-CLUSTER=cloud-ops-sandbox
+CLUSTER=cloud-ops-sandbox-${USER}
 
 cluster: check-env
+	EXISTING_CLUSTER=`gcloud container clusters list --zone=us-west1-a --format="value(name)" --filter="name:${CLUSTER}"`
```
Daniel-Sanche, line 21: this should use ${ZONE}

Daniel-Sanche (approved):
Small comment in the makefile, otherwise LGTM assuming the tests pass

Makefile
simonz130: Yeah, I just fixed that, thanks!

PR #486, opened 2020-09-30
Use certified image again

Daniel-Sanche:
We previously started using uncertified cloud shell images, when we discovered certified images force you to use a single version. Now that `cloudshell-image:latest` has been certified, we can go back to using that one on the website.

I updated the "rebuild cloudshell" action to keep the certified image up to date with the latest release. The uncertified images will continue to be built for each previous release, and I suggest we still use those for testing

### DIFF
--- a/cloud-shell/cloudbuild.yaml
+++ b/cloud-shell/cloudbuild.yaml
@@ -30,7 +30,7 @@ steps:
     # build latest image
     docker build -t $$UNCERTIFIED_PATH:latest ./cloud-shell
     docker push $$UNCERTIFIED_PATH:latest
-    for TAG in $(git tag); do
+    for TAG in $(git tag | sort -V); do
         echo "checking out $$TAG"
         git checkout $$TAG
         # if cloud-build directory exists, build new container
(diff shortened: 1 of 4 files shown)


cloud-shell/cloudbuild.yaml
```diff
@@ -24,25 +24,32 @@ steps:
...
+            docker push $$UNCERTIFIED_PATH:$$TAG
+        else
+            echo "No cloud-shell directory in tag $$TAG"
+        fi
     done
+
+    # update certified image
+    # build image using latest release
+    CERTIFIED_PATH=gcr.io/$PROJECT_ID/cloudshell-image
+    LATEST_TAG=$(git tag | grep "v[0-9]*\.[0-9]*\.[0-9]*" | sort -V | tail -n 1)
```
simonz130, line 49: can you do
```
git describe --abbrev=0 --tags
```
instead for simplicity?

simonz130 (approved):

cloud-shell/cloudbuild.yaml
Daniel-Sanche: That's much better!

PR #487, opened 2020-10-01
Use sandboxctl for terraform scripts

Daniel-Sanche:
It occurred to me that now that we are using sandboxctl for load generator shapes and sre recipes, it probably makes sense to put `install.sh` and `destroy.sh` in there as well, so we can have all sandbox functionality in one place

I'm not sure if this is the kind of change we'd want to add in at this point in a release cycle, but I thought I'd put a PR together to discuss it just in case


docs/README.md
```diff
@@ -438,9 +438,9 @@ You can expand any of the messages that matches the filter to see the full stack
...
 Once you have finished exploring the Cloud Operations Sandbox project, don't forget to destroy it to avoid incurring additional billing.
 
-Destroy your Sandbox project by opening the Cloud Shell and running the destroy script:
+Destroy your Sandbox project by opening the Cloud Shell and running the destroy command:
 ```
-$ destroy.sh
+$ sandboxctl destroy
 ```
 
 This script destroys the current Cloud Operations Sandbox project. If the install.sh script were run again, a Cloud Operations Sandbox project with a new project id would be created.
```
nicoleczhu, line 446: Nit for consistency since we stopped referring to "install.sh" to users, change "If the install.sh script were run again" to "If the `sandboxctl create` script were run again"

nicoleczhu: Also you may want to update `/hack/scratchpad.sh` line 30-31 to use `sandboxctl` as well
Do we use the scripts inside of /hack actually?

Daniel-Sanche: Good catch, I missed that one.

@simonz130 do you know if we still need `/hack`? It looks like it has old release scripts, but not the one we currently use

nicoleczhu (approved):

sre-recipes/sandboxctl
```diff
@@ -119,6 +119,19 @@ def loadgen(traffic_pattern):
     print(f'Loadgenerator deployed using {traffic_pattern} pattern')
     Recipe._auth_cluster('APP')
 
+@cli.command()
+def create():
+    """Create a new sandbox"""
+    os.chdir(os.path.abspath(sys.path[0]))
+    create_command = "../terraform/install.sh"
```
simonz130, line 126: will this work if I run sandboxctl from any random directory in cloud shell - assuming that sandboxctl is on the $PATH?

Daniel-Sanche: Yes, the line above this makes changes the working dir to be the location of the `sandboxctl` script, so it can be called from anywhere. We will need to make changes if we move the sandboxctl fine though

PR #489, opened 2020-10-02
Feat: customize how labeler bot detects languages with user config file

nicoleczhu:
Main changes fyi: 
- Language labeling is turned off by default (until we add this config file)
- Can config languages based on project directory structure, e.g. /sr/emailservice -> "lang: python"
- Can config custom file extension to language mappings, e.g. "filename.ts" -> "lang: js"
- Can config label copy e.g. "language: python" instead of "lang: python" 
- I turned off labeling issues since it's hit and miss and will take a lot more work to get accurate.
- Config file is optional

A few things to discuss: 
1. How do we want to label repos like /terraform, /cloudshell. Or not label them? @Daniel-Sanche 
2. Do we want to label with as many languages as possible, or a whitelisted set of languages? I can add a `verbose` config to enable/disable this
3. Are you ok with the current set of label names? e.g. `javascript` instead of `js` or `node.js`

Note: Bot is still running on my deployment atm.

### DIFF
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -1,3 +1,17 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 # Disable label by Drift product
 product: false
 language:
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -25,7 +25,7 @@ language:
   labelprefix: 'lang: '
   # To re-map extension to languages
   extensions:
-    yaml: ['tf']
+    manifest: ['tf']
     typescript: ['ts']
   # To map languages based on project directory structure:
   paths:
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -1,0 +1,35 @@
+# Disable label by Drift product
+product: false
+language:
+  # Disable issue labeling (not accurate)
+  issue: false
+  # Enable pr labeling
+  pullrequest: true
+  labelprefix: 'lang: '
+  # To re-map extension to languages
+  extensions:
+    JSON: ['JSON']
+    typescript: ['ts']
+  # To map languages based on project directory structure:
+  paths:
+    src:
+      frontend: 'go'
+      cartservice: 'c#'
+      productcatalogservice: 'go'
+      currencyservice: 'javascript'
+      paymentservice: 'javascript'
+      shippingservice: 'go'
+      emailservice: 'python'
+      checkoutservice: 'go'
+      recommendationservice: 'python'
+      adservice: 'java'
+      loadgenerator: 'python'
+    tests:
+      # To designate the default language for a subdirectory use '.'
+      .: 'python'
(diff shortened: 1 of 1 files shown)


.github/config.yml
```diff
@@ -0,0 +1,49 @@
...
+  labelprefix: 'lang: '
+  # To re-map extension to languages
+  extensions:
+    JSON: ['JSON']
+    typescript: ['ts']
+  # To map languages based on project directory structure:
+  paths:
+    src:
+      frontend: 'go'
+      cartservice: 'c#'
```
simonz130, line 31: it's c# indeed but in drift systems it's labeled as "dotnet", so we should use what DRIFT "understands".

.github/config.yml
```diff
@@ -0,0 +1,49 @@
...
+      emailservice: 'python'
+      checkoutservice: 'go'
+      recommendationservice: 'python'
+      adservice: 'java'
+      loadgenerator: 'python'
+    tests:
+      # To designate the default language for a subdirectory use '.'
+      .: 'python'
+      # More specific path designations will overwrite the default
+      cartservice: 'c#'
```
simonz130, line 45: again, "dotnet" for reasons above

.github/config.yml
```diff
@@ -0,0 +1,49 @@
...
+
+# Disable label by Drift product
+product: false
+language:
+  # Disable issue labeling (not accurate)
+  issue: false
+  # Enable pr labeling
+  pullrequest: true
+  labelprefix: 'lang: '
+  # To re-map extension to languages
```
simonz130, line 23: Does this override some default configuration stored elsewhere? If so, please add its location to the comment.

simonz130 (approved):

Daniel-Sanche:
> 1.  How do we want to label repos like /terraform, /cloudshell. Or not label them? @Daniel-Sanche

Can we add a new extension for terraform files? Otherwise, I'd probably say probably `bash` or `shell`, But it might be better to fall back to prediction based on extension for those if that's possible

nicoleczhu:
> > 1. How do we want to label repos like /terraform, /cloudshell. Or not label them? @Daniel-Sanche
> 
> Can we add a new extension for terraform files? Otherwise, I'd probably say probably `bash` or `shell`, But it might be better to fall back to prediction based on extension for those if that's possible

Yes, we can. Do you want to call it **"lang: terraform"** ?

Daniel-Sanche:
> Yes, we can. Do you want to call it **"lang: terraform"** ?

Yeah, that sounds good!

simonz130 (approved):

.github/config.yml
nicoleczhu: Yes it overrides a default config, will do.

simonz130:
For terraform files, I think we should call it yaml and not terraform (language)

Daniel-Sanche:
The `.tf` files aren't yaml though, right? Terraform uses their [own configuration language](https://www.terraform.io/docs/configuration/index.html). My vote would still be for terraform, but "configuration" or "config" could also work?

nicoleczhu:
> The `.tf` files aren't yaml though, right? Terraform uses their [own configuration language](https://www.terraform.io/docs/configuration/index.html). My vote would still be for terraform, but "configuration" or "config" could also work?

That's also true. In this case, it depends on why we want the language labels in the first place. i.e. whether we want our OKR dashboard to specifically track "terraform" as its own language, or if we just want to lump it under "yaml" which is broadly a "config" type language anyway. 

I suggest: 
A) Keep `yml` label for terraform, since we previously recognized a lang:yaml , OR
B) Call both `yml` and `tf` files "lang: config"

Daniel-Sanche:
If we want to group terraform and other yaml files together on our dashboards, I'd probably vote for option B. Most YAML files will probably end up being related to GKE though, so I think there's value in keeping them separate. But I'm ok with either way

PR #494, opened 2020-10-05
Upstream/cart grpc

simonz130:
- Updated cart service to .NET 5
- Updated .Net implementation for cart service to use managed grpc hosted on asp.net core. Previously it was using unmanaged grpc stack (c-based). The inputs/outputs didn't change but the underlying stack hosting this did.
- Optimized container image size for cart service 10x from over 1.8Gb to ~189Mb.

This is deployed here: http://35.247.98.18/

This work is needed because with the merge of OpenCensus to OpenTelemetry and that .net core 3.1 being the current stack developers are using, the instrumentation in c# for the OT assumes using [managed grpc stack](https://docs.microsoft.com/en-us/aspnet/core/grpc/?view=aspnetcore-3.1).


Daniel-Sanche:
Should we close the other PR and use this one instead?

src/cartservice/Protos/Cart.proto
```diff
@@ -0,0 +1,51 @@
...
+
+package hipstershop;
+
+// -----------------Cart service-----------------
+
+service CartService {
+    rpc AddItem(AddItemRequest) returns (Empty) {}
+    rpc GetCart(GetCartRequest) returns (Cart) {}
+    rpc EmptyCart(EmptyCartRequest) returns (Empty) {}
+}
```
Daniel-Sanche, line 25: I'm going to copy this comment from the other PR: it looks like these protos are a duplicate from the ones in `/pb`. Shouldn't we use those instead?

simonz130: With new .NET managed grpc stack, the protos are defined as part of the project (e.g. cartservice.csproj) and then it's being compiled and code is autogenerated. Referencing proto file from csproj such that proto is located outside of csproj folder or subfolders will not work. It will also make it problematic when building docker container.
Unfortunately, we need to keep the duplication of proto here.

Daniel-Sanche (approved):

PR #511, opened 2020-10-07
Test outsider PR

dansanche2:
testing the new contributor bot


nicoleczhu:
I think I'll add a feature where the bot comments on the fact that it cancelled the workflows. Otherwise, externals may be confused & spend time debugging this expected behavior.

PR #517, opened 2020-10-08
ci: moved project_IDs from secrets to env variable 

nicoleczhu:
This is so we can automatically run e2e tests on renovatebot PRs


nicoleczhu:
Ah I shouldn't have named this branch e2e*. Deleting this PR and renaming this branch...

PR #518, opened 2020-10-08
ci: moved project_IDs from secrets to env variable

nicoleczhu:
This lets us automatically run e2e workflows on PRs from RenovateBot

Btw: I looked into setting envvars globally (once across github actions) but it isn't easily done, like it is with setting envvars within workflow, job, or a step. It would require us to write the envvar to Github Actions' runtime disk using one of the workflows. Which feels messy/hard to read for contributors, so I opted to set envvars per each workflow. 

Let me know any feedback!


nicoleczhu:
Pls ignore the `End to End - Latest / end-to-end-latest (push)` test. It was triggered bc I called this branch "e2econfig" and I cancelled it early.

simonz130 (approved):

PR #554, opened 2020-11-01
integrate rating service into deployment pipeline

minherz:
Rating service -- a Python3 microservice that runs on GAE SE and uses CloudSQL as its data layer was refreshed to support high loads and provide quick responses. The [TDD](https://docs.google.com/document/d/1k5Zq1BEyxhVe7Ko3_4dlLxhMXl4GC_tBp-6g1nOdnPg) document (access to Googlers only) describes the design, API, data layer and rationale behind decisions.
The sandbox pipeline is updated to include `03_ratingservice.tf` that defines provisioning of the CloudSQL instance, Postgres DB, populates it with generated ratings, creates GAE application in the sandbox project and deploys the rating service and all its configurations.
The provisioning of the rating service is disabled by default and can be enabled by adding `-var 'enable_rating_service=1'` for the `terraform apply` command. After integration of the microservice with the Hipster shop's frontend this variable will be removed.

Known issues:
- CloudSQL is provisioning having public IP. It will be converted to work with private IP after the integration work is complete.
- User password is injected as environment variable. This will be changed by leveraging secret manager service.


src/ratingservice/main.py
```diff
@@ -13,103 +13,155 @@
...
-                    'count'  : str(count)
-                })
-                resp.status_code = 200
-            else:
-                # product not exists
-                resp = jsonify({'error' : 'Product not found'.format(id)})
-                resp.status_code = 404   
-        conn.commit()    
+            cursor.execute("INSERT INTO votes (eid, rating) VALUES ('{0}', {1});".format(
+                eid, rating))
```
simonz130, line 128: won't your first parameter be prone to sql injection?

src/ratingservice/main.py
```diff
@@ -13,103 +13,155 @@
...
+        return makeError(500, 'failed to connect to DB')
     try:
         with conn.cursor() as cursor:
-            cursor.execute("INSERT INTO ratings (product_id, score) VALUES ('{0}', {1});".format(product_id, score))
-            resp = jsonify({'status' : 'success'})
-            resp.status_code = 200
+            cursor.execute("UPDATE votes SET in_process=TRUE;")
+            cursor.execute(
+                "UPDATE ratings AS r SET "
+                "rating=(r.rating*r.votes/(r.votes+v.votes))+(v.avg_rating*v.votes/(r.votes+v.votes)), "
```
simonz130, line 149: What are you doing here?

src/ratingservice/main.py
```diff
@@ -13,103 +13,155 @@
...
-# rate a product
-@app.route('/rating', methods=['POST'])
-def rate():
-    conn = connpool.getconn()
-    resp = None
-    product_id = request.form['id']
-    score = request.form['score']
+
+@app.route('/recollect', methods=['GET'])
+def aggregateRatings():
```
simonz130, line 140: please document this method thoroughly

terraform/03_ratingservice.tf
```diff
@@ -0,0 +1,94 @@
...
+      DBUSER=google_sql_user.default.name
+      DBPWD=google_sql_user.default.password
+    }
+  }
+
+  depends_on = [google_sql_database.ratings, google_sql_user.default]
+}
+
+#
+# deploy rating service to GAES
```
simonz130, line 69: please don't use abbreviations in comments

terraform/README.md
```diff
@@ -65,15 +65,16 @@ comments are available in each file, but here's the short version:
 * `01_provider.tf` -- configure the terraform provider
 * `02_project.tf` -- create a GCP project, set up billing, and enable services
 * `03_gke_cluster.tf` -- provision a GKE cluster per to the Hipster Shop README
+* `03_rankingservice.tf` -- provision and populates Postgres DB over CloudSQL and deploy ranking service to AppEngine. Creates scheduled task to aggregate new placed ratings.
```
simonz130, line 68: ratingservice.tf

terraform/ratingservice/configure_ratings_db.sh
```diff
@@ -0,0 +1,69 @@
...
+    eid char(16),
+    rating real,
+    votes integer,
+    PRIMARY KEY(eid)
+);
+CREATE TABLE IF NOT EXISTS votes(
+    id SERIAL,
+    eid char(16),
+    rating integer,
+    in_process boolean DEFAULT FALSE,
```
simonz130, line 47: why do we need this?

terraform/ratingservice/deploy_rating_service.sh
```diff
@@ -0,0 +1,56 @@
...
+#-------------------------------------------------------------------------------------------
+REGION=${REGION:-us-east1}
+VERSION=${VERSION:-prod}
+INTERVAL=${VERSION:-2}
+DELETE=${DELETE:-0}
+
+if [ $DELETE -eq 1 ]; then
+    gcloud app versions stop $VERSION --quiet # application is always deployed as default service
+    exit
+fi
```
simonz130, line 35: can we use terraform provider for app engine instead of this script?

tests/ratingservice/main_test.py
```diff
@@ -19,52 +19,60 @@
...
-            'id' : products[0]
-        })
-        res2 = requests.get(url_get).json()
-        # The original total score (count * rating) plus the current score 5 must equal to the current total score
-        self.assertTrue(abs(float(res1['rating']) * float(res1['count']) + 5 
-                        - float(res2['rating']) * float(res2['count'])) < 1e-5)
-    
+
+    # def testGetRatingCorrect(self):
+    #     """ Test if getting the rating of a product returns a correct number """
```
simonz130, line 53: are you planning to resurrect this test after it's adapted to new code?

src/ratingservice/app.template
```diff
@@ -0,0 +1,11 @@
+runtime: python38
+env: standard
+env_variables:
+    DB_HOST: '${DBHOST}'
+    DB_NAME: '${DBNAME}'
+    DB_USERNAME: '${DBUSER}'
+    DB_PASSWORD: '${DBPWD}'
+    MAX_DB_CONNECTIONS: 10
```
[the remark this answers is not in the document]
minherz: GAE application is configured to handle up to max_concurrent_requests = 10 by default. The number of connection reflects the number of requests.

src/ratingservice/main.py
```diff
@@ -13,103 +13,155 @@
...
+    print('error: environment vars DB_USERNAME, DB_PASSWORD, DB_NAME and DB_HOST must be defined.')
+    exit(1)
+
+
+def getConnection():
+    global db_connection_pool
+    if db_connection_pool == None:
+        cfg = {
+            'user': os.environ.get('DB_USERNAME'),
+            'password': os.environ.get('DB_PASSWORD'),
```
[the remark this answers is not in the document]
minherz: Yes, but it will increase a complexity of the example. Regardless, in the following steps the CloudSQL instance will not have public IP which should seriously mitigate the risk of the getting access to the instance.

terraform/ratingservice/configure_ratings_db.sh
minherz: do you ask about the table [votes] or about last field of the table? the table is used to collect new votes following CQRS model of request handling which decouples store and read operations. the last field is used to mark records that are in the process of being aggregated and added to the [ratings] from new coming new votes. this process is also described in the [TDD](https://docs.google.com/document/d/1k5Zq1BEyxhVe7Ko3_4dlLxhMXl4GC_tBp-6g1nOdnPg) that is referenced in the PR description.

terraform/ratingservice/deploy_rating_service.sh
minherz: a terraform provider does not let to define the region. if we don't care about the region, i can replace `"local-exec"` provisioner with `google_app_engine_standard_app_version` provider.

tests/ratingservice/main_test.py
minherz: yes. i replaced the constant values (rating 5 with 1 vote) by randomized rating numbers. after reviewing the goal of the test, and introducing required changes i will return it. this is one of the reasons why i drafted the PR.

Daniel-Sanche (changes requested):

Daniel-Sanche:
It looks like the e2e test is failing too. Do you think that's caused by one of your changes?

src/ratingservice/main.py
```diff
@@ -13,103 +13,146 @@
...
-            resp = jsonify({'status' : 'success'})
-            resp.status_code = 200
+            cursor.execute("UPDATE votes SET in_process=TRUE")
+            cursor.execute(
+                "UPDATE ratings AS r SET "
+                "rating=(r.rating*r.votes/(r.votes+v.votes))+(v.avg_rating*v.votes/(r.votes+v.votes)), "
+                "votes=r.votes+v.votes "
+                "FROM (SELECT eid, ROUND(AVG(rating),4) AS avg_rating, COUNT(eid) AS votes FROM votes WHERE in_process=TRUE GROUP BY eid) AS v "
+                "WHERE r.eid = v.eid")
+            cursor.execute("DELETE FROM votes WHERE in_process=TRUE")
```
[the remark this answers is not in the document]
minherz: It is due to the use of the cron job feature within App Engine that supports only GET requests. The goal was to limit use of different GCP services by one microservice to required minimum. The original version used PATCH verb but apparently neither Scheduled Tasks nor Cron Jobs of the App Engine support PATCH verb.

src/ratingservice/main.py
```diff
@@ -13,103 +13,146 @@
...
         conn.commit()
+        return makeResult({})
```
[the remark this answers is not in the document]
minherz: The status is returned using HTTP Status codes. The payload is used to return the data which, in this case, is none.

minherz:
@Daniel-Sanche , the error that I see in logs show that it fails due incorrect command However, if the ratingservice_test.py runs it will fail because the e2e workflow does not deploy the rating microservice in the first place. I plan to work next on adding the microservice and Postgres into e2e environment as pods.

src/ratingservice/main.py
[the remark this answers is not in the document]
minherz: I've moved provisioning from using AppEngine cron job to Scheduler Job. The verb is changed to PUT because Scheduler does not support PATCH.

Daniel-Sanche:
> yes, it looks like the root error is due to conditional variable that I placed to avoid rating microservice provisioning after the merge and until the integration is complete. i fixed it in [afba894](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/commit/afba8944450bdc8568216cda3369cee1227cff28). need to see if there is other issues before i enable it for e2e workflow so the rating microservice will get tested.

Sounds good, let me know when you're ready for another review!

minherz:
@Daniel-Sanche, while e2e still fails, the inspection of the log shows no relation to the changes in this PR. please, review the fixes following your previous feedback.
thank you.

Daniel-Sanche:
> @Daniel-Sanche, while e2e still fails, the inspection of the log shows no relation to the changes in this PR. please, review the fixes following your previous feedback.
> thank you.

Hmm it seems like the CI failure is specific to your branch, although I'm not quite sure why that is. The [E2E test against master](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/runs/1381611539?check_suite_focus=true) is passing. I also ran a custom [Open in Cloud Shell button](https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/GoogleCloudPlatform/cloud-ops-sandbox.git&cloudshell_git_branch=minherz/ratingservice&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image/uncertified:v0.4.0&cloudshell_tutorial=docs/tutorial.md) pointed at your branch, and saw errors in the same place. (If you haven't seen this, it's a good way to test. you can run a full sandbox installation using any branch/cloudshell image)

I'm not sure why your branch would be causing issues for the istio monitoring services though. Maybe try syncing with master to make sure you're up to date?

src/ratingservice/main.py
Daniel-Sanche: +1, this could use some commenting at very least

Daniel-Sanche:
I also didn't seem to see any new GAE/Cloud SQL services set up after running  [Open in Cloud Shell button](https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/GoogleCloudPlatform/cloud-ops-sandbox.git&cloudshell_git_branch=minherz/ratingservice&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image/uncertified:v0.4.0&cloudshell_tutorial=docs/tutorial.md). Is that expected?

src/ratingservice/main.py
```diff
@@ -13,103 +13,180 @@
...
+
+    data = request.get_json()
+    if data == None:
+        return makeError(400, "missing json payload")
+
+    eid = str(data['id'])
+    if eid == None or eid == "":
+        return makeError(400, "malformed entity id")
+    rating_str = str(data['rating'])
+    if rating_str == None or rating_str == "":
```
Daniel-Sanche, line 133: nit: `if not rating_str` should cover both of these cases

src/ratingservice/main.py
```diff
@@ -13,103 +13,180 @@
...
+    '''
+
+    data = request.get_json()
+    if data == None:
+        return makeError(400, "missing json payload")
+
+    eid = str(data['id'])
+    if eid == None or eid == "":
+        return makeError(400, "malformed entity id")
+    rating_str = str(data['rating'])
```
Daniel-Sanche, line 132: It looks like there's potential for a KeyError here if 'rating' isn't in data

src/ratingservice/main.py
```diff
@@ -13,103 +13,180 @@
...
+    if rating_str == None or rating_str == "":
+        return makeError(400, "malformed rating")
+    if not rating_str.isdigit():
+        return makeError(400, "rating should be non negative number")
+    rating = int(rating_str)
```
Daniel-Sanche, line 137: You are assuming here that everything that passes `isdigit` will be parsable by `int()`. If the string is something like `"²"`, a ValueError would be thrown here

in general, it would be better to put more of this parsing code in try blocks, and return errors when an exception is thrown, rather than trying to validate types manually

src/ratingservice/main.py
```diff
@@ -13,103 +13,180 @@
...
+    if rating < 1 or rating > 5:
+        return makeError(400, "rating should be non negative number")
```
Daniel-Sanche, line 139: This error seems misleading if it only accepts 1-5

src/ratingservice/app.template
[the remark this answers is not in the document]
minherz: The number of threads per each process is defined to 10 (see entrypoint field in the [app.template](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/minherz/ratingservice/src/ratingservice/app.template))

src/ratingservice/main.py
minherz: fixed and refactored

src/ratingservice/main.py
minherz: refactored

src/ratingservice/main.py
minherz: message changed

minherz:
@Daniel-Sanche I have looked into e2e - latest actions for master branch to compare with what is happening in this PR. There are three differences that influence the execution:
1) Cloud Shell image has additional jq and psql packages
2) cloudscheduler, sql-component and sqladmin APIs are activated in addition to existing APIs
3) output variable is defined
There are no other changes that might cause a change in the behaviour because python tests under /tests/* folders aren't executed by the e2e latest workflow. I am going to remove the output variable -- it was wrong to add it to PR but I do not expect it  to make the difference.

I placed a conditional variable for provisioning rating service resources which isn't set by default. i believe this is the reason why Cloud Shell does not provision CloudSQL and AppEngine. If you think it is right way to provision these resources even if they aren't used/integrated, I will change the script by removing the conditional variable. LMK WDYT

Daniel-Sanche:
Looking at the other e2e tests, it looks like other branches are seeing this issue too. I'm not sure what's causing it, because it's not consistent either. I'll have to keep looking into it, but for now you can assume it's unrelated to your work. You can comment out [this line](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/640364722e8499650ad1ebe218447868a0077e8c/terraform/install.sh#L251) for now to see if the rest of the pipeline passes

Still though, I am not seeing your new services spun up when installing with the [Open in Cloud Shell link](https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/GoogleCloudPlatform/cloud-ops-sandbox.git&cloudshell_git_branch=v0.4.0&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image/uncertified:v0.4.0&cloudshell_tutorial=docs/tutorial.md). What are you expecting to see as part of this PR?

cloud-shell/Dockerfile
```diff
@@ -20,6 +20,9 @@ RUN sudo apt-get install unzip && \
   unzip -o terraform.zip && \
   sudo install terraform /usr/local/bin
 
+# install Postgres client and Json query
+RUN sudo apt-get -y install postgresql-client jq
```
Daniel-Sanche, line 24: jq looks to already be pre-installed on cloudshell instances. Are you sure postgresql-client is needed?

minherz: `psql` is used to populate DB with initial rating data. it allows to keep the code of the service clean from any deployment and configuration tasks.

minherz:
The cause to timeout is that Cloud Monitoring doesn't have istio services. It makes sense because the services are created as result of the execution of the terraform on /terraform/monitoring. This doesn't happened in e2e latest if launching it on a clean project.
Does your comment mean that if I comment out [the line](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/640364722e8499650ad1ebe218447868a0077e8c/terraform/install.sh#L251) and the test pass, I can merge?

Answering your question about the goal of the PR, it is to place the working code of the microservice and the provisioning scripts to main branch, so the next step will deal only with integration.

Daniel-Sanche:
@minherz Actually, I think I found the issue today. Before commenting out, try pulling [this change](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/563) into you branch and see if it makes the test pass.

> Does your comment mean that if I comment out the line and the test pass, I can merge?

Yes, I'm quite sure the issue is unrelated to your code, so I'd be fine with merging as long as everything else works well. I'll just do a final review first, and then add my approval

cloud-shell/Dockerfile
minherz: You are right. `jq` is part of the cloud-shell image. I will remove it in the following PR.

minherz:
@googlebot I consent.

terraform/ratingservice/deploy_rating_service.sh
Daniel-Sanche: It sounds like this is probably out of scope of this PR since integration is coming later, but we may have to discuss this some more. One of our requirements is we want users to be able to run the install script multiple times to re-attach to existing projects. That's easy with terraform, since it keeps the current state of the system. If we have some parts created by terraform and some through scripts, it gets a bit complicated.

I'm not too familiar with terraform myself, but it's surprising that it would be flaky on AppEngine. That's one of the oldest GCP services

Daniel-Sanche (approved):
There's still some open questions around testing and deployment, but this part LGTM!

minherz:
I will address these in the following PRs

terraform/ratingservice/deploy_rating_service.sh
minherz: I tried to use pure terraform resources (from Google provider) to install the service on AppEngine. This requires ~10 resources instead of one and does not work reliably. I agree that what can be done with terraform should work this way, but the fact is that you cannot do the same thing with 1-2 terraform resources. With that, I am ready to get it another try, maybe using Google beta provider to see if it works more reliable.

PR #556, opened 2020-11-06
chore: add blunderbuss

nicoleczhu:


.github/blunderbuss.yml
```diff
@@ -0,0 +1,4 @@
+assign_issues:
+  - GoogleCloudPlatform/observability-devx
```
simonz130, line 2: How does it work? Will it assign issues with round robin logic to this group members?

Daniel-Sanche:
looks like you'll need to run [addlicense](https://github.com/google/addlicense) to pass the CI

.github/blunderbuss.yml
nicoleczhu: Correct, it should round robin to this group's members. Same as the repos with `googleapis/api-logging` team designated

simonz130 (approved):

PR #557, opened 2020-11-09
fix: e2e release tests

Daniel-Sanche:
- currently, E2E tests against the latest release are failing. It looks like it's incorrectly parsing the clodushell image name. This PR fixes that issue
- the scheduled cronjob seems to be spawning many jobs each day. According to [crontab.guru](https://crontab.guru/#*_20_*_*_*), our old config was asking for "every minute during the hour of 8", so that could be the reason. Changed to "once every day at exactly 8", we'll see if this fixes it


nicoleczhu (approved):
Ahh niiice, the cron job wildcard fix was a good find. Can't believe I overlooked it

Daniel-Sanche:
> Ahh niiice, the cron job wildcard fix was a good find. Can't believe I overlooked it

Same here!

PR #578, opened 2020-11-25
Extend rating service API with GET ratings/ call to simplify integration with frontend

minherz:
add GET ratings/ method that returns _all_ ratings to call when showing frontend homepage.
add e2e test for the new API.
increase e2e test timeout to 5 seconds per API call.

### DIFF
--- a/src/ratingservice/main.py
+++ b/src/ratingservice/main.py
@@ -103,7 +103,7 @@ def getRatings():
             })
         else:
             return makeError(500, 'No available ratings')
-    except:
+    except DatabaseError:
         return makeError(500, 'DB error')
     finally:
         db_connection_pool.putconn(conn)
@@ -143,7 +143,7 @@ def getRatingById(eid):
             })
         else:
             return makeError(404, "invalid entity id")
-    except:
+    except DatabaseError:
         return makeError(500, 'DB error')
     finally:
         db_connection_pool.putconn(conn)
@@ -191,7 +191,7 @@ def postRating():
         return makeResult({})
     except IntegrityError:
         return makeError(404, 'invalid entity id')
-    except:
+    except DatabaseError:
         return makeError(500, 'DB error')
     finally:
         db_connection_pool.putconn(conn)
@@ -220,7 +220,7 @@ def aggregateRatings():
             cursor.execute("DELETE FROM votes WHERE in_process=TRUE")
         conn.commit()
         return makeResult({})
-    except:
+    except DatabaseError:
         return makeError(500, 'DB error')
     finally:
         db_connection_pool.putconn(conn)
--- a/src/ratingservice/main.py
+++ b/src/ratingservice/main.py
@@ -68,6 +68,7 @@ def makeError(code, message):
 
 
 def makeResult(data):
+    print("2", data)
     result = jsonify(data)
     result.status_code = 200
     return result
@@ -75,6 +76,43 @@ def makeResult(data):
 #
 # APIs
 #
+
+
+@app.route('/ratings', methods=['GET'])
+def getRatings():
+    '''Gets a list of all ratings.
+
+    Returns:
+        HTTP status 200 and Json payload { ratings: [{'id': (string), 'rating': (number)}] }
+        HTTP status 500 when there is an error querying DB or no data
+    '''
+
+    conn = getConnection()
+    if conn == None:
+        return makeError(500, 'failed to connect to DB')
+    try:
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT eid, ROUND(rating,4) FROM ratings")
+            result = cursor.fetchall()
+        conn.commit()
+        print("1")
+        if result != None:
+            print("1")
+            ratings = []
+            print("1")
+            for row in result:
+                # cast to float because flas.jsonify doesn't work with decimal
+                ratings.append({"id": row[0].strip(), "rating": float(row[1])})
+            print("1")
+            return makeResult({
+                'ratings': ratings,
+            })
+        else:
+            return makeError(500, 'No available ratings')
+    except:
+        return makeError(500, 'DB error')
+    finally:
+        db_connection_pool.putconn(conn)
 
 
 @app.route('/rating/<eid>', methods=['GET'])
--- a/src/ratingservice/main.py
+++ b/src/ratingservice/main.py
@@ -68,7 +68,6 @@ def makeError(code, message):
 
 
 def makeResult(data):
-    print("2", data)
     result = jsonify(data)
     result.status_code = 200
     return result
@@ -95,15 +94,10 @@ def getRatings():
             cursor.execute("SELECT eid, ROUND(rating,4) FROM ratings")
             result = cursor.fetchall()
         conn.commit()
-        print("1")
-        if result != None:
-            print("1")
-            ratings = []
-            print("1")
-            for row in result:
-                # cast to float because flas.jsonify doesn't work with decimal
-                ratings.append({"id": row[0].strip(), "rating": float(row[1])})
-            print("1")
+        if result is not None:
+            # cast to float because flask.jsonify doesn't work with decimal
+            ratings = [{"id": eid.strip(), "rating": float(rating)}
+                       for (eid, rating) in result]
             return makeResult({
                 'ratings': ratings,
             })
--- a/tests/ratingservice/main_test.py
+++ b/tests/ratingservice/main_test.py
@@ -18,63 +18,101 @@ import os
 import sys
 import unittest
 import requests
-from requests.adapters import HTTPAdapter
+from requests.exceptions import RequestException
+from urllib3.exceptions import ReadTimeoutError
 import json
 import decimal
 
 
-service_url = ''
-products = []
-TIMEOUT = 5.0
-
-
-def composeUrl(resource, eid=""):
-    if not eid:
-        return "{0}/{1}".format(service_url, resource)
-    else:
-        return "{0}/{1}/{2}".format(service_url, resource, eid)
-
-
 class TestEndpoints(unittest.TestCase):
 
+    @classmethod
+    def composeUrl(cls, resource, eid=""):
+        if not eid:
+            return "{0}/{1}".format(cls.service_url, resource)
+        else:
+            return "{0}/{1}/{2}".format(cls.service_url, resource, eid)
+
+    @classmethod
+    def setUpClass(cls):
+        cls.TIMEOUT = 5.0
+
+        cls.service_url = ''
+        if len(sys.argv) > 1:
+            cls.service_url = sys.argv[1].rstrip('/')
+
+        cls.products = []
+        path = os.getcwd().rstrip('/') + '/src/productcatalogservice/products.json'
+        if len(sys.argv) > 2:
+            path = sys.argv[2]
+        try:
+            with open(path) as f:
+                data = json.load(f)
+                for product in data['products']:
+                    cls.products.append(product['id'])
+        except:
+            print("failed to load product ids from ", path)
+            cls.products = []
+
     def setUp(self):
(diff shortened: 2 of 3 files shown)


tests/ratingservice/main_test.py
```diff
@@ -44,25 +45,36 @@ def setUp(self):
...
+        self.assertEqual(len(ratings), len(products))
+        ids = list(map(lambda r: r['id'], ratings))
+        self.assertTrue(all(p in products for p in ids))
+
+    def testGetRatingPerProduct(self):
+        """ test getting ratings for each shop product """
         for product in products:
             url = composeUrl("rating", product)
-            res = self.session.get(url, timeout=1.0)
+            res = self.session.get(url, timeout=TIMEOUT)
```
Daniel-Sanche, line 63: I'm not sure if I like how the products list is dynamically loaded. If there's an error and products isn't loaded properly, some of these tests will mistakenly pass. A couple thoughts:
- can we hardcode the product list with the expected data?
- If not, `products` should probably at least be loaded in the setup() function instead of `__main__`
- There should probably at least be some assertions on `products` after it is loaded, to make sure it has reasonable data

src/ratingservice/main.py
```diff
@@ -68,6 +68,7 @@ def makeError(code, message):
 
 
 def makeResult(data):
+    print("2", data)
```
Daniel-Sanche, line 71: Looks like a debugging string was accidentally committed

src/ratingservice/main.py
```diff
@@ -77,6 +78,43 @@ def makeResult(data):
...
+        conn.commit()
+        print("1")
+        if result != None:
+            print("1")
+            ratings = []
+            print("1")
+            for row in result:
+                # cast to float because flas.jsonify doesn't work with decimal
+                ratings.append({"id": row[0].strip(), "rating": float(row[1])})
+            print("1")
```
Daniel-Sanche, line 106: More print statements to remove

src/ratingservice/main.py
```diff
@@ -77,6 +78,43 @@ def makeResult(data):
...
+            cursor.execute("SELECT eid, ROUND(rating,4) FROM ratings")
+            result = cursor.fetchall()
+        conn.commit()
+        print("1")
+        if result != None:
+            print("1")
+            ratings = []
+            print("1")
+            for row in result:
+                # cast to float because flas.jsonify doesn't work with decimal
```
Daniel-Sanche, line 104: nit: typo on flask

src/ratingservice/main.py
```diff
@@ -77,6 +78,43 @@ def makeResult(data):
...
+    conn = getConnection()
+    if conn == None:
+        return makeError(500, 'failed to connect to DB')
+    try:
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT eid, ROUND(rating,4) FROM ratings")
+            result = cursor.fetchall()
+        conn.commit()
+        print("1")
+        if result != None:
```
Daniel-Sanche, line 99: nit: `if result is not None` [is preferred](https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none)

Or `if not ratings` if you want empty lists to also skip this block

src/ratingservice/main.py
```diff
@@ -77,6 +78,43 @@ def makeResult(data):
...
+            print("1")
+            ratings = []
+            print("1")
+            for row in result:
+                # cast to float because flas.jsonify doesn't work with decimal
+                ratings.append({"id": row[0].strip(), "rating": float(row[1])})
```
Daniel-Sanche, line 105: this for loop could be replaced with a list comprehension: 
`ratings = [{"id": id.strip(), "rating": float(rating)} for (float, rating) in result]

src/ratingservice/main.py
```diff
@@ -77,6 +78,43 @@ def makeResult(data):
...
+            print("1")
+            return makeResult({
+                'ratings': ratings,
+            })
+        else:
+            return makeError(500, 'No available ratings')
+    except:
```
Daniel-Sanche, line 112: are there specific exceptions you can catch here?

Daniel-Sanche (changes requested):

tests/ratingservice/main_test.py
```diff
@@ -44,25 +45,36 @@ def setUp(self):
...
-        """ test getting ratings for all shop products """
+    def testGetAllRatings(self):
+        """ test getting all ratings """
+        url = composeUrl("ratings")
+        res = self.session.get(url, timeout=TIMEOUT)
+        self.assertEqual(res.status_code, 200)
+        ratings = res.json().get('ratings')
+        self.assertTrue(type(ratings) == list)
+        self.assertEqual(len(ratings), len(products))
+        ids = list(map(lambda r: r['id'], ratings))
```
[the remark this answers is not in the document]
minherz: will do

tests/ratingservice/main_test.py
```diff
@@ -44,25 +45,36 @@ def setUp(self):
...
+        self.assertTrue(all(p in products for p in ids))
```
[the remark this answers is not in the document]
minherz: nit. thank you. will do.

tests/ratingservice/main_test.py
minherz: i do not follow. are you referencing the fact that we already test the same in the previous test method? the matter is that another method was just added. before that there was no way to test that all shop products have rating.
now, we can reduce the test for GET /product/<id> for testing a custom product from the list.
answering your questions:
* product list is hardcoded but not the ratings. the ratings are generated at each deployment. we test rating correctness in the last test which includes both the calculation and the correctness.
* list of products is loaded from the products.json at the beginning of the test. if you can explain the difference between the exact place where the code for loading the products is located, it will help me to understand your arguments.
* this last bullet is unclear. the products are loaded from the same source from which the shop loads it. there should be no need for additional assertion i think.

src/ratingservice/main.py
minherz: OMG. will be removed

src/ratingservice/main.py
minherz: :)

src/ratingservice/main.py
minherz: will be removed

src/ratingservice/main.py
minherz: thanx. 👍

src/ratingservice/main.py
minherz: it is generally catches database errors starting with connection error and ending with sql and casting. i do not know if more detailed handling is required. prefer to leave it generic at the moment.

tests/ratingservice/main_test.py
Daniel-Sanche: > list of products is loaded from the products.json at the beginning of the test

Ah ok, when I skimmed the code I thought it was querying it from the service. You can ignore my first point then

> list of products is loaded from the products.json at the beginning of the test.

For my second point, if I'm reading it right, products is populated by `getProducts()` in part of `if __name__ == '__main__'`. This means it will be populated if you run the script directly, but if you call it using pytest (ex, by running `pytest ./`), that `__main__` block will be skipped, right?

> this last bullet is unclear. the products are loaded from the same source from which the shop loads it. there should be no need for additional assertion i think.

For my third bullet, I was pointing out that your tests take as a given that products contains items, but it is initialized as an empty array. Some tests will silently pass if products is never populated, and others will fail in unexpected places. It's probably worthwhile to add an assertion at the start of each test to make sure products contains some data

minherz: ok, thanx. regarding the pytest, it will not work because it expects to receive arguments and will fail given the arguments aren't provided, the arguments are the root URL of the service and the path to products.json. also the second can be defaulted. what is a way to pass the arguments using pytest? is it a blocker for this PR?
for the last bullet, the comparison goes the returned list vs the list from the product.json. The second list cannot be empty, hence the test will fail if the returned list is empty.

src/ratingservice/main.py
Daniel-Sanche: generic catches are a bad practice, you should really add some of the exceptions you expect to catch here. Right now your block will catch exceptions like `KeyboardInterrupt` and `SystemExit`, which could cause some strange behaviours

tests/ratingservice/main_test.py
minherz: can i mark it resolved given i will take care to support pytest in one of further PRs?

src/ratingservice/main.py
minherz: i agree. with that it is not a real application and there is no practical reason to threat each exception separately.

tests/ratingservice/main_test.py
Daniel-Sanche: > is the use pytest is a hard requirement at the moment? i see that most of the tests use unittest and run by python.
it might be better to define a migration to pytest as a separate task and refactor all the modules at once. wdyt?

Is there a reason you can't just put `getProducts` in a set up function instead of as part of the `__main__` function? That's the standard practice for python tests and it seems like a simple enough change. Are there downsides I'm not thinking of?

> for the last bullet, the test compares the returned list vs. the list read from the product.json file. The second list cannot be empty, hence the test will fail if the returned list is empty.

You're assuming the second list can't be empty ;-) I think it makes sense to assert those assumptions as part of the test. If the `__main__` function isn't executed, or the the product.json file is replaced with an empty file, `products` will be empty, so I think it's reasonable to add an assertion to the top of the test to check for that. Is there a reason you'd rather not?

src/ratingservice/main.py
Daniel-Sanche: I'd rather invest the time now in code review, rather than more time later to hunt down bugs.

This is the [exception inheretance tree for psycopg2](https://www.psycopg.org/docs/module.html):
```
Exception
|__ Warning
|__ Error
    |__ InterfaceError
    |__ DatabaseError
        |__ DataError
        |__ OperationalError
        |__ IntegrityError
        |__ InternalError
        |__ ProgrammingError
        |__ NotSupportedError
```

The documentation says it's sufficient to just catch [`psycopg2.Error`](https://www.psycopg.org/docs/module.html#psycopg2.Error) to handle any psycopg2 errors. If you really want to be safe, you could even just catch the generic Python `Exception`.  That's also a bad practice, but it would at least make sure the program won't swallow exit signals

minherz: Thanx. I will use DatabaseError then.

tests/ratingservice/main_test.py
minherz: this is a very long thread. let's have a quick sync over GVC to talk about these items.

minherz: I've looked into python unittest options and used `setUpClass` to initiate the values and fail() to handle possible problems when reading products and/or passing service URL.
No code except for
```python
if __name__ == '__main__':
    unittest.main(argv=['first-arg-is-ignored'])
```
is not defined outside the class.

tests/ratingservice/main_test.py
```diff
@@ -18,76 +18,126 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.TIMEOUT = 5.0
+
+        cls.service_url = ''
+        if len(sys.argv) > 1:
+            cls.service_url = sys.argv[1].rstrip('/')
+
+        cls.products = []
+        path = os.getcwd().rstrip('/') + '/src/productcatalogservice/products.json'
```
Daniel-Sanche, line 45: nit: comments would be useful here. You may even want to keep the `getProducts`, whatever you think is most readable

Daniel-Sanche (approved):

tests/ratingservice/main_test.py
minherz: I will add commenting 👍

PR #581, opened 2020-12-02
integrate frontend service with rating service

minherz:
Setup HTML templates to have dynamic values of the rating.
Modify home page and product page handlers to populate product rating(s).
Add functionality to submit a new product rating.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,7 @@ jobs:
         set -x
         skaffold config set --global local-cluster true
         skaffold run -p local --status-check=false
+        kubectl set env deployment.apps/frontend RATING_SERVICE_ADDR=http://some.host
     - name: Wait For Pods
       timeout-minutes: 10
       run: |
--- a/src/frontend/handlers.go
+++ b/src/frontend/handlers.go
@@ -22,8 +22,8 @@ import (
 	"net/http"
 	"os"
 	"strconv"
-	"time"
 	"strings"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
@@ -53,6 +53,12 @@ func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
 		renderHTTPError(log, r, w, errors.Wrap(err, "could not retrieve products"), http.StatusInternalServerError)
 		return
 	}
+	ratings, err := fe.getAllRatings(r.Context())
+	if err != nil {
+		renderHTTPError(log, r, w, errors.Wrap(err, "could not retrieve ratings"), http.StatusInternalServerError)
+		return
+	}
+
 	cart, err := fe.getCart(r.Context(), sessionID(r))
 	if err != nil {
 		renderHTTPError(log, r, w, errors.Wrap(err, "could not retrieve cart"), http.StatusInternalServerError)
@@ -60,8 +66,9 @@ func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	type productView struct {
-		Item  *pb.Product
-		Price *pb.Money
+		Item   *pb.Product
+		Price  *pb.Money
+		Rating float64
 	}
 	ps := make([]productView, len(products))
 	for i, p := range products {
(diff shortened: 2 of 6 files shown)


minherz:
If you want to try hands-on experience of the integration, you can access a test environment using http://104.198.37.65/.
I will shutdown the environment after the review process is over.

Daniel-Sanche (commented):
I'm not too familiar with go, but overall this looks good to me. Left a couple small comments

src/frontend/rest.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		return errors.Errorf("status: %d, message: %s", result.StatusCode, errData.Error)
+	} else if result.StatusCode == 200 {
+		err := json.Unmarshal(body, response)
+		if err != nil {
+			return errors.Wrap(err, "json unmarshal failed for : "+string(body))
+		}
+		return nil
+	} else {
+		return errors.Errorf("unhandled response code %q", result.StatusCode)
```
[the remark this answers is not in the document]
minherz: A question is what would be the right behavior regarding payload processing. Should I try to read and parse the payload the same way for all 2xx? Since many 3xx statuses imply redirect, should the code support?
WDYT to handle all 2xx responses by trying to parse the response and to keep 3xx handled the same way?

.github/workflows/ci.yml
```diff
@@ -59,6 +59,7 @@ jobs:
         set -x
         skaffold config set --global local-cluster true
         skaffold run -p local --status-check=false
+        kubectl set env deployment.apps/frontend RATING_SERVICE_ADDR=http://some.host
```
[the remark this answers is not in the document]
minherz: Thanx. Will do.

src/frontend/handlers.go
```diff
@@ -319,6 +338,25 @@ func (fe *frontendServer) placeOrderHandler(w http.ResponseWriter, r *http.Reque
...
+// postNewRatingHandler posts new rating vote for product.
+// It ignores the error responses on the vote request since it is not critical.
+func (fe *frontendServer) submitRatingHandler(w http.ResponseWriter, r *http.Request) {
+	log := r.Context().Value(ctxKeyLog{}).(logrus.FieldLogger)
+	productID := mux.Vars(r)["id"]
+	if productID == "" {
+		renderHTTPError(log, r, w, errors.New("product id not specified"), http.StatusBadRequest)
+		return
+	}
+	rating, _ := strconv.ParseInt(r.FormValue("rating"), 10, 32)
```
simonz130, line 350: what is the meaning of 10 and 32 here? Perhaps, worth adding a comment if GO doesn't support named parameters.

src/frontend/rest.go
```diff
@@ -0,0 +1,158 @@
...
+func buildURL(host, resource, id string) (string, error) {
+	uri, err := url.ParseRequestURI(host)
+	if err != nil {
+		return "", errors.Errorf("%q is invalid URL", host)
+	}
+	uri.Path = path.Join(uri.Path, resource, id)
+	return uri.String(), nil
+}
+
+func sendRestRequest(ctx context.Context, method, url string, payload io.Reader, response interface{}) error {
```
simonz130, line 66: document this method

src/frontend/rest.go
```diff
@@ -0,0 +1,158 @@
...
+		if err != nil {
+			return errors.Wrap(err, "json unmarshal failed for : "+string(body))
+		}
+		return nil
+	} else {
+		return errors.Errorf("unhandled response code %q", result.StatusCode)
+	}
+}
+
+func (fe *frontendServer) getAllRatings(ctx context.Context) (map[string]float64, error) {
```
simonz130, line 108: please document this and other methods

simonz130 (approved):

src/frontend/handlers.go
minherz: I just copy/pasted it from another method in this file:

https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/6eb646640ec7f6f1965329f18cf8ff5ca401e700/src/frontend/handlers.go#L174-L176

Looked for this in the internet and it seems to be a usual [practice](https://golang.org/pkg/strconv/).

Should we deal with this?

src/frontend/rest.go
minherz: Will be done in next PR

src/frontend/rest.go
minherz: Will be done in next PR

PR #586, opened 2020-12-06
add rating service monitoring metrics

minherz:
Add a dashboard for monitoring service with the following metrics:
- responses by HTTP code
- 95th percentile of the response latency
- instance count by state
- throttled responses
Add availability (99% of non-4XX and 429 responses return 200) and latency (99% of 200 responses are returned within 150ms) SLOs for 30 window.
Refactor .gitignore to ignore terraform files under /terraform/monitoring.
Some formatting was committed due to use of VSCode for editing.

### DIFF
--- a/src/frontend/rest.go
+++ b/src/frontend/rest.go
@@ -63,6 +63,8 @@ func buildURL(host, resource, id string) (string, error) {
 	return uri.String(), nil
 }
 
+// sendRestRequest abstracts sending GET and POST requests with a timeout, parsing response and error handling.
+// It parses the returned Json into response. Any response with code that is not 200 or 4xx results in error.
 func sendRestRequest(ctx context.Context, method, url string, payload io.Reader, response interface{}) error {
 	ctx, cancel := context.WithTimeout(ctx, time.Millisecond*500)
 	defer cancel()
@@ -105,6 +107,8 @@ func sendRestRequest(ctx context.Context, method, url string, payload io.Reader,
 	}
 }
 
+// getAllRatings sends GET /ratings request to rating service.
+// It returns a map of product ids to ratings.
 func (fe *frontendServer) getAllRatings(ctx context.Context) (map[string]float64, error) {
 	result := make(map[string]float64)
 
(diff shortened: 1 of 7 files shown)


minherz:
@Daniel-Sanche I am confused with the error that I see in ci workflow. It seems like smth external to this PR. Can you please have a look and let me know your opinion?

src/frontend/rest.go
```diff
@@ -63,6 +63,8 @@ func buildURL(host, resource, id string) (string, error) {
 	return uri.String(), nil
 }
 
+// sendRestRequest abstracts sending GET and POST requests with a timeout, parsing response and error handling.
```
simonz130, line 66: nit: no need to repeat the method name in the comment (here and throughout)

terraform/monitoring/04_slos.tf
```diff
@@ -166,4 +166,79 @@ resource "google_monitoring_slo" "istio_service_latency_slo" {
...
+      # The "good" service is the number of 200 OK responses
+      good_service_filter = join(" AND ", [
+        "metric.type=\"appengine.googleapis.com/http/server/response_count\"",
+        "resource.type=\"gae_app\"",
+        "resource.label.\"version_id\"=\"prod\"",
+        "resource.label.\"module_id\"=\"ratingservice\"",
+        "metric.label.\"response_code\"=\"200\""
+      ])
+
+      # The total is the number of non-4XX and 429 responses
```
simonz130, line 197: Please update the comment to mention what is 429 HTTP code - this is mostly because a variety of folks might use this terraform script as a reference documentation on how to implement it for their projects, so it should be as clear as possible

terraform/monitoring/dashboards/ratingservice_dashboard.json
```diff
@@ -0,0 +1,129 @@
...
+          ],
+          "timeshiftDuration": "0s",
+          "yAxis": {
+            "label": "y1Axis",
+            "scale": "LINEAR"
+          }
+        }
+      },
+      {
+        "title": "HTTP Denied From Over Quota",
```
simonz130, line 98: Is this something you took from other examples? I wonder if it's applicable here?

simonz130 (approved):

terraform/monitoring/dashboards/ratingservice_dashboard.json
minherz: My idea is that given the load generator it is possible that requests to App Engine will be throttled due to quota. Having this stats should help in event SLO will be down due to number of 429 responses.

terraform/monitoring/04_slos.tf
minherz: Will do

src/frontend/rest.go
minherz: I followed the Go Doc pattern for public methods. Maybe it is excessive. 😞

Issue #601, opened 2020-12-15
Loadgenerator is not sending traffic automatically


Daniel-Sanche:
It seems like the loadgenerator is deploying into the ready state, but not sending traffic until the user turns it on manually. I'm only seeing health checks show up in Trace until the loadgenerator is turned on

PR #609, opened 2020-12-15
feat: run loadgen for a few seconds during installation

nicoleczhu:
fix: #601 
Run load generator for ~10-15s during installation. 
Adds ~5s to previous install time. 

**Load specs:** 
user_count=50
spawn_rate=10
At the end of 5 seconds we delete users that have already been created. 

**Note:** 
There's a race condition where `curl http://loadgen_ip/stop` to stop loadgen returns a successful http response (and UI indicates the swarm has terminated..) 

BUT this only terminates users that have already been spawned, not the users that were originally scheduled to spawn! [Bug filed with Locust here](https://github.com/locustio/locust/issues/1661). 

For us this might be nice as sandbox users will continue to get traces/metrics from any undeleted remaining users from Locust.

**Context:** 
We wanted to provide initial traffic telemetry for users to view && still provide Locust loadgen UI. 
Implemented the workaround suggested by this [Locust issue](https://github.com/locustio/locust/issues/831)

**Testing:** 
In Span Viewer, use filter `/http/user_agent:python-requests`

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -284,11 +284,10 @@ loadGen() {
   gcloud container clusters get-credentials loadgenerator --zone "$LOADGEN_ZONE"
   kubectx loadgenerator=.
 
-  # Kicks of initial load generation via a POST request to Locust web interface
+  # Kicks off initial load generation via a POST request to Locust web interface
   # This is because Locust currently doesn't support CLI and UI at the same time
   TRIES=0
-#  while [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip" -o /dev/null --max-time 1) -ne 200  && \
-  while [[ $(curl -XPOST -d "user_count=100&spawn_rate=10" http://$loadgen_ip/swarm -o /dev/null -w "%{http_code}" --max-time 1) -ne 200  && \
+  while [[ $(curl -XPOST -d "user_count=10&spawn_rate=2" http://$loadgen_ip/swarm -o /dev/null -w "%{http_code}" --max-time 1) -ne 200  && \
       "${TRIES}" -lt 20  ]]; do
     log "waiting for load generator instance..."
     sleep 10
@@ -435,4 +434,3 @@ if [[ -z "${skip_loadgen}" ]]; then
   loadGen;
 fi
 displaySuccessMessage;
-
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -177,11 +177,11 @@ createProject() {
           exit 0;
         fi
       done
-      folder_id="470827991545" # /cloud-ops-sandboxes  
-      gcloud projects create "$project_id" --name="Cloud Operations Sandbox Demo" --folder="$folder_id"    
+      folder_id="470827991545" # /cloud-ops-sandboxes
+      gcloud projects create "$project_id" --name="Cloud Operations Sandbox Demo" --folder="$folder_id"
     else
       sendTelemetry $project_id new-sandbox-non-googler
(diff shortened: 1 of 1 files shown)


terraform/install.sh
```diff
@@ -283,16 +284,26 @@ loadGen() {
   gcloud container clusters get-credentials loadgenerator --zone "$LOADGEN_ZONE"
   kubectx loadgenerator=.
 
-  # find the IP of the load generator web interface
+  # Kicks of initial load generation via a POST request to Locust web interface
+  # This is because Locust currently doesn't support CLI and UI at the same time
   TRIES=0
-  while [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip" -o /dev/null --max-time 1) -ne 200  && \
+#  while [[ $(curl -sL -w "%{http_code}"  "http://$loadgen_ip" -o /dev/null --max-time 1) -ne 200  && \
```
Daniel-Sanche, line 290: should this comment stick around?

terraform/install.sh
```diff
@@ -283,16 +284,26 @@ loadGen() {
...
+  while [[ $(curl -XPOST -d "user_count=100&spawn_rate=10" http://$loadgen_ip/swarm -o /dev/null -w "%{http_code}" --max-time 1) -ne 200  && \
```
Daniel-Sanche, line 291: is /swarm idempotent? Can we keep trying to hit it multiple times and always expect a 200 response?

terraform/install.sh
```diff
@@ -283,16 +284,26 @@ loadGen() {
...
     TRIES=$((TRIES + 1))
   done
+
+  # Lets load generator run for a few seconds
+  sleep 5
+
+  # Ends initial load generation
+  if [[ $(curl http://$loadgen_ip/stop -o /dev/null -w "%{http_code}") -ne 200 ]]; then
+    log "Failed to stop initial load generation"
+  fi
```
Daniel-Sanche, line 306: Are you sure we want to shut down the load generation? Maybe we should keep it running so users continuously see things come in

terraform/install.sh
```diff
@@ -283,16 +284,26 @@ loadGen() {
   gcloud container clusters get-credentials loadgenerator --zone "$LOADGEN_ZONE"
   kubectx loadgenerator=.
 
-  # find the IP of the load generator web interface
+  # Kicks of initial load generation via a POST request to Locust web interface
```
simonz130, line 287: nit: Kicks off

simonz130 (approved):

terraform/install.sh
nicoleczhu: I'm a fan of this suggestion. This actually solves the Locust bug issue.

terraform/install.sh
nicoleczhu: It is indempotent

terraform/install.sh
nicoleczhu: thanks!

terraform/install.sh
nicoleczhu: removed. thanks!

terraform/install.sh
nicoleczhu: Hm another Locust bug. If we start the swarm with a URL post request, users don't get the `STOP` button in the UI. 😬😬😬

The fundamental issue here is the Locust page is all jquery. I'll reduce the swarm size and increase the swarm run time by a little bit.

nicoleczhu:
Ok this is mergeable

PR #612, opened 2020-12-17
update cartservice

mathieu-benoit:
- Size of the image: `200MB` --> `50.3MB`
- Update `Grpc.*` nuget packages to `2.34.0`
- Simplify the `Program.Start` mechanism
- [Single-File Executables in .NET](https://levelup.gitconnected.com/single-file-executables-in-net-core-3-1-and-the-quest-for-a-sub-50mb-docker-container-f44cb1274121)
- [App Trimming in .NET 5](https://devblogs.microsoft.com/dotnet/app-trimming-in-net-5/)
- Misc cleanups


src/cartservice/.dockerignore
```diff
@@ -0,0 +1,8 @@
+**/*.sh
+**/*.bat
```
simonz130, line 2: why these files are ignored?

src/cartservice/.dockerignore
```diff
@@ -0,0 +1,8 @@
...
+tests/
+**/bin/
+**/obj/
+**/out/
+Dockerfile*
+docker-compose*
```
simonz130, line 8: similarly, here?

src/cartservice/Program.cs
```diff
@@ -12,116 +12,15 @@
...
-using System.Runtime.CompilerServices;
+using cartservice;
 
-[assembly: InternalsVisibleTo("cartservice.tests")]
-namespace cartservice
-{
-    public class Program
-    {
-        const string REDIS_ADDRESS = "REDIS_ADDR";
+CreateHostBuilder(args).Build().Run();
```
simonz130, line 19: I'm not sure I understand how this even compiles - it looks like you are removing the namespace and class and this method is not contained anywhere. Is it a glitch in the comparison?

kubernetes-manifests/cartservice.yaml
```diff
@@ -38,8 +38,6 @@ spec:
         env:
         - name: REDIS_ADDR
           value: "redis-cart:6379"
-        - name: ASPNETCORE_URLS
-          value: "http://0.0.0.0:7070"
         resources:
```
simonz130, line 41: Did you test it deployed on kubernetes?

tests/cartservice/cartservice.tests.csproj
```diff
@@ -1,19 +1,19 @@
...
-    <PackageReference Include="Grpc.Net.Client" Version="2.32.0" />
-    <PackageReference Include="Grpc.Tools" Version="2.23.0" />
-    <PackageReference Include="Google.Protobuf" Version="3.12.3" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.0" />
-    <PackageReference Include="xunit" Version="2.3.1" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
-    <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" />
+    <PackageReference Include="Grpc.Net.Client" Version="2.34.0" />
+    <PackageReference Include="Grpc.Tools" Version="2.34.0" />
+    <PackageReference Include="Google.Protobuf" Version="3.14.0" />
```
simonz130, line 12: Did you test that tests for cart service pass with your code?

src/cartservice/docker-compose.yaml
```diff
@@ -23,6 +23,5 @@ services:
       - "7070:7070"
     environment:
       REDIS_ADDR: "redis:6379"
-      ASPNETCORE_URLS: "http://0.0.0.0:7070"
```
simonz130, line 26: why was this removed?

src/cartservice/Services/CartServiceController.cs
```diff
@@ -45,11 +45,6 @@ public async override Task<Empty> EmptyCart(EmptyCartRequest request, ServerCall
         {
             await _cartStore.EmptyCartAsync(request.UserId);
             return Empty;
-        } 
-        public async void Start()
-        {
-            Console.WriteLine("Starting cart service...");
-            await _cartStore.InitializeAsync();
```
simonz130, line 52: Where do you initialize cartStore?

simonz130 (changes requested):

src/cartservice/Program.cs
mathieu-benoit: Good question, that's one of the improvements with C# 9 described here: https://devblogs.microsoft.com/dotnet/welcome-to-c-9-0/#top-level-programs.

src/cartservice/Services/CartServiceController.cs
mathieu-benoit: This method was not used, the initialization is done in `Startup.cs`: https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/mathieu-benoit/update-cartservice/src/cartservice/Startup.cs#L52

src/cartservice/.dockerignore
mathieu-benoit: Those files should not go the container itself, so to avoid any `COPY` command in the `Dockerfile` bringing them, we could add their exclusion in the `.dockerignore` file as a best practice.

src/cartservice/.dockerignore
mathieu-benoit: Those type of files should not go the container itself (at least in the context of the `cartservice`), so to avoid any `COPY` command in the `Dockerfile` bringing them, we could add their exclusion in the `.dockerignore` file as a best practice.

src/cartservice/Dockerfile
```diff
@@ -12,34 +12,18 @@
...
+FROM mcr.microsoft.com/dotnet/sdk:5.0.101 as builder
 WORKDIR /app
-COPY *.csproj .
-
-# Workaround for https://github.com/grpc/grpc/issues/18428
-# Also see https://github.com/sgerrand/alpine-pkg-glibc
-RUN apk update && apk --no-cache add ca-certificates wget
-RUN wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub
-RUN wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.30-r0/glibc-2.30-r0.apk
-RUN apk add glibc-2.30-r0.apk
```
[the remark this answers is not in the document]
mathieu-benoit: I went through those links (workaround, etc.) mentioned, and now that we got the managed gRPC implementation with your PR https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/494, we shouldn't need this anymore (`Grpc.Tools` not used anymore for that part).
With that said, I don't see why on MacOS this shouldn't work, could you confirm that's needed for MacOS (I don't see any reference of this in the workaround link)? During the `docker build` or `docker run` part? I don't have access to a MacOS to test it unfortunately.
_Note: if you have an issue on MacOS, I think I know what could be the issue, but first let's checked the above points before going that path._

src/cartservice/docker-compose.yaml
mathieu-benoit: This environment variable is now part of the `Dockerfile`, so no need to add it here.

src/cartservice/.dockerignore
simonz130: Agree on the bat files for the container but not necessarily for sh files. Let's not ignore *.sh - in case we might want to wrap the microservice executable in a script

src/cartservice/Program.cs
simonz130: Let's revert to have a main method. .NET documentation/tutorials throughout the internet haven't picked up all the new features from the .NET team and many of the customers might find this confusing.

kubernetes-manifests/cartservice.yaml
mathieu-benoit: Yes I did.

simonz130: I prefer to set environment variables for kubernetes clusters in kubernetes manifest instead of docker file. If you then decide to use buildbacks for container building, you might introduce errors in configuration as a result of handling environment-related stuff in dockerfile instead of kubernetes yaml

mathieu-benoit: Having it in the `Dockerfile` allows to have centralized way to get thei `cartservice` app working on any circunstances (`Docker`, Docker-compose`, `Kubernetes`, etc.). It's not a Kubernetes thing, for example today this command `docker run cartservice` won't work unless `-e ASPNETCORE_URLS="http://0.0.0.0:7070"` is passed.

src/cartservice/.dockerignore
mathieu-benoit: Done.

src/cartservice/Program.cs
mathieu-benoit: Done.

kubernetes-manifests/cartservice.yaml
mathieu-benoit: Done.

src/cartservice/Dockerfile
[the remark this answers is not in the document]
mathieu-benoit: Quick update here, I'm able to reproduce the issue explained with the workaround (on Linux) if the `builder` image is `alpine` which is not the case with my update which doesn't have the issue and don't need that workaround: `mcr.microsoft.com/dotnet/sdk:5.0.101`.

tests/cartservice/cartservice.tests.csproj
mathieu-benoit: I didn't, but I'm on it and will report back once done. In the meantime, I have hard time to get the `tests/carservice/tests.yaml` working, is it working on `master`?

mathieu-benoit: I was able to successfully run the tests, but I needed to apply a fix https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/612/commits/7c8c7479e717e8953552a23fdbce95c7a31d1345.

src/cartservice/Dockerfile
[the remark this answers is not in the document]
simonz130: Awesome. Then sounds like we can remove it!

simonz130 (approved):

PR #622, opened 2021-01-11
fix: istio version

Daniel-Sanche:
The E2E-Release tests are failing, because for some reason it is using Istio 1.8.0 instead of the expected Istio 1.7.1.

This PR hardcodes the Istio version used to 1.7.1 instead of accepting an environment variable override to avoid this issue going forward


.github/workflows/e2e-release.yml
```diff
@@ -53,11 +53,13 @@ jobs:
         set -x
         docker pull ${{ steps.website_variables.outputs.cloudshell_image }}
         # run install script
+        # ISTIO_VCERSION is fix for v0.5.0. Can be removed after next release
```
simonz130, line 56: nit: typo - VERSION

simonz130 (approved):

PR #632, opened 2021-01-28
Add SRE recipe for rating service

minherz:
A new recipe 'rating_freshness' is added that should cover data freshness SLO of the rating service.
To support the recipe a new (3rd) SLO is define for the rating service. This is 1 day rollowing window SLO that defines a good time as a 1 minute interval during which there were at least 1 successful call to recollect API.
The SLO uses log-based metric which counts number of calls to recollect API.

### DIFF
--- a/sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
+++ b/sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
@@ -29,7 +29,7 @@ class RatingsFreshnessRecipe(Recipe):
     stops recollect API calls to rating service.
     """
 
-    name = "rating_freshness"
+    name = "recipe_"
 
     def get_name(self):
         return self.name
@@ -49,13 +49,16 @@ class RatingsFreshnessRecipe(Recipe):
             print("Failed: cannot find project id.")
             logging.error("Failed pausing scheduled job: no project id.")
             exit(1)
-        # hide error output to avoid additional hints :)
-        pause_command = "gcloud scheduler jobs pause ratingservice-recollect-job --project {pid} 2> /dev/null".format(
+        pause_command = "gcloud scheduler jobs pause ratingservice-recollect-job --project {pid}".format(
             pid=project_id
         )
-        Recipe._run_command(pause_command)
-        print("...done")
-        logging.info("Scheduled job of the rating service paused")
+        _, err_str = Recipe._run_command(pause_command)
+        if "ERROR:" in err_str:
+            print(err_str)
+            logging.error("Failed executing service breaking command:" + err_str)
+        else:
+            print("...done")
+            logging.info("Scheduled job of the rating service paused")
 
     def restore_service(self):
         """
@@ -69,13 +72,16 @@ class RatingsFreshnessRecipe(Recipe):
             print("Failed: cannot find project id.")
             logging.error("Failed pausing scheduled job: no project id.")
             exit(1)
-        # hide error output to avoid additional hints :)
-        resume_command = "gcloud scheduler jobs resume ratingservice-recollect-job --project {pid} 2> /dev/null".format(
+        resume_command = "gcloud scheduler jobs resume ratingservice-recollect-job --project {pid}".format(
             pid=project_id
         )
-        Recipe._run_command(resume_command)
-        print("...done")
-        logging.info("Scheduled job of the rating service resumed")
+        _, err_str = Recipe._run_command(resume_command)
+        if "ERROR:" in err_str:
+            print(err_str)
+            logging.error("Failed executing service restoring command:" + err_str)
+        else:
+            print("...done")
+            logging.info("Scheduled job of the rating service resumed")
 
     def hint(self):
         """
@@ -114,12 +120,14 @@ class RatingsFreshnessRecipe(Recipe):
         """
         prompt = "What is the cause of the break?"
         choices = [
-            "service does not run",
-            "ratings data inconsistent in the database",
-            "recollect requests fail or missing",
-            "new rating votes requests fail",
+            "rating service does not run",
+            "rating votes data is missing in the database",
+            "scheduler job that sends recollect request to rating service does not work",
+            "new ratings are calculated incorrectly",
         ]
-        answer = "recollect requests fail or missing"
+        answer = (
+            "scheduler job that sends recollect request to rating service does not work"
+        )
         Recipe._generate_multiple_choice(prompt, choices, answer)
 
     def verify(self):
--- a/sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
+++ b/sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
@@ -43,7 +43,7 @@ class RatingsFreshnessRecipe(Recipe):
         the rating service
         """
         logging.info("Pausing scheduled job of the rating service")
-        print("Breaking rating operations...")
+        print("Breaking service operations...")
         project_id = Recipe._get_project_id()
         if not project_id:
             print("Failed: cannot find project id.")
@@ -66,7 +66,7 @@ class RatingsFreshnessRecipe(Recipe):
         the rating service
         """
         logging.info("Resuming scheduled job of the rating service")
-        print("Restoring rating operations...")
+        print("Restoring broken operations...")
         project_id = Recipe._get_project_id()
         if not project_id:
             print("Failed: cannot find project id.")
--- a/sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
+++ b/sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
@@ -53,7 +53,7 @@ class RatingsFreshnessRecipe(Recipe):
             pid=project_id
         )
         _, err_str = Recipe._run_command(pause_command)
-        if "ERROR:" in err_str:
+        if "ERROR:" in str(err_str, "utf-8"):
             print(err_str)
             logging.error("Failed executing service breaking command:" + err_str)
         else:
@@ -76,7 +76,7 @@ class RatingsFreshnessRecipe(Recipe):
             pid=project_id
         )
         _, err_str = Recipe._run_command(resume_command)
-        if "ERROR:" in err_str:
+        if "ERROR:" in str(err_str, "utf-8"):
             print(err_str)
             logging.error("Failed executing service restoring command:" + err_str)
         else:
@@ -89,11 +89,15 @@ class RatingsFreshnessRecipe(Recipe):
         """
         project_id = Recipe._get_project_id()
         print(
-            "Try to change a rating for a shop's items one or more times.",
-            "Try to check SLO Monitoring and look for SLOs that consume error budget.",
-            "For SLO Monitoring open https://pantheon.corp.google.com/monitoring/services?project={pid}".format(
-                pid=project_id
-            ),
+            "\n".join(
+                (
+                    'Product ratings are managed by the "rating service", hosted on Google AppEngine.',
+                    "The service provides APIs that allow other services to get and update products' ratings.",
+                    "The rating data is kept up-to-date by periodically calling an API endpoint that collects",
+                    "all recently sent new rating scores for each product and calculates the new rating",
+                    "based on the old value and the new scores. Try to check if the rating service operates normally.",
+                )
+            )
         )
 
     def verify_broken_service(self):
--- a/sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
+++ b/sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
@@ -1,0 +1,134 @@
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# -*- coding: utf-8 -*-
+"""
+This module contains the implementation of rating_freshness recipe
+"""
+
+import logging
+import subprocess
+from recipe import Recipe
+
+
+class RatingsFreshnessRecipe(Recipe):
+    """
+    This class implements 'rating_freshness' which purposefully
+    stops recollect API calls to rating service.
+    """
+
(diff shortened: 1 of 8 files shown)


sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
```diff
@@ -0,0 +1,134 @@
...
+from recipe import Recipe
+
+
+class RatingsFreshnessRecipe(Recipe):
+    """
+    This class implements 'rating_freshness' which purposefully
+    stops recollect API calls to rating service.
+    """
+
+    name = "rating_freshness"
```
Daniel-Sanche, line 32: the other active recipe is called `recipe0`. I think that naming scheme was chosen to not spoil what the cause of the error was?

I'm open to changing the naming scheme, but we should keep it consistent

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
```diff
@@ -0,0 +1,134 @@
...
+    def hint(self):
+        """
+        Provides a hint about the root cause of the issue
+        """
+        project_id = Recipe._get_project_id()
+        print(
+            "Try to change a rating for a shop's items one or more times.",
+            "Try to check SLO Monitoring and look for SLOs that consume error budget.",
+            "For SLO Monitoring open https://pantheon.corp.google.com/monitoring/services?project={pid}".format(
+                pid=project_id
```
Daniel-Sanche, line 89: I saw the rating field disappear after submitting a rating. Is that what was expected?

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def hint(self):
+        """
+        Provides a hint about the root cause of the issue
+        """
+        project_id = Recipe._get_project_id()
+        print(
+            "Try to change a rating for a shop's items one or more times.",
+            "Try to check SLO Monitoring and look for SLOs that consume error budget.",
+            "For SLO Monitoring open https://pantheon.corp.google.com/monitoring/services?project={pid}".format(
```
Daniel-Sanche, line 88: When I look at the link, I see SLOs for the frontend, recommendationservice, and currencyservice. Is that expected or did something break?

Daniel-Sanche (changes requested):

Daniel-Sanche:
One concern I have is that this recipe may be too abstract. Will users be able to find and diagnose the issue without a full understanding of the architecture?

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
```diff
@@ -0,0 +1,134 @@
...
+        the rating service
+        """
+        logging.info("Resuming scheduled job of the rating service")
+        print("Restoring rating operations...")
+        project_id = Recipe._get_project_id()
+        if not project_id:
+            print("Failed: cannot find project id.")
+            logging.error("Failed pausing scheduled job: no project id.")
+            exit(1)
+        # hide error output to avoid additional hints :)
```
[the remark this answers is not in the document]
minherz: The comment explains the reason for `2> /dev/null` redirection in the command. Both calls to `gcloud scheduler jobs` print output to stderr. The `--quiet` flag does not hide it. The output looks like "Job is paused" and can be used as a simple path for finding a fix.

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
```diff
@@ -0,0 +1,134 @@
...
+        the breakage and prompts the user for an answer
+        """
+        prompt = "What is the cause of the break?"
+        choices = [
+            "service does not run",
+            "ratings data inconsistent in the database",
+            "recollect requests fail or missing",
+            "new rating votes requests fail",
+        ]
+        answer = "recollect requests fail or missing"
```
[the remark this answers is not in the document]
minherz: correct. i will change the wording :+1:

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
```diff
@@ -0,0 +1,134 @@
...
+            print("Failed: cannot find project id.")
+            logging.error("Failed pausing scheduled job: no project id.")
+            exit(1)
+        # hide error output to avoid additional hints :)
+        pause_command = "gcloud scheduler jobs pause ratingservice-recollect-job --project {pid} 2> /dev/null".format(
+            pid=project_id
+        )
+        Recipe._run_command(pause_command)
+        print("...done")
+        logging.info("Scheduled job of the rating service paused")
```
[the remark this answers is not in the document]
minherz: :confused: it worked when i was testing it. let me re-check.

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
minherz: yes, it is a designed behavior. it is intended as a simulation of the one time rating submission.
BTW, it is not part of this change :slightly_smiling_face:

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
minherz: I expect that a first line will direct the reader to look into relevant SLOs. Would you like to propose a different wording for the hint?

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
minherz: it is not easy to find a balance about the abstraction level. you [said it your self](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/632#issuecomment-770114127). i am OK to abstract the name of the recipe. @simonz130 I forgot to ask do you have some guidelines about the recipe design or wording?

minherz: if we want to abstract the recipes and avoid repetition, the naming should be randomizes. looking on the provided options: define a custom name and ability to deactivate specific recipe, it looks like the intention is to allow to a user some way to know what to expect. at the end, anyone can have a look into the code and learn what was broken.
would it be ok to rename it to "recipe_data_freshness"?

Daniel-Sanche: The original intention was to use numbers and let users find the cause themselves using the monitoring dashbaords, but we may want to rethink that at some point. My vote for this PR would be to keep at as `recipe1` to match the existing name format, and then we can re-visit the naming choice later as we roll out more recipes.

(Simon is officially on leave now, so he may want to weigh in next month)

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
Daniel-Sanche: Ok, I just wanted to make sure I was seeing what was expected!

Can you give a quick walkthrough of the thought process you think users will have to debug the issue? I want to make sure there's a clear metric or something they could notice to track down the problem. If the only clue is multiple failing SLOs, it can be hard to find the root cause

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
minherz: sure. since "recipe1" is a name of the [encoding_recipe](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/master/sre-recipes/recipes/encoding_recipe/encoding_recipe.py#L30) i will set the name to be "recipe2".

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
minherz: A starting point can be anything and depends on the background of the investigator, the initial info and setup, etc.
After reading the hint, I expect that a user will try to change rating for one or more products few times and see that it does not refresh neither immediately nor after some time.
After that, a user will go to look which services have problems with SLO or directly go to check error budget graph for rating service SLOs. There, i expect the user to see that error budget for data freshness is in decrease.
There are few possible paths from here but all are based on looking into architecture diagram and understanding what components of the rating service are there. The fact that availability SLO isn't hurt should hint that the problem is not with the service itself and to direct the user to check DB and Scheduler.

Daniel-Sanche: > After reading the hint, I expect that a user will try to change rating for one or more products few times and see that it does not refresh neither immediately nor after some time.

By refresh, you mean update the score, right? Personally, I'm not sure if this is clear enough. As a user, I'd assume an individual rating just didn't affect the average much

@nicoleczhu do you have thoughts on this?

minherz: One might not. Also there are few assumptions that, for my opinion, are reasonable to make:
- the long living sandbox environments are rare
- due to configuration user sees changes to the product rating within a minute since last change due to a small initial count of votes (=1) and recurrent call to recollect API (1min)
These assumptions seem reasonable to me. With these assumptions the updating of the product rating even once will have visible result. Also if a rating is 4 and a user selects to update it to 5 then the results will be less visible. But again, I am building on a common sense of the user who knows that the rating is calculating using average (or weighted average) and the new rating selection will be made to change the product's rating faster. At the end, this is a sandbox environment intended for users with a certain background.

nicoleczhu: UX comments while looking at this recipe with new eyes: 

1. Remove the first hint as it detracts from the ultimate goal of learning about SLOs.
2. We should roughly explain the components of ratings service, especially what recollect API does. We shouldn't require users to study app architecture. We're showcasing that you can get "observability" in a single/accessible view, so we shouldn't make users dig into code/READMEs. 
3. Given that we made issue identification easer for the user, your quiz questions could actually be a bit more advanced. i.e. asking users to think through changing SLO parameters, more critical thinking type questions that SREs encounter.

minherz: @nicoleczhu yes, the goal of the sandbox is to help a user to learn about SLOs and Cloud Ops services. Unfortunately, the existing UI does not support production grade SLOs and using simple SLOs does not teach too much about available options. I would argue that
# User has to use `gcloud` and not UI to do certain things with SLOs
# SLO does not provide "observability" about application problems. By default they raise alert about negative user experience

Regarding your suggestion in 1, if we were able to run the recipes in a virtual environment where everything is working except the one thing that we break, it make sense to take user directly to SLOs. Unfortunately, it is possible, especially if the load generator is working, there will be multiple issues with SLOs of different services. Without the first sentence, user my do a work unrelated to the recipe.

Can you please elaborate (2) and (3)? Where do you propose to put the explanation you mentioned? I appreciate if you can provide tangible example for more advanced quiz questions? I have them copy/pasted from two other recipes.

Daniel-Sanche: Maybe something like this?

```
Product ratings are managed by the "rating service", hosted on Google AppEngine.
The service provides APIs that allow other services to get and update products' ratings.
The rating data is kept up-to-date by periodically calling an API endpoint that collects
all recently sent new rating scores for each product and calculates the new rating
based on the old value and the new scores. Try to check if the rating service operates normally.
```

I still have concerns the recipe may not be discoverable enough, but we had the same thoughts about some of the existing recipes too. I think after updating the hint, we can merge this as-is and review all the recipes later (before we start to publicize the recipes feature)

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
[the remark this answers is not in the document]
Daniel-Sanche: Actually I'm seeing another issue:

```
Traceback (most recent call last):
  File "/home/sanche/cloudshell_open/cloud-ops-sandbox/sre-recipes/sandboxctl", line 172, in <module>
    cli()
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/sanche/cloudshell_open/cloud-ops-sandbox/sre-recipes/sandboxctl", line 98, in sre_recipes
    recipe.break_service()
  File "/home/sanche/cloudshell_open/cloud-ops-sandbox/sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py", line 56, in break_service
    if "ERROR:" in err_str:
TypeError: a bytes-like object is required, not 'str'
```

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
```diff
@@ -0,0 +1,146 @@
...
+    def is_active(self):
+        return True
+
+    def break_service(self):
+        """
+        Pause Cloud scheduler job to stop calls to recollect API of
+        the rating service
+        """
+        logging.info("Pausing scheduled job of the rating service")
+        print("Breaking rating operations...")
```
Daniel-Sanche, line 46: Do you want to tell them which service is being broken? I thought that was part of what they were fining out

Maybe we should copy `recipe0` and say "Deploying broken service" instead

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
```diff
@@ -0,0 +1,146 @@
...
+            print("...done")
+            logging.info("Scheduled job of the rating service paused")
+
+    def restore_service(self):
+        """
+        Resume Cloud scheduler job to restore calls to recollect API of
+        the rating service
+        """
+        logging.info("Resuming scheduled job of the rating service")
+        print("Restoring rating operations...")
```
Daniel-Sanche, line 69: same comment here

Daniel-Sanche (approved):

sre-recipes/recipes/ratings_freshness_recipe/ratings_freshness_recipe.py
minherz: yes, i will change it to a neutral message

Issue #648, opened 2021-03-23
Add sandboxctl describe


Ayelet41:
get all the URLs for the current sandbox cluster: workspaces URL, hipstershop URL (GKE FE), load generation URL, etc. 
sandboxctl describe should output the same URLs that are presented: after you create a new Sandbox, you'll see a bunch of URLs that our script outputs.
Note:  it's not as simple as storing these at the end of script to a file and reading from that file in "describe". 

Also we want to support users switching context between different sandboxes, like kubectl switch context.

PR #649, opened 2021-03-23
Add sandboxctl describe #648

Ayelet41:
Added describe and supported functions + separated  get project id for reuse

### DIFF
--- a/sre-recipes/recipe.py
+++ b/sre-recipes/recipe.py
@@ -75,6 +75,16 @@ class Recipe(abc.ABC):
         if not project_id:
             logging.error("Could not retrieve project id: " + error)
         return project_id
+    
+    @staticmethod
+    def _get_external_ip():
+        Recipe._auth_cluster()
+        external_ip_command = "kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'"
+        ip, error = Recipe._run_command(external_ip_command)
+        ip = ip.decode("utf-8").replace("'", '')
+        if not ip:
+            logging.error('No external IP found:' + error)
+        return ip
 
     @staticmethod
     def _auth_cluster(cluster="APP"):
--- a/sre-recipes/recipe.py
+++ b/sre-recipes/recipe.py
@@ -81,10 +81,9 @@ class Recipe(abc.ABC):
         Recipe._auth_cluster()
         external_ip_command = "kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'"
         ip, error = Recipe._run_command(external_ip_command)
-        ip = ip.decode("utf-8").replace("'", '')
         if not ip:
             logging.error('No external IP found:' + error)
-        return ip
+        return ip.decode("utf-8").replace("'", '')
 
     @staticmethod
     def _auth_cluster(cluster="APP"):
--- a/sre-recipes/sandboxctl
+++ b/sre-recipes/sandboxctl
@@ -81,6 +81,21 @@ def get_recipes():
     recipes = get_valid_recipes(attributes)
     return recipes
 
+def get_loadgen_ip():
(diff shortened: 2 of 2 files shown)


sre-recipes/recipe.py
```diff
@@ -75,6 +75,16 @@ def _get_project_id():
         if not project_id:
             logging.error("Could not retrieve project id: " + error)
         return project_id
+    
+    @staticmethod
+    def _get_external_ip():
+        Recipe._auth_cluster()
```
simonz130, line 81: I see it's a draft so not reviewing this yet but as soon as it's ready for review - remove the draft label from your PR and let us know on team chat.

sre-recipes/sandboxctl
```diff
@@ -161,6 +172,23 @@ def create(project, *args, **kwargs):
...
+def describe():
+    """Show information about an existing sandbox"""
+    gcp_path = "https://console.cloud.google.com"
+    loadgen_ip = get_loadgen_ip()
+    external_ip = get_external_ip()
+    project_id = get_project_id()
+    gcp_kubernetes_path = gcp_path + '/kubernetes/workload?project=' + project_id
+    gcp_monitoring_path = gcp_path + '/monitoring?project=' + project_id
+    print("""
+    Cloud Operations Sandbox info for project {}:
```
Daniel-Sanche, line 185: nit: it would be best if the order of these fields matched the [initial success message](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/a46746143530cdad0c371931ccb88e953a0afa40/terraform/install.sh#L328) (I wonder if we should even change that block to just call `sandboxctl describe`?)

Daniel-Sanche:
One testing tip I didn't mention when we met: you can copy/paste the open in cloud shell link from https://cloud-ops-sandbox.dev and replace the "cloudshell_git_branch" argument to directly test your branch:

https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/GoogleCloudPlatform/cloud-ops-sandbox.git&cloudshell_git_branch=b-648&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image/uncertified:v0.5.0

I tested this with your branch, and it looks good to me

Daniel-Sanche:
I left a few small comments you may want to consider, but overall looks good

sre-recipes/recipe.py
```diff
@@ -75,6 +75,16 @@ def _get_project_id():
...
+        external_ip_command = "kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}'"
+        ip, error = Recipe._run_command(external_ip_command)
+        ip = ip.decode("utf-8").replace("'", '')
+        if not ip:
```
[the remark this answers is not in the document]
Ayelet41: Yes, I haven't change the code just "moved it":
But, what about something like that?
 if ip:
            ip = ip.decode("utf-8").replace("'", '')
        else:

sre-recipes/sandboxctl
Ayelet41: Moved project_id, but not sure what you meant, can you clarify the brackets?

Daniel-Sanche: To match the [initial success message](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/a46746143530cdad0c371931ccb88e953a0afa40/terraform/install.sh#L328), it should look like this:

```
Cloud Operations Sandbox info for project {}:
    - Google Cloud Console KBE Dashboard: {}
    - Google Cloud Console Monitoring Workspace {}
    - Load generator web interface: {}
    - Hipstershop web app address: {}""".format(project_id, loadgen_ip, external_ip, gcp_kubernetes_path, gcp_monitoring_path)
```

In the brackets, I was suggesting maybe we should call this function to print the success message instead of having similar code in two places. But probably not worth changing at this point

Daniel-Sanche (approved):

PR #703, opened 2021-05-06
feat: publicworks entrypoints

simonz130:


daniel-sanche (changes requested):
Currently blocking merging until we implement new git system

daniel-sanche (commented):
@simonz130 is this still relevant?

PR #719, opened 2021-05-15
Refactor LoadGen to use Library Mode, and add SRE Recipe LoadGen API placeholder

Gan-Tu:
In this PR, I

- added code to run LoadGen as a flask app, using Locust library mode code, listening on port 8089
- added a second Locust server (aka "environment") that will be started at the same time to serve as a decoupled, SRE Recipe API endpoint listening on port 8090
- added basic integration test for the SRE Recipe API endpoint in the e2e tests

I kept it so that functionally, the user facing web UI still supports the same set of existing tasks and behaviors, while the additional API endpoint will support a simple `/api/ping` endpoint as placeholder. 

I updated the Kubernetes and Docker file to expose the additional API port (8090) as port 81 (an arbitrary choice by me) on the main locust IP.

I will in future PRs complete the implementation of the SRE Recipe API endpoint. 

Thanks

---

Test it by deploying this change in GCP and try to access the `http://<locust ip>:81/api/ping`.

[Open in Cloud Shell](https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/GoogleCloudPlatform/cloud-ops-sandbox.git&cloudshell_git_branch=gan-loadgen&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image/uncertified:v0.6.0&cloudshell_tutorial=docs/tutorial.md)


src/loadgenerator/Dockerfile
```diff
@@ -24,10 +24,11 @@ RUN pip install -r requirements.txt
 
 # Copy tasks and runner script
 COPY run.sh .
-COPY locust-tasks ./locust-tasks
+COPY *.py ./
+COPY locust_tasks ./locust_tasks
 
 # Expose the required Locust ports
-EXPOSE 5557 5558 8089
+EXPOSE 5557 5558 8089 8090
```
simonz130, line 31: curious, what is this port used for (8090)? Perhaps, worth adding a comment.

src/loadgenerator/logger.py
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+
+import logging
+import sys
+from pythonjsonlogger import jsonlogger
+
+# TODO(yoshifumi) this class is duplicated since other Python services are
+# not sharing the modules for logging.
+class CustomJsonFormatter(jsonlogger.JsonFormatter):
+  def add_fields(self, log_record, record, message_dict):
```
simonz130, line 24: @daniel-sanche can we use python-logging library for structured logging into stdout instead of having this json implementation?

loadgenerator-manifests/loadgenerator.yaml
```diff
@@ -60,7 +64,11 @@ spec:
       targetPort: loc-master-web
       protocol: TCP
       name: loc-master-web
+    - port: 81
```
daniel-sanche, line 67: Are we expecting users to find and use this UI? If so, we might want to expose it in a different way. (Maybe[ fan out ingress?](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#multiple_backend_services))

loadgenerator-manifests/loadgenerator.yaml
```diff
@@ -31,13 +31,14 @@ spec:
     spec:
       containers:
         - name: locust-main
-          image: gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator/gke:latest
+          # TODO(tugan): update this to official stack-driver container image, when updated
+          image: tugan0329/cloud-ops-sandbox-loadgen:latest
```
daniel-sanche, line 35: When are you planning on updating this? Before merging?

src/loadgenerator/logger.py
daniel-sanche: We could, although it wouldn't capture the timestamp in the exact same way.

What's the purpose of the logging here? Are users expected to see them in CLoud Logging, or is this just for internal use? Why does it need to be JSON formatted?

Gan-Tu: > What's the purpose of the logging here? Are users expected to see them in CLoud Logging, or is this just for internal use? Why does it need to be JSON formatted?

I copied this file exactly as it is from our other Python-based services (e.g. email, recommendation). I thought this is what is preferred in this project in terms of logging in python. I don't have a preference in terms of logging and whether they should be visible, but I think it is ok to be visible in Cloud logging. However, I will defer to you all to decide what loggings we want.

src/loadgenerator/Dockerfile
Gan-Tu: Sounds good. I added more documentation for the original ports exposed here as well

loadgenerator-manifests/loadgenerator.yaml
Gan-Tu: Users are not supposed to find this UI. It is only intended for access/use by SRE Recipe in terms of sending requests for generating specific traffic loads.

loadgenerator-manifests/loadgenerator.yaml
Gan-Tu: I don't have access to create a new image there. 

After this PR is approved, either someone can help me create one and I update it before merging, or we can have a different PR afterwards.

src/loadgenerator/logger.py
daniel-sanche: Ok, I think we should leave it as-is for this PR then. We could maybe open an issue to re-assess the logging method later, but I'm not sure I know the full scope

loadgenerator-manifests/loadgenerator.yaml
daniel-sanche: I think you can leave it as `gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator/gke:latest`, and the image will be [updated automatically when you push to master](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/e1c33e6f6d7b8d43499942da52c762f94a3b20e7/.github/workflows/push-master.yml#L39), right?

daniel-sanche (approved):
LGTM, but you should revert the loadgenerator container path (and let the tests pass) before merging

loadgenerator-manifests/loadgenerator.yaml
Gan-Tu: sounds good!

Gan-Tu: @daniel-sanche 
If I change it to gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator/gke:latest, the E2E test for CI will run using the old latest image (because this PR hasn't been merged to master), but with the updated E2E test for the new API endpoint, thus fail. Should we send a follow up one line PR instead to change the path, once the new image is auto updated

daniel-sanche: The intention of the CI is that it should be automatically building and testing the image based on what's in the repo HEAD, not the latest published image. (see [here](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/e1c33e6f6d7b8d43499942da52c762f94a3b20e7/.github/workflows/e2e-latest.yml#L54) where it builds a new tag, and [here](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/e1c33e6f6d7b8d43499942da52c762f94a3b20e7/.github/workflows/e2e-latest.yml#L63) where it re-writes the manifests to use it)

Although looking at the code, it's [updating `./kubernetes-manifests`](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/e1c33e6f6d7b8d43499942da52c762f94a3b20e7/.github/workflows/e2e-latest.yml#L63), but may be missing `loadgenerator-manifests`. You may have to update that part of the CI script if it's failing on your changes. But I don't think we should merge code to master pointing at a different container path

PR #724, opened 2021-05-20
Update gRPC and protobuf in Go services

ymotongpoo:
* Update proto file in the root (`health.proto` and `demo.proto`) to meet latest protobuf grammer
* Update gPRC and protobuf tool version in Go services (`shipping`, `productcatalog` and `checkout`)
* Add `tool.go` to manage the versions of gRPC and protobuf plugin for `protoc`
* Update `genproto.sh` to use proper option for `protoc`


ymotongpoo:
This is related to #548 and #549, which is auto upgrade by renovatebot to Java services.

pb/demo.proto
```diff
@@ -16,6 +16,8 @@ syntax = "proto3";
 
 package hipstershop;
 
+option go_package = "./;hipstershop";
```
nicoleczhu, line 19: Is the semicolon intentional?

src/shippingservice/Dockerfile
```diff
@@ -18,14 +18,16 @@ ENV PROJECT github.com/GoogleCloudPlatform/cloud-ops-sandbox/src/shippingservice
 WORKDIR /go/src/$PROJECT
 
 # restore dependencies
-COPY go.mod go.sum *.go ./
+COPY go.mod go.sum main.go quote.go tracker.go ./
 COPY genproto ./genproto/
 RUN go mod vendor -v
+
+COPY . .
```
nicoleczhu, line 25: Does this make lines 21, 22 redundant? Since we end up copying everything anyway. 

Or we can just be explicit about what additional vendor/files we're copying over.

nicoleczhu (approved):

src/shippingservice/Dockerfile
ymotongpoo: ohhh, it was left after some try and error process. I removed the `COPY . .` line.

pb/demo.proto
ymotongpoo: the meaning of semicolon is described [here](https://developers.google.com/protocol-buffers/docs/reference/go-generated#package). Though I realize it is discouraged to use now, I keep it for now.

PR #730, opened 2021-05-27
Remove msg about creating Workspace in Cloud Operations   

Ayelet41:
There is no need anymore for setupOperationsWorkspace step as there is a default workspace now in SD.
Verified against master in current version, release is now GA so should support customers as well.


daniel-sanche (approved):
Just tested this and it works! Finally

PR #731, opened 2021-06-01
chore: update auto-label.yaml

tmatsuo:
I have a plan to implement the config schema check upon config file retrieval in auto-label bot.
I want to remove the wrong field (it's not in the schema definition) before deploying the change.


tmatsuo:
@nicoleczhu @tbpg FYI

nicoleczhu (approved):

PR #732, opened 2021-06-04
Implement LoadGen SRE Recipe API

Gan-Tu:
In this PR, I

1. added three additional handlers to the SRE Recipe API endpoint for Load Generator
2. added a few end to end tests
3. added initial Locust task (`BasicPurchasingUser`) trigger-able via SRE Recipes

Endpoints added are:

- POST `/api/spawn/<user_identifier>`: takes required form parameters  `user_count` and `spawn_rate` for generating the load tasks associated with `user_identifier` at `spawn_rate` users per second, until we achieve a total of `user_count` users. An optional form parameter `stop_after` can be set to automatically stop all load generating users after a certain time has passed.
- GET `/api/user_count`: return the current number of users (which is also available via `/stats/requests` response)
- POST`/api/stop`: stop all load generating users in SRE Recipe LoadGen backend

---
Demo (with newly built loadgen image for this PR)

[Open in Cloud Shell](https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/GoogleCloudPlatform/cloud-ops-sandbox.git&cloudshell_git_branch=gan-loadgen-demo&shellonly=true&cloudshell_image=gcr.io/stackdriver-sandbox-230822/cloudshell-image/uncertified:v0.6.0&cloudshell_tutorial=docs/tutorial.md)

Then, try:

```
curl -X POST "http://<loadgen IP>:81/api/spawn/BasicPurchasingUser" -F "user_count=10" -F "spawn_rate=2" -F "stop_after=20"

curl "http://<loadgen IP>:81/api/user_count"

curl -X POST "http://<loadgen IP>:81/api/stop"
```


src/loadgenerator/locust_tasks/sre_recipe_load_tasks.py
```diff
@@ -0,0 +1,69 @@
...
+PRODUCTS = [
+    '0PUK6V6EV0',
+    '1YMWWN1N4O',
+    '2ZYFJ3GM2N',
+    '66VCHSJNUP',
+    '6E92ZMYYFZ',
+    '9SIQT8TOJO',
+    'L9ECAV7KIM',
+    'LS4PSXUNUM',
+    'OLJCESPC7Z',
```
simonz130, line 30: 2 things:

1) maybe we can centralize the list of products in one place, accessible both by the sample app microservices as well as loadgen.

2) wherever this list is created, we should add inline comment saying which product of hipstershop it represents so it's easier to track the code with context.

src/loadgenerator/sre_recipe_utils.py
```diff
@@ -17,15 +17,99 @@
...
+            if user_count is None:
+                return 400, {"err": f"Must specify a valid, non-empty, integer value for query parameter 'user_count': {user_count}"}
+            elif spawn_rate is None:
+                return 400, {"err": f"Must specify a valid, non-empty, integer value for query parameter 'spawn_rate': {spawn_rate}"}
+            elif user_count <= 0:
+                return 400, {"err": f"Query parameter 'user_count' must be positive: {user_count}"}
+            elif spawn_rate <= 0:
+                return 400, {"err": f"Query parameter 'spawn_rate' must be positive: {spawn_rate}"}
+            elif user_class is None:
+                return 400, {"err": f"Cannot find SRE Recipe Load for: {user_identifier}"}
```
daniel-sanche, line 87: would this be 404?

src/loadgenerator/sre_recipe_utils.py
```diff
@@ -17,15 +17,99 @@
...
+
+            # We currently only support running one SRE Recipe load each time
+            # for implementation simplicity.
+            env.runner.quit()  # stop existing load generating users, if any
+            env.user_classes = [user_class]  # replace with the new users
+
+            # wait a short while for all existing users to stop, then
+            # start generating new load with the new user types
+            gevent.spawn_later(WAIT_SECONDS_BEFORE_SPAWN,
+                               lambda: env.runner.start(user_count, spawn_rate))
```
daniel-sanche, line 102: Is there a way to watch for existing users to stop, rather than hard-coding a time?

src/loadgenerator/sre_recipe_utils.py
```diff
@@ -17,15 +17,99 @@
...
+
+            if stop_after:
+                gevent.spawn_later(WAIT_SECONDS_BEFORE_SPAWN + stop_after,
+                                   lambda: env.runner.quit())
+
+            return 200, {"msg": f"Started spawning {user_count} users at {spawn_rate} users/second"}
```
daniel-sanche, line 108: nit: This message may be a bit misleading since spawning doesn't start immediately. Maybe "spawn request received: {user_count} users at {spawn_rate} users/second"?

daniel-sanche (approved):
LGTM, just added a few comments

tests/provisioning/loadgen_test.py
```diff
@@ -111,13 +111,85 @@ def testStartSwarm(self, pattern):
...
+            stats = json.loads(resp.text)
+            if stats["user_count"] == 10:
+                all_users_spawned = True
+            if stats["total_rps"] > 0:
+                has_rps = True
+            tries += 1
+        self.assertTrue(all_users_spawned and has_rps)
+
+        # wait 20 more seconds and check if users are being stopped
+        # This give us a plenty of buffer for auto stop conditions
```
simonz130, line 185: Is it possible to get a callback on stopping? Waiting interval and testing might lead to instable tests

simonz130 (commented):
Great job @Gan-Tu! Thank you a lot!

One general comment - we should add a readme file that explains how to use the new http endpoints you added. It's explained in code but it should be explained in a readme, perhaps, with an image that illustrates the architecture.

src/loadgenerator/locust_tasks/sre_recipe_load_tasks.py
Gan-Tu: 1. ACKed. It seems that doing that will involve more changes affecting multiple directories to ensure the data files are correctly packaged across all the builds and images, and resolve any cross folder python import issues. I will leave it out for this PR. 
2. Done.

src/loadgenerator/locust_tasks/sre_recipe_load_tasks.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # check out
+        self.client.post("/cart/checkout", {
+            'email': 'someone@example.com',
+            'street_address': '1600 Amphitheatre Parkway',
+            'zip_code': '94043',
+            'city': 'Mountain View',
+            'state': 'CA',
+            'country': 'United States',
+            'credit_card_number': '4432-8015-6152-0454',
```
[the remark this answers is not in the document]
Gan-Tu: No significance. I just copied it from the example credit card number in the micro service. Changed it to 1111-1111-1111-1111

src/loadgenerator/sre_recipe_utils.py
```diff
@@ -17,15 +17,99 @@
...
+                return 400, {"err": f"Query parameter 'user_count' must be positive: {user_count}"}
+            elif spawn_rate <= 0:
+                return 400, {"err": f"Query parameter 'spawn_rate' must be positive: {spawn_rate}"}
+            elif user_class is None:
+                return 400, {"err": f"Cannot find SRE Recipe Load for: {user_identifier}"}
+
+            # Optional Query Parameters
+            stop_after = request.form.get("stop_after", default=None, type=int)
+            if stop_after is not None and stop_after <= 0:
+                return 400, {"err": f"Query parameter 'stop_after' must be positive: {stop_after}"}
```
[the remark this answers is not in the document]
Gan-Tu: Ideally, we can throw ValueError for 400, and some other for NotFound, but readability wise for now, there's no immediate benefit using `return status_code, response` vs  `raise ValueError(....)`? I think using the return schematics is also common.

src/loadgenerator/sre_recipe_utils.py
Gan-Tu: Done. Good catch

src/loadgenerator/sre_recipe_utils.py
Gan-Tu: Unfortunately, no. There's event hooks when a single user is stopped, but making it work when all users are stopped prob need some state tracking, and can be more convoluted than needed. However, I acknowledge this is not ideal neither. We can also instead do a busy wait loop to fire this only when the number of users are zero. Thoughts?

@daniel-sanche

src/loadgenerator/sre_recipe_utils.py
```diff
@@ -17,15 +17,99 @@
...
-            resp = make_response(jsonify({"msg": "pong"}), 200)
-            resp.headers["Content-Type"] = 'application/json'
-            return resp
+            return 200, {"msg": "pong"}
+
+        @env.web_ui.app.route("/api/user_count")
+        @return_as_json_response
+        def user_count():
+            """Return the number of currently running users"""
+            return 200, {"user_count": env.runner.user_count}
```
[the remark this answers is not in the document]
Gan-Tu: Done

src/loadgenerator/sre_recipe_utils.py
```diff
@@ -17,15 +17,99 @@
...
+            Form Paramters:
+            - user_count: Required. The total number of users to spawn
+            - spawn_rate: Required. The spawn rate for the users.
+            - stop_after: Optional. If specified, run the load generation only 
+                          for the given number of seconds.
+            """
+            # Required Query Parameters
+            user_count = request.form.get("user_count", default=None, type=int)
+            spawn_rate = request.form.get("spawn_rate", default=None, type=int)
+            user_class = get_sre_recipe_user_class(user_identifier)
```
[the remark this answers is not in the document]
Gan-Tu: It returns None. Added comment

tests/provisioning/loadgen_test.py
```diff
@@ -111,13 +111,85 @@ def testStartSwarm(self, pattern):
...
+        if spawn_rate:
+            form_data["spawn_rate"] = spawn_rate
+
+        r = requests.post(
+            f"{TestLoadGenerator.api_url}/api/spawn/BasicPurchasingUser", form_data)
+        self.assertFalse(r.ok)
+        resp = json.loads(r.text)
+        self.assertTrue(len(resp.get("err", "")) > 0)
+
+    def testApiSpawnUserClassExists(self):
```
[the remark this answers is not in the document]
Gan-Tu: Done. Also renamed other tests in the format of `ErrorOnXXX` and fixed documentation

Gan-Tu:
Addressed all comments.

src/loadgenerator/sre_recipe_utils.py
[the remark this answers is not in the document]
daniel-sanche: One argument for using exceptions is that it's easier to bubble up errors from helper functions, which could be useful if you want to share logic as the codebase grows. Or you may just find it more readable to pull the validation checks into their own function for readibility. See this [example from bank of anthos](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/d1848ad56e5579703bfd425e84808ba324109814/src/userservice/userservice.py#L135)

But if you don't expect the the function to get much more complex later, I think this pattern is fine too

src/loadgenerator/sre_recipe_utils.py
daniel-sanche: Personally I think the busy wait loop would be better if waiting is important, but I'm not familiar with how locust works. Do we have tests in place that would catch issues around this?

src/loadgenerator/sre_recipe_utils.py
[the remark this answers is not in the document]
Gan-Tu: Agree. Changed to exception pattern. Done.

src/loadgenerator/sre_recipe_utils.py
Gan-Tu: We do have tests on this part, but I agree with you, so changed to busy wait with a timeout.

Gan-Tu:
Addressed followup comments

PR #736, opened 2021-06-08
nit: Fix the target version to launch in getting started doc

ymotongpoo:
Update the target version in the launch URL for "Open in Cloud Shell" button.


simonz130 (approved):

simonz130:
@daniel-sanche - don't we auto-update this with the release script?

PR #743, opened 2021-06-09
Fix sandboxctl destroy timeout

Ayelet41:
Added run_command_interactive to solve sandboxctl destroy hang without affecting other call to run_command function.

### DIFF
--- a/sre-recipes/recipe.py
+++ b/sre-recipes/recipe.py
@@ -66,6 +66,16 @@ class Recipe(abc.ABC):
         )
         output, error = process.communicate()
         return output, error
+    
+    @staticmethod
+    def _run_command_interactive(command):
+        """Runs the given interactive command that waits for user input and returns any output and error"""
+        process = subprocess.Popen(
+            command.split(), shell=True, stderr=subprocess.PIPE
+        )
+        output, error = process.communicate()
+        return output, error
+
 
     @staticmethod
     def _get_project_id():
--- a/sre-recipes/recipe.py
+++ b/sre-recipes/recipe.py
@@ -70,11 +70,8 @@ class Recipe(abc.ABC):
     @staticmethod
     def _run_command_interactive(command):
         """Runs the given interactive command that waits for user input and returns any output and error"""
-        process = subprocess.Popen(
-            command.split(), shell=True, stderr=subprocess.PIPE
-        )
-        output, error = process.communicate()
-        return output, error
+        process = subprocess.run(command.split())
+        return 
 
 
     @staticmethod
--- a/sre-recipes/sandboxctl
+++ b/sre-recipes/sandboxctl
@@ -200,7 +200,7 @@ def destroy():
     """Delete an existing sandbox"""
     os.chdir(os.path.abspath(sys.path[0]))
     destroy_command = "../terraform/destroy.sh"
-    Recipe._run_command(destroy_command)
+    Recipe._run_command_interactive(destroy_command)
 
 
 if __name__ == "__main__":
--- a/terraform/destroy.sh
+++ b/terraform/destroy.sh
@@ -20,7 +20,7 @@ set -o errexit  # Exit on error
 #set -x
 
 log() { echo "$1" >&2;  }
-echo 'Start shutting down resources and deleting the project, this should take about 20 mins. Please do not close the session.'
+
 IFS=$'\n'
 acct=$(gcloud info --format="value(config.account)")
 
@@ -83,7 +83,7 @@ gcloud projects delete $PROJECT_ID
 
 
 # if user backed out or deletion failed, stop script
-found=$(gcloud projects list --filter="${PROJECT_ID}" --format="value(projectId)")
+found=$(gcloud projects describe ${PROJECT_ID} --flatten="lifecycleState" |grep "ACTIVE")
 if [[ -n "${found}" ]]; then
     log "project $PROJECT_ID not deleted"
     sendTelemetry $PROJECT_ID sandbox-not-destroyed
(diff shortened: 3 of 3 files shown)


sre-recipes/recipe.py
```diff
@@ -66,6 +66,16 @@ def _run_command(command):
         )
         output, error = process.communicate()
         return output, error
+    
+    @staticmethod
+    def _run_command_interactive(command):
+        """Runs the given interactive command that waits for user input and returns any output and error"""
+        process = subprocess.Popen(
+            command.split(), shell=True, stderr=subprocess.PIPE
```
nicoleczhu, line 74: maybe also add `stdout=subprocess.PIPE`

terraform/destroy.sh
```diff
@@ -20,7 +20,7 @@ set -o errexit  # Exit on error
 #set -x
 
 log() { echo "$1" >&2;  }
-
+echo 'Start shutting down resources and deleteing the project, this should take about 20 mins. Please do not close the session.'
```
nicoleczhu, line 23: `deleteing` -> `deleting`

sre-recipes/recipe.py
Ayelet41: In this case, that the user will be promoted to choose the project they want to delete.

sre-recipes/sandboxctl
```diff
@@ -200,7 +200,7 @@ def destroy():
     """Delete an existing sandbox"""
     os.chdir(os.path.abspath(sys.path[0]))
     destroy_command = "../terraform/destroy.sh"
-    Recipe._run_command(destroy_command)
+    Recipe._run_command_interactive(destroy_command)
```
simonz130, line 203: Could you please open a new issue for refactoring sandboxctl (unrelated to your changes).
It looks weird now that sandboxctl is hidden in sre-recipes folder. We originally meant that it will only be used for recipes but since then we extended it to support automated cluster creation, destroy and other things.

We need to move sandboxctl to its own folder under root of the repo. Maybe it would make sense to make it modular so that sre recipe commands will stay in sre-recipes folder while sandbox-wide commands will be under [root]/sandboxctl folder.

simonz130 (approved):

terraform/destroy.sh
```diff
@@ -20,7 +20,7 @@ set -o errexit  # Exit on error
 #set -x
 
 log() { echo "$1" >&2;  }
-
+echo 'Start shutting down resources and deleting the project, this should take about 20 mins. Please do not close the session.'
```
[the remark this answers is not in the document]
Ayelet41: The time yes is correct, I tested it a few dozen times and it took me between 10-22˜ minutes.

I can change to up to 20 minutes.
Regarding the text, yes the deletion start only after choosing the project, but if we are not promoting any message here from user percepite it feel like nothing happen. 
Do you have a suggestion for alternative text ? e.g.  
````
You will shortly be promoted to choose a project to delete and shutting down associated resources. This process can take up to 20 mins, please do not close the session in the meantime.
```

daniel-sanche (changes requested):
Feedback on the warning message

terraform/destroy.sh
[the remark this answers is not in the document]
daniel-sanche: Marking the project for deletion should be a simple API call, we shouldn't expect a 20 minute wait for that.

I think the problem here is that the project look-up query before the deletion is now stalling. We should look into how to speed up that part of the code and fix that first. Let me know if you want to take that on, otherwise I can take a look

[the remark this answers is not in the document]
daniel-sanche: Actually I did some testing, and it looks like the problem is still related to sandboxctl interactivity. It seems to load the list  after 1 minute with `terraform/destroy.sh`, but sandboxctl hangs when it gets to the `select` line

sre-recipes/sandboxctl
Ayelet41: Opened [issue/748](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/748)

terraform/destroy.sh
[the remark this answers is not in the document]
Ayelet41: I checked again, as you stated list takes about a minute and it's mainly do to the loop. I couldn't find a better way to find "owned projects". 
 `gcloud projects delete $PROJECT_ID` took about 12 minutes from the script , email was sent immediately and outside of script also got immediate response.  It seems its not due to the look-up before, but due to `gcloud projects list ` after project deletion.
 Instead if we use `gcloud projects describe cloud-ops-sandbox-3270241197 --flatten="lifecycleState"  |grep "ACTIVE"` it will speed things up (used grep as filter nor format dont seems to work well with describe, if you find notehr hack let me know)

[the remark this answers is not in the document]
Ayelet41: Please see the update. 
- I still recommend though to add the message before project list as it can take time depends on permissions and it can be confusing to the user.
- Also do we still to filter the old format `id:stackdriver-sandbox-* AND name='Stackdriver Sandbox Demo'`  ?

sre-recipes/recipe.py
Ayelet41: After further investigation it might be better to use `run` instead i.e. `process =subprocess.run(command.split())` thoughts?

sre-recipes/recipe.py
```diff
@@ -70,11 +70,8 @@ def _run_command(command):
     @staticmethod
     def _run_command_interactive(command):
         """Runs the given interactive command that waits for user input and returns any output and error"""
-        process = subprocess.Popen(
-            command.split(), shell=True, stderr=subprocess.PIPE
-        )
-        output, error = process.communicate()
-        return output, error
+        process = subprocess.run(command.split())
```
daniel-sanche, line 73: nit: It might be useful to return some data from the completed process. If not, then you should get rid of the unused process variable

daniel-sanche (approved):
Tested it and it's working for me now, thanks for putting this together!

PR #745, opened 2021-06-12
Support streaming outputs for sandboxctl destroy

Gan-Tu:
**Problem:** 

`sandboxctl destroy` can take a while to execute, and it uses the subprocess module, which doesn't stream the terraform destroy logs to users, so it looks as if the command is frozen.

**Solution:**

We add a flag/toggle for streaming stdout and stderr in the module that runs the command.

**Misc**:

We also fix some documentation issues, and simplify the subprocess run logic. Specifically, `subprocess.run` is preferred, and `subprocess.Popen` is the underlying module that it calls.

### DIFF
--- a/sre-recipes/README.md
+++ b/sre-recipes/README.md
@@ -7,23 +7,23 @@ SRE Recipes is a tool to help users familiarize themselves with finding the root
 To view which recipes exist, run the command below in this directory:
 
 ```
-$ ./sandboxctl --help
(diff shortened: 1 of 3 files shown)


Gan-Tu:
Closing this issue as it seems https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/743 is working on this, and there is still the issue of prompting user for project selection

PR #752, opened 2021-06-17
Add Redis Trace instrumentation to cartservice

arbrown:
This automatically adds trace information for the call from the cartservice to Redis and puts it in the context of the existing call from the frontend to the cartservice.

image

The service call is essentially just a wrapper around the Redis call, so they are basically the same in this case, but this is still a useful example of using OpenTelemetry to instrument a third-party library for trace.


daniel-sanche:
Would you mind adding some context to the PR description?

simonz130 (approved):

PR #769, opened 2021-06-28
fix: auto scaling min_node_count to align with spec

Ayelet41:


daniel-sanche (changes requested):
Currently blocking merging until we implement new git system

daniel-sanche (approved):

Ayelet41:
The initial count will be changed separately.

PR #788, opened 2021-07-06
feat: add app version to GKE cluster

Ayelet41:
For previous version will be 0 as we don't know what is the version. For that introduced a new GKE param(could use default, but wanted to be explicit).
 
- Check is if cluster exist, if so check if label exist. If not add it as 0, else use $VERSION(WITHOUT "." as it is not supported)

+Run fmt


daniel-sanche (changes requested):
Currently blocking merging until we implement new branching system

terraform/install.sh
```diff
@@ -188,6 +188,10 @@ createProject() {
 }
 
 applyTerraform() {
+  if [[ -n "$VERSION" ]]; then
+    app_ver="$(echo $VERSION | tr "." "_")"
```
daniel-sanche, line 192: what if there is no VERSION?

terraform/install.sh
```diff
@@ -201,12 +205,27 @@ applyTerraform() {
...
+  #check if new installtion or if cluster already and what is the version
+  gke_location="$(gcloud container clusters list --format="value(location)" --filter name=cloud-ops-sandbox)"
+  if [[ -n "$gke_location" ]]; then 
+    gke_version="$(gcloud container clusters describe cloud-ops-sandbox --region "${gke_location}"  --format="value(resourceLabels.version)")"
+  fi
+  #If cluster exist and it's older version use backward comp. params
+  if [[ -n "$gke_location"  && -z "$gke_version" ]]; then 
+    terraform_command+=' -var="app_version=0"'
+  else
+    terraform_command+=" -var=\"app_version=${app_ver}\""
```
daniel-sanche, line 225: This logic seems like it can be cleaned up a bit to me. Some of these checks are redundant

terraform/install.sh
```diff
@@ -201,12 +205,27 @@ applyTerraform() {
...
+  #If billing account provided specify it 
   if [[ -n "$billing_id" ]]; then
-    terraform apply -auto-approve -var="billing_account=${billing_acct}" -var="project_id=${project_id}" -var="bucket_name=${bucket_name}" -var="skip_loadgen=${skip_loadgen:-false}"
-  else
-    terraform apply -auto-approve -var="project_id=${project_id}" -var="bucket_name=${bucket_name}"  -var="skip_loadgen=${skip_loadgen:-false}"
+    terraform_command+=" -var=\"billing_account=${billing_acct}\"" 
   fi
+
+  #check if new installtion or if cluster already and what is the version
+  gke_location="$(gcloud container clusters list --format="value(location)" --filter name=cloud-ops-sandbox)"
```
daniel-sanche, line 217: Would this exit 1 if the cluster doesn't exist?

daniel-sanche (changes requested):

terraform/install.sh
Ayelet41: No, I doubled checked in this case gke_location will be empty and gke_version will not be checked

terraform/install.sh
Ayelet41: Modified

terraform/install.sh
Ayelet41: Modified, PTAL.

terraform/install.sh
```diff
@@ -201,12 +202,31 @@ applyTerraform() {
...
-    terraform apply -auto-approve -var="billing_account=${billing_acct}" -var="project_id=${project_id}" -var="bucket_name=${bucket_name}" -var="skip_loadgen=${skip_loadgen:-false}"
-  else
-    terraform apply -auto-approve -var="project_id=${project_id}" -var="bucket_name=${bucket_name}"  -var="skip_loadgen=${skip_loadgen:-false}"
+    terraform_command+=" -var=\"billing_account=${billing_acct}\"" 
   fi
+  #Check application version
+  #If cluster already exist leave original version
+  gke_location="$(gcloud container clusters list --format="value(location)" --filter name=cloud-ops-sandbox)"
+  if [[ -n "$gke_location" ]]; then 
+    gke_version="$(gcloud container clusters describe cloud-ops-sandbox --region "${gke_location}"  --format="value(resourceLabels.version)")"
```
minherz, line 216: do we deploy the cluster into region? does the command work if the `--region` parameter is passed with the value of the zone?

terraform/install.sh
```diff
@@ -201,12 +202,31 @@ applyTerraform() {
...
+    #If cluster exist and it's older version use backward comp params
+    if [[ -n "$gke_version" ]]; then 
+       app_ver=$gke_version
+    fi
+  elif [[ -n "$VERSION" ]]; then
+    app_ver="$(echo $VERSION | tr "." "_")"
+  else
+    app_ver="0" 
+  fi
```
minherz, line 225: if `gke_version` is not initialized we may end up with undefined `app_ver`. Maybe it worth to initiate it to `"0"` and then to run if-else logic?

terraform/install.sh
Ayelet41: Changed to zone

terraform/install.sh
Ayelet41: @daniel-sanche currently I'm using VERSION (which is the branch), is there a better param that is part of the code (vs env var) ?

daniel-sanche (approved):

daniel-sanche:
It might be a good idea to add a test if you can think of one. Otherwise LGTM

Ayelet41:
Issue to add test later: https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/834

terraform/variables.tf
```diff
@@ -37,6 +37,11 @@ variable "skip_loadgen" {
   description = "If true, the load generator will not be deployed."
 }
 
+variable "app_version" {
+  type        = string
+  description = "Cloud Operastions Sandbox's Version."
```
minherz, line 42: nit: consider more verbatim message and fix "Operastions"

minherz (approved):

PR #811, opened 2021-07-28
feat(terraform): add GKE location to TF apply

Ayelet41:
Use current GKE cluster location to avoid recreation +fmt


minherz (approved):
Good work!

minherz:
Mind to update the PR name to follow commits convetion.

PR #817, opened 2021-07-28
feat: recommendationservice crash recipe

daniel-sanche:
Added a new SRE recipe that triggers a crash in the recommendation service. This will show an HTTP500 page on the frontend, and trigger obvious error messages in the logs, making it a good candidate for future Qwiklabs

I also added basic tests using bash scripts, triggered in our E2E test pipeline


minherz:
@daniel-sanche can you have a look into an error of the e2e test? the error claims that it cannot find `.github/workflow/e2e_scripts/clean.sh`. It might be related to recent changes that were done to the repo.

sre-recipes/recipes/recommendation_crash_recipe/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2020 Google LLC
```
simonz130, line 1: perhaps, 2021?

sre-recipes/recipes/recommendation_crash_recipe/recommendation_crash_recipe.py
```diff
@@ -0,0 +1,137 @@
+# Copyright 2020 Google LLC
```
simonz130, line 1: 2021

simonz130 (approved):
Excellent work!

Issue #809, opened 2021-07-27
Fresh install should have no error logs


daniel-sanche:
Currently, when you install sandbox, some services create errors. We want a fresh installation to be stable, so we can break stuff with sre-recipes later

Before closing this issue, we should put tests in place to make it permanent

minherz:
I marked it as P:2 just to remove from triage list. Feel free to change priority label later.

PR #818, opened 2021-07-29
fix: currencyservice errors

daniel-sanche:
Partially addresses https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/809

There was an issue in the currencyservice that was causing it to spam error logs in the Logging dashboard. This PR bumps up the OT package versions, and updates the code to be in line with current best practices found in the OT documentation


minherz:
@daniel-sanche similar issue with PR #817 and PR #819

src/currencyservice/server.js
```diff
@@ -141,9 +139,7 @@ function _carry (amount) {
 function getSupportedCurrencies (call, callback) {
   // Extract the span context received from the gRPC client.
   // Create a child span and add an event.
-  const currentSpan = tracer.getCurrentSpan();
   const span = tracer.startSpan('currencyservice:GetSupportedCurrencies()', {
-    parent: currentSpan,
```
simonz130, line 146: is the new API automatically associating the new span with its parent?

daniel-sanche: I did some testing and it looks like the spans are captured correctly in the Cloud Console, but it would be good if someone with more OT experience could verify

daniel-sanche: Looking at more documentation, it does seem like it is supposed to capture the parent trace from the environment. But just to be safe, I added back in an explicit parent using the new API

simonz130 (approved):

PR #831, opened 2021-09-01
feat(terraform): Create GKE module

Ayelet41:
- Created GKE Module - using k8s provider 
- Apply manifests using k8s provider instead of kubectl apply
- To be able to do the former namespace is needed to be explicit in manifest and manifest files needed to be split as multiple manifest in one file are not supported. 
- For each is used to run on all folders


daniel-sanche:
Following up on our discussion today: If we move the manifest location, we'll also have to change the manifest reference in the [make-release workflow here](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/a7a6152aaafbcb1eb3eef30e79d7c4bf23b43492/.github/workflows/make-release.yml#L65) 

Make sure you do a repo search for "manifests" to see if there are any others we are forgetting

Ayelet41:
@daniel-sanche eventually I haven't change the path only split the yaml files, how would you recommend to document it? disclaimer at the top of all the pages, that terraform GKE manifests do not support multi yamls? adding readme to the dir? both? something else?

daniel-sanche:
This is looking out of date, so I'm going to close it. Let me know if you're still interested in working on it at some point in the future

PR #836, opened 2021-09-25
tests: Add integration tests for validating SRE Recipe config schema

Gan-Tu:
The validation mainly relies on the newest JSON Schema (https://json-schema.org) definition, and it works by first converting the YAML config to JSON, and then validate directly on the converted JSON. Added relevant README changes as well.

In order to make decoupling JSON schema easier, I decided to write a custom schema bundling logic for schema "references", since the Python version of the JSON schema validator doesn't support referencing local files well.

### DIFF
--- a/.github/workflows/e2e-latest.yml
+++ b/.github/workflows/e2e-latest.yml
@@ -54,14 +54,6 @@ jobs:
         IMAGE_PATH=gcr.io/${{ env.PROJECT_ID }}/tests/gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator/gke:$GITHUB_SHA
         docker build -t $IMAGE_PATH ./src/loadgenerator
         docker push $IMAGE_PATH
-    - name: Validate SRE Recipes Configs
-      timeout-minutes: 30
-      run: |
-        set -x
-        # install dependencies
-        python3 -m pip install -r tests/requirements.txt
-        # run validations
-        python3 tests/recipes/validate_recipe_configs.py
     - name: Rewrite Manifests
       run: |
         set -x
--- a/.github/workflows/e2e-latest.yml
+++ b/.github/workflows/e2e-latest.yml
@@ -95,6 +95,14 @@ jobs:
           -e LOADGEN_ZONE=$LOADGEN_ZONE \
           -v ~/.config:/root/.config \
           test-provisioning:$GITHUB_SHA
+    - name: Run SRE Recipes Config Validations
+      timeout-minutes: 30
+      run: |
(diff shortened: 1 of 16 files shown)


daniel-sanche:
So to make sure I'm reading this correctly: the purpose of this PR is to add a test to ensure recipes are valid by defining the expected JSON schema, and adding a test to check that all recipe YAML conforms to it?

And going forward, if we want to build the recipe codebase (eg, adding actions), we have to adjust the scemas, but when adding future recipes we don't have to touch anything?

Let me know if I have any misconceptions. Thanks!

Gan-Tu:
> So to make sure I'm reading this correctly: the purpose of this PR is to add a test to ensure recipes are valid by defining the expected JSON schema, and adding a test to check that all recipe YAML conforms to it?
> 
> And going forward, if we want to build the recipe codebase (eg, adding actions), we have to adjust the scemas, but when adding future recipes we don't have to touch anything?
> 
> Let me know if I have any misconceptions. Thanks!

Your understanding is correct. It ensures the YAML schema is followed, but no adjustment needed for new recipes, unless new actions are defined.

.github/workflows/e2e-latest.yml
```diff
@@ -54,6 +54,14 @@ jobs:
         IMAGE_PATH=gcr.io/${{ env.PROJECT_ID }}/tests/gcr.io/stackdriver-sandbox-230822/sandbox/loadgenerator/gke:$GITHUB_SHA
         docker build -t $IMAGE_PATH ./src/loadgenerator
         docker push $IMAGE_PATH
+    - name: Validate SRE Recipes Configs
```
daniel-sanche, line 57: This might do better in the standard [CI workflow](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/main/.github/workflows/ci.yml). The E2E workflow is used by tests that need to run against a live sandbox environment, but I don't think these tests should need that

daniel-sanche (changes requested):
LGTM

Thanks for expanding our test coverage around the recipes! That was a high priority need, and this seems like a great step

.github/workflows/e2e-latest.yml
Gan-Tu: Moved it out to ci.yaml.

It took quite some try because there are installation issues for pip with python3.5 that's running in ci.yaml. We should probably upgrade the default python version in the self hosted containers that's running the CI workflow.

Gan-Tu:
Moved test to CI

daniel-sanche (approved):

PR #871, opened 2021-11-04
fix(terraform): update deprecated GKE params

Ayelet41:
Update deprecated workload_identity_config and workload_metadata_config


minherz (approved):
lgtm

Issue #877, opened 2021-11-12
Ratings Service - python requirements.txt pyparsing version


balldarrens:
File:
https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/main/src/ratingservice/requirements.txt
Issue: pyparsing needs to be explicity added as a requirement to avoid the error below.

Issue in main, develop branches.

uwsgi==2.0.19.1
flask==1.1.4
psycopg2-binary==2.8.5
pyparsing==2.4.7
google-python-cloud-debugger

Full deployment with no issues.

╷
│ Error: Error waiting to create StandardAppVersion: Error waiting for Creating StandardAppVersion: Error code 9, message: Cloud build de469eea-4904-4c02-9f11-f135cb7594b0 status: FAILURE
│ found incompatible dependencies: "packaging 21.2 has requirement pyparsing<3,>=2.0.2, but you have pyparsing 3.0.6."
│ Full build logs: https://console.cloud.google.com/cloud-build/builds;region=us-east1/de469eea-4904-4c02-9f11-f135cb7594b0?project=792025362782
│
│   with module.ratingservice.google_app_engine_standard_app_version.default[0],
│   on ratingservice/main.tf line 117, in resource "google_app_engine_standard_app_version" "default":
│  117: resource "google_app_engine_standard_app_version" "default" {
│
╵

daniel-sanche:
This will be fixed in a new v0.7.3 release

PR #878, opened 2021-11-15
fix: rating service dependencies

daniel-sanche:
The rating service recently started to break deployment, due to a new dependency issue. This PR pins pyparsing, which should fix the issue

Fixes https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/877


minherz (approved):
lgtm
nit: might worth to explain why pyparsing dependency is there because we do not use it explicitly anywhere in the code.

PR #881, opened 2021-12-01
chore(docs): Update contributing.md

winter-code:
* updated steps to set up resources in docker
* update gcloud command flags to create gke cluster

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,10 +86,10 @@ brew install coreutils
    * [skaffold](https://github.com/GoogleContainerTools/skaffold/#installation)
      (ensure version ≥v0.20)
 
-1. Launch “Docker for Desktop”. Go to **Preferences**:
-   * Choose **Enable Kubernetes**
-   * Set **CPUs** to at least 3
-   * Set **Memory** to at least 6.0 GiB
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -137,7 +137,8 @@ gcloud services enable container.googleapis.com
 gcloud container clusters create demo --zone=us-central1-a \
     --machine-type=n1-standard-2 \
     --num-nodes=2 \
-    --enable-stackdriver-kubernetes \
+    --logging \
+    --monitoring \
```
daniel-sanche, line 141: It looks like this command isn't valid. `--logging` and `--monitoring` [expect a `COMPONENT` field](https://cloud.google.com/sdk/gcloud/reference/container/clusters/create#--monitoring) (either `SYSTEM`, `WORKLOAD`, or `NONE`).

I'm not too familiar with these options, but from a skim it seems like the default of `SYSTEM` might be enough? So maybe we don't need these two lines?

What ever we end up using, we should make the [same change to the Makefile](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/c984ea54dfd18a009d5018c66ea6e5a5b76f88b1/Makefile#L29)

PR #902, opened 2022-02-04
chore: install kubectl extension

daniel-sanche:
kubectl authentication is changing in the latest gcloud builds. We need to install an extension in the custom cloud shell image to be able to reliably authenticate to our GKE clusters


losalex (approved):

minherz (commented):
@daniel-sanche please edit your commit or make the commit validity check optional

daniel-sanche:
@minherz The commit text shouldn't be a problem since it will be replaced with the PR title when merged. (The bot usually knows this, but it fails when the PR is just a single commit)

In any case, the check is already optional

PR #923, opened 2022-04-29
feat: add flag to skip ratingservice deployment

daniel-sanche:
Some sandbox users don't make use of the ratingservice, and the extra ~10 minutes to deploy is seen as too costly.

This PR adds a new `--skip-ratingservice` flag to `sandboxctl` and `install.sh`, which will skip the deployment step of the ratingservice and related terraform resources

Deployment times:
~21 minutes with everything installed
~21 minutes with `--skip-loadgenerator`
~12 minutes with `--skip-ratingservice`
~9 minutes with `--skip-ratingservice` and `--skip-loadgenerator`


terraform/install.sh
```diff
@@ -260,7 +260,7 @@ installMonitoring() {
   log "Creating monitoring examples (dashboards, uptime checks, alerting policies, etc.)..."
   pushd monitoring/
   terraform init -lockfile=false
-  terraform apply --auto-approve -var="project_id=${project_id}" -var="external_ip=${external_ip}" -var="project_owner_email=${acct}" -var="zone=${CLUSTER_ZONE}"
+  terraform apply --auto-approve -var="project_id=${project_id}" -var="external_ip=${external_ip}" -var="project_owner_email=${acct}" -var="zone=${CLUSTER_ZONE}" -var="skip_ratingservice=${skip_ratingservice:-false}"
```
minherz, line 263: do we need to add `skip_loadgen` here too?

terraform/ratingservice/01_project.tf
```diff
@@ -24,4 +24,10 @@ resource "google_project_service" "cloudscheduler" {
...
   project                    = var.gcp_project_id
-}
\ No newline at end of file
+}
+
+resource "google_project_service" "cloudbuild" {
+  service                    = "cloudbuild.googleapis.com"
+  disable_dependent_services = true
+  project                    = var.gcp_project_id
+}
```
minherz, line 33: nit: why to add activation of Cloud Build?

terraform/ratingservice/02_cloudsql.tf
```diff
@@ -27,6 +27,7 @@ resource "random_password" "db_password" {
 }
 
 resource "google_sql_database_instance" "rating_service" {
+  project             = var.gcp_project_id
```
minherz, line 30: nit: this one and following additions are due to delete of the `ratingservice/01_provider.tf`. Would it be less trouble and less error prone keeping the provider declaration instead of introducing the need for this line in every google's resource?

minherz (changes requested):
Good work on fixes dependencies on resources.
There are a couple of changes that have to be re-validated before we adding them.

terraform/install.sh
daniel-sanche: yes, we need to disable ratingservice dashboards when ratingservice is disabled, right?

terraform/ratingservice/01_project.tf
daniel-sanche: It looks like app engine uses cloud build to deploy the instance. 

I was sometimes seeing issues where App Enggine deployment would fail unexpectedly, and I think it was because some of the necessary APIs weren't ready yet

terraform/ratingservice/02_cloudsql.tf
daniel-sanche: I had to remove the provider from the module, because terraform doesn't support providers when using `count` in modules.

There may be another way to set provider globally if you think this is an anti-pattern. I'd have to look into it

terraform/05_loadgen.tf
```diff
@@ -20,5 +20,5 @@ module "loadgen" {
   external_ip = data.external.terraform_vars.result.external_ip
   project_id  = data.google_project.project.project_id
 
-  depends_on = [null_resource.delay]
+  depends_on = [null_resource.delay, null_resource.ratingservice_address_rewrite]
```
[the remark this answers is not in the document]
daniel-sanche: I thought the intention of `depends_on = [null_resource.delay]` was that we wanted to wait until the primary cluster was finished deployment and stable before starting the load generator. If that's still the case, we now have to wait until the ratingservice rewrite step is complete before moving forward.

That said, I would like to find a way to make deploying the two clusters more parallel. But I thought that should probably come in a separate PR

terraform/03_gke_cluster.tf
```diff
@@ -225,6 +224,22 @@ resource "null_resource" "delay" {
...
+# Rewrite ratingservice address
+resource "null_resource" "ratingservice_address_rewrite" {
+  count = var.skip_ratingservice ? 0 : 1
+
+  provisioner "local-exec" {
+    command = <<-EOT
+    kubectl wait \-\-for=condition=available \-\-timeout=600s deployment/frontend
+    kubectl set env deployment.apps/frontend RATING_SERVICE_ADDR=${one(module.ratingservice[*].service_url)}
+    kubectl delete pod -l app=frontend
+    kubectl wait \-\-for=condition=available \-\-timeout=600s deployment/frontend
```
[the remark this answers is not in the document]
daniel-sanche: > The previous solution worked only because the first pod of the deployment was ready after the environment variable was updated. In result, its initialization code executed with the correct value of the env var.

The old code assumed that the pod would spin up after the environment variable was set, but there is no guarantee in the ordering. Or no guarantee that the deployment would be ready when the `set env` line was run. I was hoping to fix that to protect against those race conditions here

> The new implementation waits for the frontend deployment to complete.

Waiting for the deployment is necessary, since we are re-writing the environment variable on the deployment object. We want to make sure the deployment is in the cluster and in a known state before editing it

> The RATING_SERVICE_ADDR env var is used at initialization of the frontend service

For this reason, we then delete the old pod and wait for a fresh pod to re-initialize with the new variable and become active

terraform/01_provider.tf
```diff
@@ -41,3 +41,8 @@ terraform {
 
   }
 }
+
+provider "google" {
+  project = var.project_id
+  region  = var.appengine_region
```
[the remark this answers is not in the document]
daniel-sanche: Hmm you're right that this could be a little confusing. The primary GKE cluster and load generator cluster explicitly set their own regions, which I assume would take precedence (but I should probably verify those assumptions).

Do you think we should remove the region from the provider, and just set it on each resource as needed?

terraform/ratingservice/01_project.tf
minherz: Just checked and saw that the Cloud Build API in new projects is not activated. Indeed, it makes sense to activate it before AppEngine will do it. It improves execution time and might resolve the corner cases that you mentioned.

terraform/install.sh
minherz: I mean that this call does not set `skip_loadgen` parameter. My question was do we need to add the `--skip_loadgen` in addition to `skip_ratingservice`?

terraform/05_loadgen.tf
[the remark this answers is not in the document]
minherz: AFAIK load generator does not start sending requests on the start. the "ratingservice_address_rewrite" that you created is expected to happened before the overall deploy is completed and a user can initiate a simulated load.

terraform/03_gke_cluster.tf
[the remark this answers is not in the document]
minherz: Yes, I see now. Missed the pods' restart. Given that we start only one frontend pod it should be OK to wait for the complete startup

terraform/01_provider.tf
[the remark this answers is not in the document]
minherz: I think we can explicitly define this parameter for `provider` argument in the module call for rating service.
See [passing providers explicitly](https://www.terraform.io/language/modules/develop/providers#passing-providers-explicitly) example in the Terraform doc.

minherz (approved):
lgtm. i have addressed your feedback. please, have a look if you want to make some refactoring improvements before merging this PR.

terraform/01_provider.tf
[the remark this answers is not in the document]
daniel-sanche: good catch, I made the change

terraform/install.sh
daniel-sanche: Oh sorry, misread that. No, slip_loadgen isn't needed because there are no monitoring dashboards for the loadgen cluster

Issue #575, opened 2020-11-20
Refactor terraform installation to use /terraform/monitoring as terraform module


minherz:
_on behalf of @simonz130_ 

Current installation explicitly calls `terraform apply` on /terraform/monitoring when it will be more clear to execute it as a part of the centralized terraform provisioning by using the /terraform/monitoring as a terraform module.

The python script can be executed from within the module by using [local](https://www.terraform.io/docs/provisioners/local-exec.html) provisioner. For example, one can look into /terraform/ratingservice/main.tf for "google_sql_database" resource provisioning.

Ayelet41:
Consideration: 
- Such change will be breaking change(to some extent) moving forward, so will need to keep the 2 versions
- Currently istio installation add complexity, 
    - We currently have GKE+Istio, can we use it as a solution instead? 
    - Is waiting for istio services propagation is needed, given the recent changes in Cloud Ops? I tested it several times and `$service_wait` could be skipped.

PR #930, opened 2022-05-23
feat: upgrade Istio 1.7.1 to ASM

daniel-sanche:
Users had reported that monitoring data hasn't been showing up in Cloud Monitoring. This seems to be due to incompatibility with the out-dated Istio version being used by this project. To resolve this, I updated to use the latest version of ASM, which should be more useful for demos

- Updated Istio installation script to install ASM instead
- updated resources that referenced istio service ids with the ASM counterpart ids
- changed minimum node counts to align with ASM's expectation of 8 vCPU


daniel-sanche:
After doing some testing, it looks like some of the improvements in this branch aren't tied to ASM. I decided to break out the work into smaller PRs (https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/933, https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/932, https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/931), and then we can decide if upgrading to ASM still looks like a good idea from there:

losalex (approved):

daniel-sanche:
@mathieu-benoit Don mentioned you had ASM and terraform experience, so adding you as an optional review if you have the time. Feel free to ignore though. Thanks!

terraform/03_gke_cluster.tf
```diff
@@ -95,7 +97,7 @@ resource "google_container_cluster" "gke" {
       }
     }
 
-    initial_node_count = 4
+    initial_node_count = 2
```
mathieu-benoit, line 100: Other than cost because you also increased the size of the nodes, any particular reason to have 2 nodes now? Could we have at least 3 nodes to have corum? Furthermore, with Online Boutique the recommendation is to have 4 nodes at least, especially with ASM: https://github.com/GoogleCloudPlatform/microservices-demo#quickstart-gke. So maybe let's keep 4?

terraform/istio/install_asm.sh
```diff
@@ -0,0 +1,57 @@
...
+export SCRIPTPATH=$(dirname $(realpath $0))
+cd $SCRIPTPATH
+export WORK_DIR=$SCRIPTPATH
+
+echo "Downloading asmcli ${ASM_VERSION}..."
+curl https://storage.googleapis.com/csm-artifacts/asm/asmcli_${ASM_VERSION} > asmcli
+chmod +x asmcli
+
+echo "Installing ASM..."
+./asmcli install \
```
mathieu-benoit, line 39: Any reason why we can't use ASM Managed Control Plane (MCP) here? With MCP, `istiod` is out of the cluster which is saving some resources (CPU/Memory) usage as well as avoiding having to deal with ASM version upgrades. Could we consider this?

JFYI: If MCP is an option, this ASM TF module could help you with that: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/modules/asm

mathieu-benoit (commented):
That's great and important stuffs moving to ASM for the Cloud Ops Sandbox, more than happy to provide any feedback on this as far as I can, thanks for including me!

I will take some times later next week. In the meantime just left 2 comments/questions.

Also, do you mind sharing with me how quickly and easily I could deploy this if I can and if I have time please? Thanks!

terraform/03_gke_cluster.tf
daniel-sanche: Yeah, we can switch from 2 n1-standard-4s back to 4 n1-standard-2s if that's preferred. I assumed all things equal, less nodes would have lower overhead. But, I'm happy to change if that's not the case

I am interested in if you have any advice to lower the costs though, as an 8 vCPU cluster can be expensive for a demo app. Is there a minimal ASM cluster recommendation? Could we get away comfortably with something like a e2-small?

Also, is the 8vCPU requirement a hard cap? Sandbox is used in Qwiklabs, and they require special approvals to run labs with more than 6 vCPUs

terraform/istio/install_asm.sh
daniel-sanche: No reason. I'm not super knowledgeable about ASM, so I was just basing this off of what [I found in the public docs](https://cloud.google.com/service-mesh/docs/unified-install/install-anthos-service-mesh). I'll take a look at MCP.

I had seen that terraform module, but I was having trouble getting it working. I think we'd have to refactor more of our cluster creating configuration to use modules to get it to work properly. But I'm not super experienced with terrform either, so maybe it's just me.

We have an [open issue](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/575) to refactor terraform into modules later, so I was thinking we would revisit using the ASM TF module as a part of that

terraform/03_gke_cluster.tf
mathieu-benoit: The recommendation for ASM with the controlplane in-cluser, what you have now, is explained [here](https://cloud.google.com/service-mesh/docs/unified-install/anthos-service-mesh-prerequisites#cluster_requirements): `Anthos Service Mesh requires at least 8 vCPUs`. With that said, with ASM MCP (other comment) you have the advantage to remove the `istiod` component and its CPU/memory usage from in-cluster to outside the cluster. Very convenient to decrease size of the nodes of the cluster too.

For [Online Boutique](https://github.com/GoogleCloudPlatform/microservices-demo#quickstart-gke), the recommendation is `--machine-type=e2-standard-2 --num-nodes=4`.

The potential issue with 2 nodes is it's not respecting the quorum and best practice in case of failure while upgrading the clusters, etc. So 3 nodes at least could show best practice/recommendation here.

terraform/istio/install_asm.sh
mathieu-benoit: Yeah, MCP has plenty of advantages, when possible that's the recommendation to use it. In the case of the Cloud Ops Sandbox, here are the 2 main reasons why it's appealing:
- Get rid of the `istiod` component from the cluster to outside the clusters (less CPU/Memory usage in the cluster, more room for apps)
- The upgrade of ASM version is managed by Google, no need to upgrade the ASM/Istio version (`istiod` and the sidecar proxies).

To implement this with the new ASM TF module, it's very easy, see the example here: https://github.com/GoogleCloudPlatform/anthos-service-mesh-samples/blob/main/docs/terraform-asm-mcp/main.tf. No need to wrap the `amscli` bin yourself, very convenient and simplifying the installation experience.

Alternatively you could still leverage the ASM MCP by just updating what you do currently with the `asmcli` by adding the 2 parameters below.
```
./asmcli install
...
  --managed \
  --channel regular \
  --use_managed_cni
```
Here, like you mentioned it could be converted to the ASM TF module if you don't want to do it as part of this PR.

Hope that makes sense? We could also have a quick 1:1 chat internally, it will be my pleasure to connect with you :)

terraform/istio/install_asm.sh
```diff
@@ -0,0 +1,60 @@
...
+  --option legacy-default-ingressgateway \
+  --ca mesh_ca \
+  --enable_gcp_components \
+  --managed \
+  --channel regular \
+  --use_managed_cni
+
+echo "Enabling Sidecar Injection..."
+cd $WORK_DIR
+REVISION=$(kubectl get deploy -n istio-system -l app=istiod -o json | jq -r '.["items"][0]["metadata"]["labels"]["istio.io/rev"]')
```
mathieu-benoit, line 56: Now you are with MCP, you don't need this ;)

terraform/istio/install_asm.sh
daniel-sanche: Great, thanks for the detailed feedback, this is very helpful! I'm going to experiment a bit more with the terraform module, and will reach out to chat if needed.

Let me know if you notice anything else that seems off when you have time. We're hoping to merge this sometime next week

terraform/02_project.tf
```diff
@@ -85,6 +85,22 @@ resource "google_project_service" "sourcerepo" {
...
+
+  service = "gkehub.googleapis.com"
+
+  disable_dependent_services = true
+}
+
+resource "google_project_service" "meshconfig" {
+  project = data.google_project.project.project_id
+
+  service = "meshconfig.googleapis.com"
```
mathieu-benoit, line 99: Any particular reason why it's `meshconfig.googleapis.com` here?

I think, what you want here is `mesh.googleapis.com`? Ref:https://cloud.google.com/service-mesh/docs/unified-install/project-cluster-setup#required_apis. `mesh` is by the way enabling other APIs such as `meshconfig`, etc.

terraform/istio/install_asm.sh
```diff
@@ -0,0 +1,59 @@
...
+echo "Installing ASM..."
+./asmcli install \
+  --project_id $PROJECT_ID \
+  --cluster_name cloud-ops-sandbox \
+  --cluster_location $CLUSTER_ZONE \
+  --fleet_id $PROJECT_ID \
+  --output_dir $WORK_DIR/asm_output \
+  --enable_all \
+  --option prometheus-and-stackdriver \
+  --option legacy-default-ingressgateway \
```
mathieu-benoit, line 47: Are you using an Ingress Gateway as part of the Cloud Ops Sandbox? In other word, where is coming the Public IP address of the HipsterShop app you are deploying? If it's coming from `kubectl get svc frontend-external` you don't need to use this option.

mathieu-benoit (approved):
I did my review, this LGTM, great job!

Disclaimer: I haven't tested this PR myself.

For a future PR, here are what could be done in addition/on top of that:
- Use the [ASM TF Module](https://github.com/GoogleCloudPlatform/anthos-service-mesh-samples/blob/main/docs/terraform-asm-mcp/main.tf) to remove the need of the `asmcli` bin
- The Ingress Gateway, based on the above point will have its own dedicated installation manifests
- [Managed Data plane](https://cloud.google.com/service-mesh/docs/managed/configure-managed-anthos-service-mesh#managed-data-plane) seems to be a good candidate too for the Cloud Ops sandbox in order to simplify the upgrade of the sidecar proxies by Google
- Note: the `frontend-external` `Service` could also be removed from the manifests since the Ingress Gateway is the only Public IP needed.

@daniel-sanche, do you need to update the description of this PR to better reflect where you landed? I see some notes about WI but don't see anymore these changes as an example.

terraform/02_project.tf
daniel-sanche: Thanks, fixed!

daniel-sanche:
@mathieu-benoit  Thanks for the feedback! I added your comments to the [issue around terraform refactoring](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/575) for next quarter.

And yeah, it seems the description is a bit out of date as I ended up moving some parts into other PRs, but I'll make sure to fix it before merging

PR #931, opened 2022-05-25
fix: Cloud Ops permission errors

daniel-sanche:
Over the last few weeks, we've been seeing missing data in the Cloud Ops products on new sandboxes. I thought this was due to our outdated version of Istio, but after some experimentation, it looks like it was due to incorrect permissions on the GKE cluster service account

This PR fixes the issue by adding extra roles to the default compute instance service account, ensuring GCP observability client libraries can send data to Cloud Ops

### DIFF
--- a/terraform/03_gke_cluster.tf
+++ b/terraform/03_gke_cluster.tf
@@ -142,6 +142,43 @@ data "google_compute_default_service_account" "default" {
   ]
 }
 
+# Give service account Observability permissions
+resource "google_project_iam_member" "trace_role" {
+  project = data.google_project.project.project_id
+  role    = "roles/cloudtrace.admin"
+  member  = "serviceAccount:${data.google_compute_default_service_account.default.email}"
+  depends_on = [data.google_compute_default_service_account.default]
+}
+
+resource "google_project_iam_member" "monitoring_role" {
+  project = data.google_project.project.project_id
+  role    = "roles/monitoring.editor"
+  member  = "serviceAccount:${data.google_compute_default_service_account.default.email}"
+  depends_on = [data.google_compute_default_service_account.default]
+}
+
+
+resource "google_project_iam_member" "profiler_role" {
+  project = data.google_project.project.project_id
+  role    = "roles/cloudprofiler.agent"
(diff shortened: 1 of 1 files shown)


terraform/03_gke_cluster.tf
```diff
@@ -142,6 +142,43 @@ data "google_compute_default_service_account" "default" {
...
+resource "google_project_iam_member" "trace_role" {
+  project = data.google_project.project.project_id
+  role    = "roles/cloudtrace.admin"
+  member  = "serviceAccount:${data.google_compute_default_service_account.default.email}"
+  depends_on = [data.google_compute_default_service_account.default]
+}
+
+resource "google_project_iam_member" "monitoring_role" {
+  project = data.google_project.project.project_id
+  role    = "roles/monitoring.editor"
```
arbrown, line 155: If you want to limit permissions as much as possible, could this be `roles/monitoring.metricWriter` or does it need read access for something?
https://cloud.google.com/monitoring/access-control

terraform/03_gke_cluster.tf
```diff
@@ -142,6 +142,43 @@ data "google_compute_default_service_account" "default" {
   ]
 }
 
+# Give service account Observability permissions
+resource "google_project_iam_member" "trace_role" {
+  project = data.google_project.project.project_id
+  role    = "roles/cloudtrace.admin"
```
arbrown, line 148: Could possibly be `roles/cloudtrace.agent`

arbrown (approved):
Consider tightening the permissions if possible.  I _think_ that the permissions needed for the service account could be somewhat reduced on the first two entries, but I'm not as close to the project, so I could be missing something and I'll trust your judgment.

terraform/03_gke_cluster.tf
arbrown: Please excuse the double review.  Somehow this comment got left out of my first.

terraform/03_gke_cluster.tf
daniel-sanche: Good catch, I meant to come back to these to tighten the permissions, but must have been distracted by a different PR

Issue #838, opened 2021-10-04
Add tests to make sure services are healthy


daniel-sanche:
We want to make sure fresh sandboxes don't have any errors. Either error logs, or failing SLOs

daniel-sanche:
Also, the previous comment was focused on the GKE code, but we also have a new service running on App Engine. So after the GKE part, you'll have to do the same [deploy,](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/main/CONTRIBUTING.md#-run-app-engine-rating-micro-service) view logs, and write tests steps against that one too. But we can do that part at the end

daniel-sanche:
This is partially addressed in https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/932, although we may want to check for error logs. (It currently just checks to make sure logs are coming in)

minherz:
Correctness of Online Boutique is controlled via the pinned version in kustomization configuration.

PR #932, opened 2022-05-25
chore(tests): Add cloud ops data tests

daniel-sanche:
We recently had an error where no logs, metrics, or other cloud ops data was making it's way into sandbox's Cloud Operations dashboards, making the demo useless. This case was missed by our existing tests

- This PR attempts to fix this by rewriting `monitoring_integration_test.py` with tests that read from live observability data, to make sure it is available
- Note that these tests alone would not fix the issue, because our CI test project will always have the proper permissions set up. To help us test live sandbox environments easier, I also added `sandboxctl test`. This will let us run our test suite on fresh environments to make sure everything looks as it should, just like the CI tests do
- I also added a note to `RELEASING.md`, so we remember to check this case in our release process checklist

Related issue: https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/838


tests/monitoring_integration_test.py
```diff
@@ -77,6 +78,31 @@ def getZone():
 
     return zone
 
+_services_short = [
```
losalex, line 81: I would combine it into single array of tuples - what happens if you add a value in one array and forget the second?

tests/monitoring_integration_test.py
```diff
@@ -137,352 +163,226 @@ def testUptimeCheckAlertingPolicyNotificationChannel(self):
...
+                print(f"✅  {service_name} {slo_type} SLO created")
+
+    def test_availability_slos_passing(self):
+        """
+        Ensure that availability is at expected levels to pass the SLO
+        """
+        for service_name in _services_short:
+            # get SLO
+            istio_service_name = self.getIstioService(service_name)
+            slo_id = f"{service_name}-availability-slo"
```
losalex, line 325: I see that this piece of code also exists in `test_latency_slos_passing()` - can you please take a look and see if you can move reusable code into separate functions?

losalex (approved):
Good change! Left couple of comments - can you please see if you can avoid code duplication in rest of the places?

tests/monitoring_integration_test.py
daniel-sanche: Good point, done

tests/monitoring_integration_test.py
daniel-sanche: Done

PR #933, opened 2022-05-25
feat: simplify service dashboards

daniel-sanche:
- Previously, terraform was setting up duplicate dashboards; some created by istio, and some created manually. This PR removes the custom ones, so the Cloud Monitoring interface is easier to read
- I also loosened some of the SLO definitions. Currently they would instantly fail and send out alerts. We would prefer a fresh sandbox to be within its SLO budget

### DIFF
--- a/terraform/monitoring/03_services.tf
+++ b/terraform/monitoring/03_services.tf
@@ -12,81 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Define specifics for each of the services that will receive SLOs through a custom service
-# The data members needed are to specify the custom service, goals for each SLO (availability and latency),
-# and maximum error budget burn rates.
-variable "custom_services" {
-  type = list(object({
-    service_name           = string,
-    service_id             = string,
-    availability_goal      = number,
-    availability_burn_rate = number,
-    latency_goal           = number,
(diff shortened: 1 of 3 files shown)


arbrown (approved):
LGTM. I love the cleaner format.

With regards to the increased latency thresholds, were you observing consistent violations on specific services?  Or is this just generally to reduce noise on the dashboards? I'm curious if, for example, the currency service ever does go above 500 ms.

arbrown:
I re-read the previous comment, and I do recall seeing some violations during deployment.  I think I'm correct in assuming that the latencies are generally well within even the old threshold after that though, right?

daniel-sanche:
> I re-read the previous comment, and I do recall seeing some violations during deployment. I think I'm correct in assuming that the latencies are generally well within even the old threshold after that though, right?

These changes were based on the SLOs that I saw turning red on the dashboard. I didn't spend too much time looking into why the latencys are so high. It may just be warm-up costs, or it may be something else. I've been doing testing without the ratingservice for quicker deployment, so that may also be adding extra latency to the app.

In general, my thinking has been that we should use high SLO values for our sandbox dashboards instead of realistic production ones, to make sure everything looks good on a fresh deployment. Let me know if you think keeping it low is a better idea

PR #944, opened 2022-06-15
fix: enabled cloud sql admin api

daniel-sanche:
The ratingservice was failing to connect to Cloud SQL due to an unenabled API. This PR enables the API as part of terraform deployment

PR #945, opened 2022-06-15
fix: Enable the cloud-sql-proxy path to Cloud SQL for ratingservice

joyoza:
Enable the Cloud SQL Admin API and add Cloud SQL Client role for the
App Engine service account. 
See https://cloud.google.com/sql/docs/mysql/connect-app-engine-standard for more background.


daniel-sanche:
Thanks for putting this together. I just opened https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/pull/944 with the same purpose, but I'll close it in favor of yours (with the extra GAE service account changes added)

daniel-sanche (approved):

Issue #943, opened 2022-06-15
add dependency on release pipeline jobs


daniel-sanche:
Currently the two jobs in `make-release.yaml` run in parallel. We want the second job to run only after the first is complete. We should [add a `needs` tag](https://docs.github.com/en/actions/using-workflows/about-workflows#creating-dependent-jobs) to the second job

daniel-sanche:
Also, it looks like the redundant "rebuild cloudshell" job didn't trigger as expected. Looking at [the examples](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow), it may be because we used quotes in the trigger name

PR #946, opened 2022-06-15
chore: test adjustments

daniel-sanche:
- We recently added `sandboxctl test` to run tests against a live sandbox environment, but it wasn't working for the ratingservice tests. This fixes that
- added a dependency to the make-release release pipeline, so it will run steps in sequence instead of parallel (https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/943)
- removed quotes from workflow custom image trigger, since the trigger wasn't firing and removing quotes seems to be better in line with the spec


minherz (approved):
lgtm

Issue #917, opened 2022-04-07
Sandbox installation lists billing accounts even if a user does not have permission to link the billing account to a project  


minherz:
The standard installation process lists all billing accounts that a user can see. However, it does not check whether a user has permissions to link the billing account to a project. it leads to situations when a user tries to use the listed billing account and the installation process fails with the following error:
```
ERROR: (gcloud.beta.billing.projects.link) User [LDAP@google.com] does not have permission to access projects instance [cloud-ops-sandbox-<some_id>] (or it may not exist): The caller does not have permission
```

The discovery of the billing accounts should validate for `roles/billing.user` or `roles/billing.projectManager` or `roles/billing.admin`. Alternatively, a user can be validated to have a permissions `billing.resourceAssociations.create`.

minherz:
Closing this issue as obsolete.
Starting from next release, the standard installation of the Sandbox will not include a new project provisioning. The installation path with project provisioning and installation of the Online Boutique (instead of internally-managed microservice demo) will be deprecated.

PR #948, opened 2022-06-16
fix: Add billing permission validation to link to project

yuhuayang-google:
I added a validation to first check if the user has the correct IAM role to link a billing account to a project before including those accounts for selection (or if only 1 billing account make sure the user has permission and if not, exit with the appropriate error)

This PR is to fix https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/917

### DIFF
--- a/terraform/install.sh
+++ b/terraform/install.sh
@@ -37,8 +37,8 @@ sendTelemetry() {
 
 promptForBillingAccount() {
   log "Checking for billing accounts..."
-  found_accounts=$(gcloud beta billing accounts list --format="value(displayName)" --filter open=true --sort-by=displayName)
-  if [ -z "$found_accounts" ] || [[ ${#found_accounts[@]} -eq 0 ]]; then
+  found_all_accounts=$(gcloud beta billing accounts list --format="value(displayName,name)" --filter open=true --sort-by=displayName)
+  if [ -z "$found_all_accounts" ] || [[ ${#found_all_accounts[@]} -eq 0 ]]; then
     log "error: no active billing accounts were detected. In order to create a sandboxed environment,"
     log "the script needs to create a new GCP project and associate it with an active billing account"
     log "Follow this link to setup a billing account:"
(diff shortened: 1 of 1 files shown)


minherz:
@yuhuayang-google please, rename your commit to follow the convention (prefix with `fix:`)

terraform/install.sh
```diff
@@ -50,6 +50,30 @@ promptForBillingAccount() {
...
+    fi
+  done
+
+  if [[ ${#found_accounts[@]} -eq 0 ]]; then
+    log "error: you don't have permissions to link projects to billing accounts,"
+    log "Follow this link to setup a new billing account that you can link:"
+    log "https://cloud.google.com/billing/docs/how-to/manage-billing-account"
+    sendTelemetry "none" no-active-billing
+    exit 1;
+  fi
```
minherz, line 74: nit: the case when no billing accounts found and when the authorized user does not have permissions to an account are identical. it might be useful to refactor the code and to print the same message asking to ensure that the executor has the required permissions to at least one billing account.

terraform/install.sh
```diff
@@ -50,6 +50,30 @@ promptForBillingAccount() {
...
+ #check for permission to link account
+  found_accounts=()
+  IFS=$'\n'
+  found_all_accounts=($found_all_accounts)
+  for acc in ${found_all_accounts[@]}
+  do
+    IFS=$'\t'
+    acc=($acc)
+    IFS=$'\n'
+    test_perm=$(gcloud beta billing accounts get-iam-policy "${acc[1]}" --flatten=bindings --filter="bindings.role:roles/billing.projectManager OR bindings.role:roles/billing.user OR bindings.role:roles/billing.admin" --format='value(bindings.members)')
```
minherz, line 62: it seems that the [required roles](https://cloud.google.com/billing/docs/how-to/modify-project#required-permissions-change) to link a project (assuming a user is the owner) and a billing account are:
`roles/billing.admin` **OR** (`roles/billing.user` **+** `roles/billing.viewer`)

Checking for `roles/billing.projectManager` is not required because the projects in use are owned by the user.

Also, it is possible that the caller is not in the list (`bindings.members`). Please, make sure that the authenticated identity is in the list when you validate the billing permissions. You can use the following command to acquire the identity:
```bash
gcloud auth list --filter="status:active" --format="value(account)"
```

terraform/install.sh
```diff
@@ -50,6 +50,30 @@ promptForBillingAccount() {
     exit 1;
   fi
 
+ #check for permission to link account
+  found_accounts=()
+  IFS=$'\n'
```
minherz, line 55: please use `IFS_BAK` to remember the previous settings, so it can be effectively restored.

minherz (changes requested):
Thank you for contributing your solution. This problem causes frictions when users have access to multiple billing accounts and some of the accounts are for view only.

The validation is not trivial and we are grateful to you for taking time to solve this problem.
Reviewing the current implementation I see that it validates only that the IAM policy of the billing account has a relevant IAM role. It still does not ensure that the caller (currently authenticated user) is in the `members` list for this role. Please, modify the test to include this validation.

minherz:
Closing due the breaking changes in 0.9.0 release.

PR #953, opened 2022-08-18
docs: removes `$` from shell commands

### LINKED ISSUE #882: Neos tutorial copy/paste commands prepended with `$`
When you use the copy and paste functionality in the sandbox, it includes the `$`:
```
$ sandboxctl sre-recipes hint recipe0
```
This should be removed so the copy and paste does not error with: `-bash: $: command not found`

cc @kelsk

minherz:
removes `$` symbol from beginning of the code blocks to
allow copy/paste commands into shell

Fixes #882


losalex (approved):

minherz:
PR was merged despite failing SRE recipes. The change is unrelated to recipes. The problem will be investigated separately.

PR #961, opened 2022-09-08
chore: Run terraform fmt in ci

arbrown:


.github/workflows/ci.yml
```diff
@@ -28,11 +28,17 @@ jobs:
   run-tests:
     runs-on: [self-hosted, kind-cluster]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
+    - uses: hashicorp/setup-terraform@v2
+      with:
+        terraform_version: 0.15.1
```
minherz, line 34: we constraint to run the plan with terraform `~>1.0`. Does it make sense to run the `terraform fmt` with older version?

.github/workflows/install-dependencies.sh
```diff
@@ -55,5 +55,11 @@ gcloud auth configure-docker --quiet
 # install pip3
 sudo apt-get install python3-pip
 
+# install unzip
+sudo apt-get install unzip
+
+# install node (for terraform)
+sudo apt-get install nodejs
+
```
minherz, line 63: why do we need to install these if everything worked before? does it mean the terraform actions do not work otherwise?

minherz (commented):
i think we should sync constraints for terraform version between CI and the sandbox deployment.

.github/workflows/install-dependencies.sh
arbrown: Previously, there was no terraform as part of the CI jobs.  Unzip is necessary to install the pre-packaged Github action, though other methods are possible.  Nodejs was necessary to run terraform on that machine.

.github/workflows/ci.yml
arbrown: Good point.  I've bumped it.

.github/workflows/ci.yml
```diff
@@ -28,11 +28,17 @@ jobs:
   run-tests:
     runs-on: [self-hosted, kind-cluster]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
+    - uses: hashicorp/setup-terraform@v2
+      with:
+        terraform_version: 1.0.11
```
minherz, line 34: nit: would we want to keep the same notation of "latest patch version": `~>1.0`?

minherz (approved):
lgtm with one question.

.github/workflows/ci.yml
arbrown: Sure.  The syntax is a bit different for this config, but I updated it.

minherz (approved):

PR #969, opened 2022-09-23
fix: ratingservice uwsgi build issue

daniel-sanche:
We are currently seeing a uwsgi issue when deploying the ratingservice:

image

I believe this issue comes down to python3.8 being removed from the Cloud Shell base image in favor of Python 3.9. This PR attempts to fix the issue by updating the ratingservice to use python39, and updates the uwsgi package


minherz:
changes were already merged

Issue #995, opened 2023-01-10
Broken cartservice - busybox `nslookup` misreturns 1 instead of 0


palladius:
urrently lastest version of https://cloud-ops-sandbox.dev/ is broken.

A fresh isntall fails on the cartservice.

I've done a long investigation with @alml in [1] shared with leoy@

A quick/cheap fix would be good enough.

[1] https://docs.google.com/document/d/1RTEKaDlP9PwoNfKvpAFxjbQYCZ0o5kA9Pj_V26kqu3Y/edit# [2]

palladius:
From Leonid: 

The investigation is still in progress. So far I can confirm that the cause of the problem is init container in the cartservice pod that fails. The workaround of the problem is to delete the cartservice deployment. Ensure that redis-cart deployment and service are in ready state. Delete the initContainers section from the cartservice.yaml (in the kubernetes-manifests/ folder and re-deploy the cartservice

palladius:
the part which needs to be removed is

```
 initContainers:
      - command:
        - bin/sh
        - -c
        - until nslookup redis-cart; do echo waiting for redis; sleep 2; done;
        image: busybox
        imagePullPolicy: Always
        name: init-redis-ready
```

palladius:
Alex and I noticed that this command returns correctly on main container but poorly on the init container:

```$ nslookup redis-cart
Server:		10.28.0.10
Address:	10.28.0.10:53

Non-authoritative answer:
Name:	redis-cart.default.svc.cluster.local
Address: 10.28.2.181

** server can't find redis-cart.svc.cluster.local: NXDOMAIN

** server can't find redis-cart.cluster.local: NXDOMAIN

** server can't find redis-cart.cluster.local: NXDOMAIN

** server can't find redis-cart.svc.cluster.local: NXDOMAIN

** server can't find redis-cart.google.internal: NXDOMAIN

** server can't find redis-cart.google.internal: NXDOMAIN

** server can't find redis-cart.c.cloud-ops-sandbox-2646743255.internal: NXDOMAIN

** server can't find redis-cart.c.cloud-ops-sandbox-2646743255.internal: NXDOMAIN

/app # echo $?
0
```

It would incorrectly return 1 on the init (where the SHELL env was slightly different, maybe a differen versioj busybox? Leonid suggests it might be a bug in busybox and I agree.

palladius:
I can confirm this change works: 

```
  initContainers:
[...]

PR #996, opened 2023-01-10
Update cartservice.yaml

palladius:
workaround for https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/995

### DIFF
--- a/kubernetes-manifests/cartservice.yaml
+++ b/kubernetes-manifests/cartservice.yaml
@@ -27,9 +27,12 @@ spec:
     spec:
       terminationGracePeriodSeconds: 5
       initContainers:
-        - name: init-redis-ready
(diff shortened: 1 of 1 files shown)


minherz:
I am going to close this PR because it does not follow the current process of submitting changes to Sandbox.
For more details about the process you can read at [CONTRIBUTING.md][1].
Also, it seems that the problem is due to the [issue][2] in the busybox container image. Would it have more sense to pin the "good" version of the busybox (to 1.36.0) instead of modifying the command line?

[1]: https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/main/CONTRIBUTING.md#git-workflow
[2]: https://github.com/docker-library/busybox/issues/48

PR #997, opened 2023-01-10
Update cartservice.yaml v2

palladius:
Fixing by backporting busybox


minherz:
The issue was handled in #999

PR #1044, opened 2023-08-25
Update existing issue templates

minherz:
Make existing issue templates more robust by using [issue forms][1].

[1]: https://github.com/GoogleCloudPlatform/microservices-demo/issues/new/choose


minherz:
Changes to the templates were enrolled via a sequence of direct commits.

PR #1050, opened 2023-09-19
fix: Fix root README badges to use the main branch' results

bourgeoisor:
This PR fixes https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/1051:
- Makes sure that the CLI and Terraform workflows run on the main branch instead of just in PRs
- Makes sure that the root README Action badges only uses the results taken from latest main commit, rather than "whichever last PR was opened / had the workflows run" which leads to a mismatch between what users think the badges mean (latest commit) and what the badges used to do (latest PR, regardless if it was even merged)

### DIFF
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -14,6 +14,9 @@
 name: CLI
 
 on:
+  push:
+    branches:
+      - main
   pull_request:
     types: [opened,synchronize,reopened]
 
--- a/.github/workflows/terraform.yaml
+++ b/.github/workflows/terraform.yaml
@@ -14,6 +14,9 @@
 name: Terraform
 
 on:
+  push:
+    branches:
+      - main
   pull_request:
     types: [opened,synchronize,reopened]
     paths:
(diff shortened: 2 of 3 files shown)


minherz (changes requested):
These workflows are intended to run on more branches than only `main`.
they are configured as "required" for protected branches.

Updating README makes sense. however, the trunk-based model does not allow pushing any change to main. any change has to go through the release process. I do not see urgency in this change or a reason for releasing a version patch that carries it.

Please, create an issue that describes the problem and suggest the fix. It will be applied to 0.10 release.

bourgeoisor:
- This PR does not affect the workflows running on PRs. Workflows still run on PRs *and* will now also run on the main branch.
- Created an issue and linked it to this PR: https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/issues/1051

minherz:
As discussed, we will remove the badge display from the README file at the root of the repo.
The configuration does lot support merging directly to the `main` branch. The change will be applied either as a part of the next release or by a patch.

PR #1052, opened 2023-10-02
feat: create `default` VPC network if it does not exist already

### LINKED ISSUE #1043: default VPC Requirement
Thanks for stopping by to let us know something could be better!

 **Is your feature request related to a problem? Please describe.**
Sometimes the new projects don't have the default VPC network by default and the create script fails because GCE instances cannot be created.

 **Describe the solution you'd like**
It might be good to mention that "default" auto VPC is a pre-req for the script to create the sandbox environment

 **Describe alternatives you've considered**
Another alternative could be creating it as part of the terraform script if it doesn't exist.

 **Additional context**

lvaylet:

- [X] Make sure to open an issue before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Your change follows guidelines in the [Contributor guide](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/main/CONTRIBUTING.md)

🛠️ Fixes #1043


lvaylet:
Hi @minherz, any chance you could review this PR for `milestone/0.10`?

minherz:
**updated:** This change is rolled back. The rationale behind the rollback includes:
* COB deprecates support for provisioning demo applications in order to focus on demonstration of the observability features of Cloud Ops services. The expectations for the default VPC configuration exists in the provided project is reasonable and works like this in majority of deployments.
* Provisioning a new VPC on the user's project may have unexpected side-effects on other workflows running in the project
* The implemented provisioning lacks dependencies which results in unstable deployment and multiple failures due to unresolved dependencies.

lvaylet:
Thanks for the update. I just tested this approach in a clean project with no `default` VPC. It does not work indeed. I was pretty sure I had tested it prior to submitting the PR. I probably did so in a project that already had a `default` VPC.

PR #1061, opened 2024-04-26
fix(terraform): update min_count and max_count to of min_node_count

Ayelet41:

- [ ] Make sure to open an issue before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Your change follows guidelines in the [Contributor guide](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/main/CONTRIBUTING.md)

🛠️ Fixes #<issue_number_goes_here>

### DIFF
--- a/provisioning/terraform/asm.tf
+++ b/provisioning/terraform/asm.tf
@@ -13,36 +13,13 @@
 # limitations under the License.
 
 
-resource "null_resource" "install_asm" {
-  count = var.enable_asm ? 1 : 0
+module "asm" {
+  source = "terraform-google-modules/kubernetes-engine/google//modules/asm"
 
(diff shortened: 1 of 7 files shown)


minherz:
`min_count` and `max_count` variables are used in `gke.tf` as input parameters to `gke` module (terraform-google-modules/kubernetes-engine/google).
I am reviewing the fact that the "main" and "release/0.9" branches are out of sync. The result of this work will be a patch `0.9.4` that should address the current problem with installing `0.9.3` and synchronize the branches.
If the use of `gke` and `asm` modules will be proved impossible, the changes to the relative terraform modules will be rolled back.

PR #1062, opened 2024-04-26
fix(terraform): Update module names

Ayelet41:

- [ ] Make sure to open an issue before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Your change follows guidelines in the [Contributor guide](https://github.com/GoogleCloudPlatform/cloud-ops-sandbox/blob/main/CONTRIBUTING.md)

🛠️ Fixes #<issue_number_goes_here>

### DIFF
--- a/provisioning/terraform/asm.tf
+++ b/provisioning/terraform/asm.tf
@@ -13,36 +13,13 @@
 # limitations under the License.
 
 
-resource "null_resource" "install_asm" {
-  count = var.enable_asm ? 1 : 0
+module "asm" {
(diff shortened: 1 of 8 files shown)


minherz:
Thank you for catching up the problem. Indeed, the provisioning of Online Boutique should start _after_ ASM is installed.
I am reviewing the fact that the "main" and "release/0.9" branches are out of sync. The result of this work will be a patch `0.9.4` that should address the current problem with installing `0.9.3` and synchronize the branches.
If the use of `gke` and `asm` modules will be proved impossible, the changes to the relative terraform modules will be rolled back.

